PR #4, opened 2024-04-28
Implement boilerplate code for tracking events with Mixpanel

hongil0316:


yoland68:
Looks great! love the wrapper style, we should have an option to track the first time running comfy-cli and prompt user to opt out of tracking.

yoland68:
As for whether it was the first time running, I believe @ltdrdata is adding a config to keep track of comfy-cli level of meta data (e.g. the most recent repo that has been run through comfy-cli). Ideally the comfy-cli config is done using similar singleton class as MetadataManager or EnvChecker, so it can be used systematically in multiple places.

hongil0316:
> Looks great! love the wrapper style, we should have an option to track the first time running comfy-cli and prompt user to opt out of tracking.

Yeah left out as a TODO, I'll handle that in a follow-up PR!

PR #15, opened 2024-05-01
feat/install snapshot

ltdrdata:
`comfy install --snapshot=<.yaml path>` is added.

Currently, only the installation of ComfyUI and custom nodes is being applied.


comfy_cli/cmdline.py
```diff
@@ -67,6 +69,46 @@ def init():
...
+    ],
+):
+
+    if not output.endswith(".yaml"):
+        print(f"[bold red]The output path must end with '.yaml'.[/bold red]")
+        raise typer.Exit(code=1)
+
+    output_path = os.path.abspath(output)
+
+    config_manager = workspace_manager.config_manager
```
yoland68, line 89: config_manager is also singleton, you can safely initiating it with `ConfigManager()`

mattmillerai:
Closing as stale. This has been open ~2 years and now conflicts with `main`, and the snapshot capability it proposed has since shipped another way: `comfy node save-snapshot` / `comfy node restore-snapshot` (delegating to ComfyUI-Manager's cm-cli).

The one piece not covered by `restore-snapshot` is the specific one-shot `comfy install --snapshot=<yaml>` flow (bootstrap ComfyUI **and** nodes in a single command). If that's still wanted, it's best re-filed as a fresh feature request against today's `install.py` rather than rebased from here.

Thanks for the original contribution. Related: #217 (fast-deps for restore-snapshot).

PR #45, opened 2024-05-11
add install script for ubuntu

ltdrdata:


yoland68:
This is to address the installer script task, right?
We should try to have a universal script for all UNIX systems and one for PS1/BAT.
E.g. https://raw.githubusercontent.com/richinsley/comfycli/main/install_scripts/installer.sh
https://github.com/mamba-org/micromamba-releases

yoland68:
Also, we should prob assume that python is installed already and check if venv/conda is installed, but avoid helping users to install those (or else we are putting too much burden to making sure python/venv/conda is installed correctly)

PR #64, opened 2024-05-16
fix: civitai downloading

ltdrdata:
fix: Download the model to the appropriate location based on its type.
fix: civitai download fail  when model page contains training data
fix: model remove command
improve: prompt support for unknown path specifying


yoland68:
Thank you! LGTM

PR #101, opened 2024-07-03
Fix: Install command with just --workspace flag doesn't work

punitda:
**Issue:**
When you try to run following command to install comfy in particular workspace it doesn't works
```bash
comfy --workspace=<workspace path> install
```

It throws following error:

```
Error 
│ Invalid value for '--version' / '-v': option `version` is mutually     │
│ exclusive with option `workspace`                                      │
╰─────────────────────────────────────
```

**Expected:**
The install command should work with just workspace flag provided that we're not adding other mutually exclusive flags like `--here` and `--recent`

**Fix:**
We need to wrap [version](https://github.com/Comfy-Org/comfy-cli/blob/main/comfy_cli/cmdline.py#L115-L122) with `Annotated` and making it Optional to avoid triggering exclusivity_callback on it. 

**Results:**

Before fix:

Screenshot 2024-07-03 at 11 43 49 AM

After fix:

Screenshot 2024-07-03 at 11 43 04 AM

Let me know if there is a better way to fix it. I'd be happy to change it

### DIFF
--- a/comfy_cli/cmdline.py
+++ b/comfy_cli/cmdline.py
@@ -112,14 +112,15 @@ def entry(
             help="Enable tracking",
         ),
     ] = True,
-    version: bool = typer.Option(
-        False,
-        "--version",
-        "-v",
-        help="Print version and exit",
-        is_flag=True,
-        callback=exclusivity_callback,
-    ),
+    version: Annotated[
+        Optional[bool],
+        typer.Option(
+            show_default=False,
+            is_flag=True,
+            help="Print version and exit",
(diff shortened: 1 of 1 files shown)


punitda:
Closing

robinjhuang:
@punitda Thanks for contributing! I didn't see this PR before I merged mine 😅

punitda:
> @punitda Thanks for contributing! I didn't see this PR before I merged mine 😅

@robinjhuang No worries! Glad that fix was merged in 1.0.32! 🙂

PR #105, opened 2024-07-09
Run additional commands related to model and nodes in integration test

james03160927:


comfy_cli/registry/api.py
```diff
@@ -18,6 +18,10 @@ def __init__(self):
     def determine_base_url(self):
         if os.getenv("ENVIRONMENT") == "dev":
             return "http://localhost:8080"
+        elif os.getenv("ENVIRONMENT") == "staging":
+            # print environment variable
+            print(os.getenv("ENVIRONMENT"))
+            return "https://staging-comfy-backend-qod3oz2v2q-uc.a.run.app"
```
robinjhuang, line 24: `https://stagingapi.comfy.org` works

PR #108, opened 2024-07-10
Fix `check_comfy_repo`

huchenlei:
Currently `check_comfy_repo` is returning `Tuple[bool, Optional[git.Repo]]`, but most call site is treating the return value as a bool, so the check at most callsites just always pass, because the a tuple of 2 items always evals to True.

This PR does following changes:
- Rename `check_comfy_repo` to `get_comfy_repo` as the git.Repo is the most important return value
- Simplify the return value to `Optional[git.Repo]` as the bool can be obtained from nullity of `Optional[git.Repo]`
- Change call site where a tuple of 2 itmes are expected

Note:
This will likely break things. So don't merge it until we have more test coverage.

### DIFF
--- a/comfy_cli/cmdline.py
+++ b/comfy_cli/cmdline.py
@@ -6,6 +6,7 @@ import uuid
 import webbrowser
 from typing import Optional
 
+import git
 import questionary
 import typer
 from rich import print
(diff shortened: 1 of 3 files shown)


huchenlei:
Closing as conflict with #109

PR #109, opened 2024-07-10
Fixed: Install --here

robinjhuang:
- Updated WorkspaceManager get_workspace_path() to only return the path based on given user inputs
- It does not validate the existence of a ComfyUI repo at that path 
- This can be done in the command itself. eg `install` might not need a validate repo because it will install it

### DIFF
--- a/comfy_cli/test_cmdline.py
+++ b/comfy_cli/test_cmdline.py
@@ -5,17 +5,6 @@ from .cmdline import app
 runner = CliRunner()
 
 
-def test_install_here():
-    result = runner.invoke(
-        app,
-        ["--here", "--skip-prompt", "install", "--cpu"],
-    )
-    print("Stdout:")
-    print(result.stdout)
-    assert result.exit_code == 0
-    assert "ComfyUI is installed at" in result.stdout
-
-
 def test_display_version():
     result = runner.invoke(
         app,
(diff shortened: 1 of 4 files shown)


huchenlei:
Fix verified working.

comfy_cli/workspace_manager.py
```diff
@@ -215,50 +214,33 @@ def get_workspace_path(self) -> Tuple[str, WorkspaceType]:
...
             found_comfy_repo, comfy_repo = check_comfy_repo(current_directory)
             if found_comfy_repo:
                 return comfy_repo.working_dir, WorkspaceType.CURRENT_DIR
-
-            print(
-                "[bold red]warn: you are not currently in a ComfyUI directory.[/bold red]"
-            )
-            raise typer.Exit(code=1)
+            else:
+                return current_directory + "/ComfyUI", WorkspaceType.CURRENT_DIR
```
huchenlei, line 243: `os.path.join` here

comfy_cli/test_cmdline.py
```diff
@@ -0,0 +1,26 @@
+from typer.testing import CliRunner
```
huchenlei, line 1: I added test on `--workspace` as well on https://github.com/Comfy-Org/comfy-cli/pull/110.

Maybe merge my PR first and remove this test file? I also fixed the issue on mutually exclusive test between test runs.

PR #112, opened 2024-07-10
Move `test_cmdline.py` to `test/`

huchenlei:


.pre-commit-config.yaml
```diff
@@ -3,7 +3,6 @@ repos:
     rev: 22.3.0
     hooks:
       - id: black
-        language_version: python3.9
```
huchenlei, line 6: python 3.9 for a hook is pretty unreasonable. Removing the constraint for developer who use higher version of python.

PR #113, opened 2024-07-10
Add E0401 check

huchenlei:
https://pylint.readthedocs.io/en/latest/user_guide/messages/error/import-error.html

Add back check for import error

### DIFF
--- a/requirements.txt
+++ b/requirements.txt
@@ -12,3 +12,4 @@ httpx
 packaging
 charset-normalizer>=3.0.0
 websocket-client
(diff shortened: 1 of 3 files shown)


requirements.txt
```diff
@@ -12,3 +12,4 @@ httpx
 packaging
 charset-normalizer>=3.0.0
 websocket-client
+psutil
```
huchenlei, line 15: psutil is specified in pyproject.toml but not in requirements.txt

PR #139, opened 2024-07-24
[DRAFT] Support new ComfyUI-Manager

ltdrdata:
requires `ComfyUI-Manager/feat/cnr`

* fixed: install
* added: show-versions

### DIFF
--- a/comfy_cli/command/custom_nodes/command.py
+++ b/comfy_cli/command/custom_nodes/command.py
@@ -497,10 +497,15 @@ def update_node_id_cache():
     config_manager = ConfigManager()
     workspace_path = workspace_manager.workspace_path
 
-    cm_cli_path = os.path.join(
-        workspace_path, "custom_nodes", "ComfyUI-Manager", "cm-cli.py"
-    )
+    cm_cli_path = workspace_manager.get_cm_cli_path()
 
+    if cm_cli_path is None:
+        print(
+            f"\n[bold red]ComfyUI-Manager not found[/bold red]\n",
+            file=sys.stderr,
+        )
+        raise typer.Exit(code=1)
+        
     tmp_path = os.path.join(config_manager.get_config_path(), "tmp")
     if not os.path.exists(tmp_path):
         os.makedirs(tmp_path)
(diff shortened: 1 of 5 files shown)


comfy_cli/command/custom_nodes/command.py
```diff
@@ -824,92 +859,6 @@ def display_all_nodes():
     )
 
 
-@app.command(
-    "registry-install",
-    help="Install a node from the registry",
-    hidden=True,
-)
-@tracking.track_command("node")
-def registry_install(
```
james03160927, line 833: `registry_install` -> we would need to update the commands inside the registry_web repository https://github.com/Comfy-Org/registry-web/blob/23948cfe4faa73f640a48139479a718a2d561ccd/components/nodes/NodeDetails.tsx#L209. 

This should be replaced with a new command. `node install <node_id>@<version>`

comfy_cli/command/custom_nodes/command.py
```diff
@@ -630,6 +635,36 @@ def fix(
     execute_cm_cli(["fix"] + nodes, channel, mode)
 
 
+@app.command(
+    "show-versions",
+    help="Show the list of available versions for the target custom node.",
+)
+@tracking.track_command("node")
+def show_versions(
+    node_name: str,
```
james03160927, line 644: `node_name` or `node_id`? Should we use `node_id` here?

comfy_cli/workspace_manager.py
```diff
@@ -281,21 +281,51 @@ def get_comfyui_manager_path(self):
         if self.workspace_path is None:
             return None
 
-        # To check more robustly, verify up to the `.git` path.
-        manager_path = os.path.join(
-            self.workspace_path, "custom_nodes", "ComfyUI-Manager"
+        cached_manager_path = self.config_manager.get(
```
james03160927, line 284: Previously, we knew the absolute path of the comfyUI manager. With the new changes, the comfyUI manager directory name would change depending on which version you've installed (e.g., nightly vs. specific version). We need to figure out the directory of the ComfyUI Manager. 

Let's add comments explaining what this logic is.

comfy_cli/command/install.py
```diff
@@ -214,7 +214,9 @@ def execute(
     if skip_manager:
         print("Skipping installation of ComfyUI-Manager. (by --skip-manager)")
     else:
-        manager_repo_dir = os.path.join(repo_dir, "custom_nodes", "ComfyUI-Manager")
+        manager_repo_dir = os.path.join(
+            repo_dir, "custom_nodes", "comfyui-manager@nightly"
```
james03160927, line 218: We may need to change this. Depending on the version of comfyUI manager, the repo directory would change.

comfy_cli/workspace_manager.py
james03160927: This logic may not work if we are caching the repository name for multiple workspace.

james03160927:
Would have to update `test_node` and e2e test when it's ready for merge.

PR #149, opened 2024-08-02
added support for ruff linting and autoformatting

telamonian:


huchenlei:
Are we nuking the current setup of linting with pylint and autoformatting with black?

telamonian:
> Are we nuking the current setup of linting with pylint and autoformatting with black?

@huchenlei That's the idea. ruff alone can handle both linting and formatting (it has tools for both). It should also support more features/rules and be much more flexible than black

telamonian:
@SocketSecurity ignore pypi/ruff@0.6.1

telamonian:
No need to worry about the failure of the ruff_check GA, that gets fixed when the initial ruff lint/format is applied in #157

PR #156, opened 2024-08-16
Ensure GPU URL before writing override configuration

yoland68:


yoland68:
Hmm, I must have the wrong linter setup locally
What are we on now? ruff?

yoland68:
We still have this: https://github.com/Comfy-Org/comfy-cli/blob/main/.pre-commit-config.yaml

telamonian:
yeah, I wasn't using pre-commit. So your version of the `uv.py` file should be the more "correct" one. I'll just pull it in

PR #158, opened 2024-08-20
Add builder for standalone python for comfy

telamonian:
Adds a command `comfy-cli standalone` that will:

1. grab an appropriate standalone python interpreter from https://github.com/indygreg/python-build-standalone
2. determine requirements by analyzing an existing comfyui install and then recompile the requirements W.R.T. the version/platform/whatever of the standalone python interpreter
3. fetch all comfyui requirements for the standalone python interpreter by doing one of
  3a. install all requirements to the standalone's internal site-packages directory
  3b. precache the package artifacts of all requirements
  3c. fetch (if exists) or build the wheels of all requirements
4. create a `.tgz` tarball containing the standalone interpreter and the requirements

Code exists to run all of 3a-c, though currently the `standalone` command itself only supports 3b.


comfy_cli/utils.py
```diff
@@ -78,3 +95,23 @@ def f(incomplete: str) -> list[str]:
         return [opt for opt in opts if opt.startswith(incomplete)]
 
     return f
+
+def download_progress(url, fname, cwd=".", allow_redirects=True):
+    """download url to local file fname and show a progress bar.
+    See https://stackoverflow.com/q/37573483"""
```
robinjhuang, line 101: Typer already has a progress bar I believe: 

https://typer.tiangolo.com/tutorial/progressbar/

telamonian: hmm, this application of the `tqdm` progress bar turns out to be particularly cursed to replace with the one from `rich`. And technically `tqdm` isn't a new requirement since it's also needed by comfyui core. But I'll give replacing it a shot.

telamonian: done! Turned out to be easier than I thought it would be

telamonian:
This PR is now out of draft status, is more-or-less feature complete, and is ready for testing. Especially if you have a windows machine, please checkout this PR branch and try out the new `standalone` command.

There is still a ton of cleanup to do, and any suggestions are welcome

PR #161, opened 2024-08-24
Fixed: MAC_M_SERIES enum not found

robinjhuang:
Because Resolve_Gpu converts a string to a GPU_Option, we need to use the exact same string as upper and lowercase. `gpu.uppper` converts the previous mac_m_series into `MAC_M_SERIES` which was a non-existent GPU_OPTION.

```
@staticmethod
    def Resolve_Gpu(gpu: Union[GPU_OPTION, str, None]):
        if gpu is None:
            try:
                tver = metadata.version("torch")
                if "+cu" in tver:
                    return GPU_OPTION.NVIDIA
                elif "+rocm" in tver:
                    return GPU_OPTION.AMD
                else:
                    return None
            except metadata.PackageNotFoundError:
                return None
        elif isinstance(gpu, str):
            return GPU_OPTION[gpu.upper()]
        else:
            return gpu
            ```

### DIFF
--- a/comfy_cli/cmdline.py
+++ b/comfy_cli/cmdline.py
@@ -275,14 +275,14 @@ def install(
     elif amd:
         gpu = GPU_OPTION.AMD
     elif m_series:
-        gpu = GPU_OPTION.M_SERIES
+        gpu = GPU_OPTION.MAC_M_SERIES
     elif intel_arc:
         gpu = GPU_OPTION.INTEL_ARC
     else:
         if platform == constants.OS.MACOS:
             gpu = ui.prompt_select_enum(
                 "What type of Mac do you have?",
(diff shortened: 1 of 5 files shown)

### CODE REVIEW

**Comment by robinjhuang** [comment]
@telamonian Does gpu in `uv.py` class need to be a Union type of str and the `GPU_OPTION` enum? That's the root cause of this uncaught bug. What if we just make the `gpu` value typed as `GPU_OPTION`?

**Comment by telamonian** [comment]
@robinjhuang You make a good point. I think I made that change to make testing during development easier. I often import functions into a notebook cell to mess with them as I develop, eg:

```python
from comfy_cli.uv import DependencyCompiler
dp = DependencyCompiler(gpu="amd", ...)
```

But there shouldn't be any need for it in the final product. Alright, let me make those changes and I'll add them to this PR

PR #169, opened 2024-08-29
Adding mac / windows specific tests for standalone command.

robinjhuang:


robinjhuang:
Windows `standalone --rehydrate` is caused by https://github.com/astral-sh/uv/issues/1368

Issue #175, opened 2024-09-08
Containerized execution hangs because of update checks


jason-weirather:
**Describe the bug**
While executing comfy-cli in a docker container, execution hangs when the code calls `comfy_cli.update`'s `check_for_updates` function.

**To Reproduce**
1. Install comfy-cli in a docker container
2. Disable tracking in comfy-cli since its intended to run in a container without phoning out
3. Use `comfy launch` to spin up a server in the background  (`--listen` in the args should be set to 0.0.0.0 in docker)

**Expected behavior**
Upon running `comfy launch` the server should start up and run, indicating a background server is up and return to a prompt.

**Additional context**
I have succeeded to get comfy-cli to work in a container by modifying the code to not check for updates in commands that have nothing to do with updating.

My Dockerfile that does succeed with the modified comfy-cli

https://github.com/jason-weirather/comfyui-image-api/commits/main/Dockerfile

My branch of the fork where I've gotten it working by removing a few calls checking for updates from the `comfy launch` and `comfy env` commands.

https://github.com/jason-weirather/comfy-cli/tree/fixhangbug/comfy_cli

robinjhuang:
Can you share the error message you see? Does it halt the comfy launch?

jason-weirather:
> Can you share the error message you see? Does it halt the comfy launch?

Theres no error message.  It just hangs and halts the launch if it cannot reach out to talk to pypi.

robinjhuang:
How about we make the request timeout 3 seconds

Sent via Superhuman iOS ***@***.***>

jason-weirather:
> How about we make the request timeout 3 seconds 

I think it sounds great solution to the hang bug.  Should I make a PR request that implements this?  

In the future, I think it would be good to add a command line option for an offline mode that could save that time and skip bonking pypi.  A few seconds if you are spinning up dockers as like a lambda function thing is still not great, but thats something I could raise in a feature request rather than bug fix.

robinjhuang:
Please make a PR. Having an offline mode makes sense as well! Please make
another issue for that and we will get to it.

Sent via Superhuman iOS ***@***.***>

PR #176, opened 2024-09-08
Removing pypi `check_for_updates` from commands where it halts offline execution

jason-weirather:
Checking the version of comfy-cli in commands where internet-access is not necessary, such as `comfy launch` and `comfy env` causes comfy-cli to hang in a docker container (or offline) environment.

I opened an issue to describe this
https://github.com/Comfy-Org/comfy-cli/issues/175

This PR removes just those calls to `check_for_updates()` that ping pypi each time a `comfy launch` is run or a `comfy env` is checked where offline operation can otherwise work perfectly.

### DIFF
--- a/comfy_cli/cmdline.py
+++ b/comfy_cli/cmdline.py
@@ -20,7 +20,6 @@ from comfy_cli.config_manager import ConfigManager
 from comfy_cli.constants import GPU_OPTION, CUDAVersion
 from comfy_cli.env_checker import EnvChecker
 from comfy_cli.standalone import StandalonePython
-#from comfy_cli.update import check_for_updates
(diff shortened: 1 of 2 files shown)


robinjhuang:
Can we change the check_for_update function to fail gracefully instead?

jason-weirather:
> Can we change the check_for_update function to fail gracefully instead?

I think thats a reasonable solution if its very fast to fail gracefully.  Otherwise we will have every command hanging for seconds while it waits trying to have a conversation with the pypi servers, and that would reduce performance in some deployments.

jason-weirather:
I've submitted a new PR following @robinjhuang suggestion to add a timeout and fail quietly.

https://github.com/Comfy-Org/comfy-cli/pull/183

PR #179, opened 2024-09-13
removed patch that forced numpy<2 on windows

telamonian:
now that pytorch has updated beyond the broken 2.4 release, we can now remove the patch that forced numpy<2 on windows. We should test this on windows first before pulling this in


ltdrdata:
This requires some caution. It's not just an issue with torch; there are cases where other dependency packages frequently used in custom nodes can cause serious problems, so testing is necessary.

bigcat88:
Thanks for this PR, we have implemented the changes described in it in a separate PR #297

PR #186, opened 2024-09-16
`Find_Req_Files` now returns at most one requirements file per dir

telamonian:
prioritizes based on the order of names in `DependencyCompiler.reqNames`. This PR is done, but should get a unittest before getting pulled in

ping @yoland68


yoland68:
👍

Issue #210, opened 2024-11-17
`comfy model download` command fails.


oprogramadorreal:
**Describe the bug**
When I use the `comfy model download` command to download a model from Civitai, I frequently encounter an exception, and the download fails to complete. See the image below.

![image](https://github.com/user-attachments/assets/01e78189-c4c9-4c0d-94fc-71182b8261f5)

**To Reproduce**
After installing comfy-cli, run the command:
`comfy model download --url https://civitai.com/models/497255?modelVersionId=574841 --relative-path "./models" --filename "stableDiffusion3SD3_sd3MediumInclClip.safetensors"`

The download is very likely to fail at some point.

**Expected behavior**
I expect the `comfy model download` command to successfully download the model file.

**Additional context**

- comfy-cli: 1.3.1
- Python 3.12.7
- Windows 11

oprogramadorreal:
I've opened the following PR with a suggestion to address this issue: https://github.com/Comfy-Org/comfy-cli/pull/211

oprogramadorreal:
I noticed that the issue occurs only with CivitAI URLs (not HuggingFace URLs, for example).

robinjhuang:
This is because the Civitai model requires you to login.

PR #211, opened 2024-11-17
Fixes the 'read operation timed out' exception.

oprogramadorreal:
When using the 'comfy model download' command to download large models from Civitai, an exception with the message 'The read operation timed out' is frequently thrown. This change prevents the exception from occurring.

This is a proposal to fix the issue: https://github.com/Comfy-Org/comfy-cli/issues/210


oprogramadorreal:
I noticed that the issue occurs only with CivitAI URLs (not HuggingFace URLs, for example). This approach disables the 'read' timeout for anyone calling the 'download_file' function, which is not ideal. I will close it and provide a more specific solution.

PR #214, opened 2024-11-19
add '--depth=1' to speedup git clone speed

icedfish:
git clone --help

      --depth <depth>
           Create a shallow clone with a history truncated to the specified number of commits. Implies --single-branch
           unless --no-single-branch is given to fetch the histories near the tips of all branches. If you want to
           clone submodules shallowly, also pass --shallow-submodules.


robinjhuang:
Will this prevent installations of specific git tags? eg We allow installing stable versions of comfy as well as specific tags.

icedfish:
Yes, you are right, sorry for not considering this situation.

PR #224, opened 2024-12-29
[Registry] Publish all files tracked by git

robinjhuang:
Previously, we checked gitignore for files that should not be published. However, some .gitignores were out of sync with files that are already tracked by git. Simplify the implementation by calling `git ls-files`


robinjhuang:
Suggestion by @webfiltered , we can use `git ls-files`

PR #242, opened 2025-02-25
Prevent gzip when downloading models @ models.py

ansonkao:
gzip breaks Content-Length and makes downloads slower.
Content-Length is expected by the progress-bar logger in download_file()


ansonkao:
Confirming this resolved an issue I had with a CDN I was downloading files from

bigcat88:
Should be fixed in main.

PR #244, opened 2025-03-01
fix: unable get ComfyUI-Manager path when its folder name is lowercase in case-sensitive file system

jimlee2048:
ComfyUI-Manager now [suggests to install in lowercase folder name](https://github.com/ltdrdata/ComfyUI-Manager#installation) (also its registry node name). 
CleanShot 2025-03-01 at 15 06 45@2x

However, the current logic only use the name `ComfyUI-Manager` to get path, which would ignores the lowercase name `comfyui-manager` in case-sensitive filesystems.

In addition, I changed the default ComfyUI-Manager installation path to use lowercase folder names as suggested in its `README.md`.


polarathene:
Related: https://github.com/Comfy-Org/comfy-cli/pull/233

jimlee2048:
close stale pr

Issue #258, opened 2025-04-15
Multiple background server for multi GPUs?


nvdamien:
**Describe the solution you'd like**
The CLI is checking is a server is already running in the background here: https://github.com/Comfy-Org/comfy-cli/blob/main/comfy_cli/command/launch.py#L143

But I want to run one server for each GPU (on different port). Like, I have 1 node (H100) with 8 GPUs, I would run 8 servers in the background with different port (8188, 8288, 8388, 8488, 8588, 8688, 8788, 8888) and the extra GPU arg to select the GPU I want to run. 

**Describe alternatives you've considered**
Can we add a flag to "force" to be able to run multiple servers?
Or remove the check completely, and add some arg to specify what port we want to stop when we run "comfy stop"

I can do a MR if you tell me what direction we want to take here.

mattmillerai:
Closing as stale. The single-background-server model is intentional for `comfy launch --background`'s process tracking. For a multi-GPU setup you can run multiple servers directly today — each on its own port with the appropriate GPU/CUDA selection — without the CLI's background bookkeeping.

If first-class multi-server management is important to your workflow, please open a focused feature request. Reopen if needed.

PR #260, opened 2025-04-25
Add multi gpus feature

nvdamien:
Related to https://github.com/Comfy-Org/comfy-cli/issues/258

Comfyui-cli uses a config file `config.ini` to save and check if a Comfyui server was started:

```
[DEFAULT]
enable_tracking = False
recent_workspace = /home/user/Documents/comfyui/ComfyUI
background = ('127.0.0.1', '8188', 1131305)
```

A new command line flag ("--name") was added to the "launch" and "stop" command. When "--name" is added, this is adding a new section into the `config.ini`.

For example:

```
comfy --launch --name GPU_0 -- --port 8188 --cuda-device 0
comfy --launch --name GPU_1 -- --port 8288 --cuda-device 1
```

Will give in the `config.ini`:

```
[DEFAULT]
enable_tracking = False
recent_workspace = /home/user/Documents/comfyui/ComfyUI

[GPU_0]
background = ('127.0.0.1', '8188', 1131305)

[GPU_1]
background = ('127.0.0.1', '8288', 1132502)
```

From here, we can stop a server using the same `--name` flag:

```
comfy stop --name GPU_0
```

Will stop the server `GPU_0` and remove the session `GPU_0` from the `config.ini`

This feature let us starts multiple servers on the same machine that has multiple GPUs.

### DIFF
--- a/comfy_cli/cmdline.py
+++ b/comfy_cli/cmdline.py
@@ -454,7 +454,7 @@ def validate_comfyui(_env_checker):
 @tracking.track_command()
 def stop(
     name: Annotated[
-        str | None, typer.Option(help="If running multiple servers, a name should be specified for each of them")
+        Optional[str], typer.Option(help="If running multiple servers, a name should be specified for each of them")
     ] = None,
 ):
     if name is not None:
--- a/comfy_cli/cmdline.py
+++ b/comfy_cli/cmdline.py
@@ -453,7 +453,9 @@ def validate_comfyui(_env_checker):
 @app.command(help="Stop background ComfyUI: ?[--name str]")
 @tracking.track_command()
 def stop(
-        name: Annotated[str | None, typer.Option(help="If running multiple servers, a name should be specified for each of them")] = None
+    name: Annotated[
+        str | None, typer.Option(help="If running multiple servers, a name should be specified for each of them")
+    ] = None,
 ):
     if name is not None:
         ConfigManager().load(name=name)
@@ -480,7 +482,9 @@ def stop(
 def launch(
     background: Annotated[bool, typer.Option(help="Launch ComfyUI in background")] = False,
     extra: List[str] = typer.Argument(None),
-    name: Annotated[str | None, typer.Option(help="If running multiple servers, a name should be specified for each of them")] = None
+    name: Annotated[
+        str | None, typer.Option(help="If running multiple servers, a name should be specified for each of them")
+    ] = None,
 ):
     launch_command(background, extra, name=name)
 
--- a/comfy_cli/cmdline.py
+++ b/comfy_cli/cmdline.py
@@ -450,10 +450,14 @@ def validate_comfyui(_env_checker):
         raise typer.Exit(code=1)
 
 
-@app.command(help="Stop background ComfyUI")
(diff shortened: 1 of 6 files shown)


mattmillerai:
Thanks for this, @nvdamien, and apologies it sat open this long.

Closing it, for two reasons:

**1. The parent issue was resolved the other way.** #258 was closed as stale in July: the single-background-server model is intentional for `comfy launch --background`'s process tracking, so named multi-session bookkeeping in `config.ini` isn't a direction we're taking.

**2. The code underneath has moved too far to rebase.** Since April 2025:

- `comfy_cli/command/run.py` no longer exists. It was split into `comfy_cli/command/run/{execution,watcher,loader,preflight}.py`
- `ConfigManager.fill_print_env()` is gone, replaced by `get_env_data()` / `get_data()` (the latter backs `comfy env --json`)
- `remove_background()` now delegates to `clear_background_process()`, which deliberately preserves the logfile pointer so a post-mortem `comfy logs` still works
- `launch.py` was substantially rewritten (JSON output envelopes, `--port` validation, per-port log paths, telemetry drain on hard exit)

All six touched files conflict, and a rebase would be closer to a rewrite than a merge.

**Two of your incidental fixes already landed independently**, so that work wasn't lost:

- Closing the websocket after execution, now at `comfy_cli/command/run/execution.py:54-55`
- Serializing config writes. We ended up with an in-house `comfy_cli/locking.py` (fcntl on Unix, msvcrt on Windows) rather than a `filelock` dependency

**One piece is still valid and I'd like to take it**: the `escape()` calls on the background redirector output. `launch.py:188` and `:193` still pass raw ComfyUI stdout/stderr through Rich markup, so a log line containing something like `[/foo]` still raises `MarkupError` or silently swallows text. I'll open that as a standalone PR crediting you.

**For the multi-GPU use case in the meantime**: `comfy stop --port <N>` now stops any local ComfyUI on that port, including servers comfy-cli didn't start, and background logs are port-suffixed (`user/comfyui_<port>.log`). So you can launch one server per GPU directly and still manage them by port. The only thing in the way is the single-background-server guard in `background_launch`, which the `--port` path sidesteps.

If first-class multi-server management matters to your workflow, please do open a focused feature request describing the lifecycle you want (list / status / stop by name), and we'll scope it against the current architecture rather than the 2025 one.

Created by Claude

mattmillerai:
Following up as promised: the escaping fix from this PR is now open as https://github.com/Comfy-Org/comfy-cli/pull/713, with you as co-author.

It turned out to be worse than either of us described. Passing ComfyUI's output through Rich markup does not just risk a `MarkupError`, it also *silently deletes* bracketed text, so `Progress: [####  ] 50%` was being logged as `Progress:  50%` and tracebacks were losing quoted paths. And because the `MarkupError` kills the redirector thread, `comfyui_<port>.log` stopped being written from that line onward, truncating the log at exactly the moment something was going wrong.

Your `escape()` was the right call. The PR adds a regression test around it (verified failing against the unescaped version) and a note on why `escape` rather than the repo's `sanitize_markup`: this is a capture path, so ANSI is preserved verbatim and sanitizing happens at render time instead.

Thanks again for spotting it.

Created by Claude

Issue #264, opened 2025-05-12
Fails to output help with error: `TypeError: Parameter.make_metavar() missing 1 required positional argument: 'ctx'`


polarathene:
**Describe the bug**

I have tried to install Comfy CLI via `pip` package and via cloning this repo. Basic functionality to get a help output is failing.

**To Reproduce**

```console
# Using either Ubuntu or Fedora:
$ docker run --rm -it --env DEBIAN_FRONTEND=noninteractive ubuntu:24.04
$ apt-get -qq update && apt-get -qq install pipx git
$ python3 --version
Python 3.12.3

$ docker run --rm -it fedora:42
$ dnf install -yq pipx git-core
$ python --version
Python 3.13.3

$ pipx install comfy-cli
```

There are 3 different commands to output help (_`comfy-cli`, `comfy-cli help`, `comfy-cli --help`_) that all have different Traceback failures but output the same error at the end:

```
TypeError: Parameter.make_metavar() missing 1 required positional argument: 'ctx'
```

<details>
<summary>CLI outputs with errors/tracebacks</summary>

```console
$ comfy-cli

warn: The recent workspace None is not a valid ComfyUI path.
Welcome to Comfy CLI!: https://github.com/Comfy-Org/comfy-cli

 Usage: comfy-cli [OPTIONS] COMMAND [ARGS]...

╭───────────────────────────────────────────────────────────────────────────────────────── Traceback (most recent call last) ─────────────────────────────────────────────────────────────────────────────────────────╮
│ /root/.local/share/pipx/venvs/comfy-cli/lib64/python3.13/site-packages/comfy_cli/cmdline.py:129 in entry                                                                                                            │
│                                                                                                                                                                                                                     │
│   126 │                                                                                        ╭───────────────────────────── locals ─────────────────────────────╮                                                 │
│   127 │   if ctx.invoked_subcommand is None:                                                   │              ctx = <click.core.Context object at 0x7ff7e0e29160> │                                                 │
│   128 │   │   rprint("[bold yellow]Welcome to Comfy CLI![/bold yellow]: https://github.com/Com │ enable_telemetry = False                                         │                                                 │
│ ❱ 129 │   │   rprint(ctx.get_help())                                                           │             here = None                                          │                                                 │
│   130 │   │   ctx.exit()                                                                       │           recent = None                                          │                                                 │
│   131 │                                                                                        │      skip_prompt = False                                         │                                                 │
[...]

PR #265, opened 2025-05-13
Fixed: Pin version of click dependency.

robinjhuang:
https://github.com/Comfy-Org/comfy-cli/issues/264

### DIFF
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -44,6 +44,7 @@ dependencies = [
   "ruff",
   "semver~=3.0.2",
   "cookiecutter",
+  "click<=8.1.8"
 ]
 
 [project.optional-dependencies]


pyproject.toml
```diff
@@ -44,6 +44,7 @@ dependencies = [
   "ruff",
   "semver~=3.0.2",
   "cookiecutter",
+  "click<=8.1.8"
```
polarathene, line 47: Assuming comments here are valid in this file:

```suggestion
  # Temporarily pin until `typer` is compatible:
  # https://github.com/Comfy-Org/comfy-cli/issues/264#issuecomment-2871116559
  "click<=8.1.8"
```

Better for maintainers to be aware that this isn't a direct dep required by `comfy-cli`, and that it can be removed at a later date.

Ref: https://github.com/Comfy-Org/comfy-cli/issues/264#issuecomment-2878118243

PR #270, opened 2025-05-16
Added: publish all files in includes field in pyproject.toml

jtydhr88:
Added a parameter in pyproject.toml, includes = [], which contains folder paths, for example includes = ['/js', '/dist']. 
The purpose of doing this is, in certain situations, such as after packaging through npm, the output directory might be added to gitignore, but when publishing to the registry, It should be included.


huchenlei (approved):

robinjhuang:
LGTM!

PR #286, opened 2025-06-18
feat: refactor env command to use ui.display_table for consistent table display

neverbiasu:
## What changed
Updated the `comfy env` command to use our existing `ui.display_table()` function instead of manually creating Rich Table objects.

## Why
- Keeps table rendering consistent across the codebase
- Follows the pattern we already use elsewhere
- Makes the code cleaner by separating data from presentation

## Changes made
- Modified `EnvChecker.fill_print_table()` to return data tuples instead of creating a table
- Updated `WorkspaceManager.fill_print_table()` to return data tuples too
- Changed the env command to collect all data first, then render it with `ui.display_table()`

The output looks exactly the same to users.
| after | before |
| ----- | -------| 
| after | before |

### DIFF
--- a/comfy_cli/cmdline.py
+++ b/comfy_cli/cmdline.py
@@ -527,9 +527,8 @@ def which():
 @tracking.track_command()
 def env():
     check_for_updates()
-    _env_checker = EnvChecker()
     
-    env_data = _env_checker.fill_print_table()
+    env_data = EnvChecker().fill_print_table()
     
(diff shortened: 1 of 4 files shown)


bigcat88 (changes requested):
Overall the PR is good (tested locally), but some small changes need to be made.

bigcat88:
thanks for this contribution

Arcitec:
Nice cleanup and unification. Good job! :+1:

PR #299, opened 2025-08-02
fix: use constants in ConfigManager

Arcitec:
Spotted this issue while looking at some recent changes.

Fixes ConfigManager to always use the defined constants, so that it won't break if the constant values are changed in the future.

I also ripgrep'd the rest of the codebase and did not come up with any other locations where the strings are used instead of the constants, so it seems fine now.

---

I was unsure about this location though:

https://github.com/Comfy-Org/comfy-cli/blob/6be00e4a84ce9279e98fa8a0b29c8a5dd97a749a/comfy_cli/workspace_manager.py#L135-L139

It **doesn't** seem like it should be using `CONTEXT_KEY_RECENT = "recent"` instead of hardcoding. Perhaps someone can confirm that the workspace manager keys are indeed totally separate from the config keys.

### DIFF
--- a/comfy_cli/config_manager.py
+++ b/comfy_cli/config_manager.py
@@ -75,8 +75,8 @@ class ConfigManager(object):
         if not os.path.exists(tmp_path):
             os.makedirs(tmp_path)
 
-        if "background" in self.config["DEFAULT"]:
-            bg_info = self.config["DEFAULT"]["background"].strip("()").split(",")
+        if constants.CONFIG_KEY_BACKGROUND in self.config["DEFAULT"]:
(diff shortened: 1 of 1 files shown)


bigcat88 (approved):
> Perhaps someone can confirm that the workspace manager keys are indeed totally separate from the config keys.

I guess `WorkspaceType` is related only to path of workspace.

It was added by this [commit](https://github.com/Comfy-Org/comfy-cli/commit/500cfb0c6647bbd739bd7b11dc2c90a0caeb9cc6)

so no need to replace `recent` there with `CONTEXT_KEY_RECENT` (at least for now)

---

Regarding this PR - I think we can replace the values with constants, logically everything seems fine.

PR #323, opened 2025-09-30
fix(CivitAI): correct parse URL with query params

### LINKED ISSUE #322: CivitAI download with token broken
**Describe the bug**
Since comfy-cli 1.5.0+ the CivitAI download with token is broken.

**To Reproduce**
- install comfy-cli 1.5.1
- run command `comfy --skip-prompt model download --url '{url}' --relative-path 'models/{local_dir}' --filename '{filename}' --set-civitai-api-token '{token}'` with any url that requires login
- Download fails with exception:
```
DownloadException: Failed to download file.
Unauthorized download (401).
The creator of this asset requires you to be logged in to download it
or you can set a CivitAI API token using `comfy model download
--set-civitai-api-token` or via the `CIVITAI_API_TOKEN` environment variable
Traceback (most recent call last):
  File "/pkg/modal/_runtime/container_io_manager.py", line 778, in handle_input_exception
    yield
  File "/pkg/modal/_container_entrypoint.py", line 243, in run_input_sync
    res = io_context.call_finalized_function()
          ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "/pkg/modal/_runtime/container_io_manager.py", line 197, in call_finalized_function
    res = self.finalized_function.callable(*args, **kwargs)
          ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "/root/app.py", line 70, in civitai_dl
    subprocess.run(cmd, shell=True, check=True)
```
(Same exception when setting the token as environment variable `CIVITAI_API_TOKEN` instead of with --set-civitai-api-token)

**Expected behavior**
Download completes without error like it does with comfy-cli 1.4.1 or prior.

bigcat88:
Resolves #322 

Completely rewrote the `check_civitai_url` implementation to use `urlparse` and `parse_qs` functions instead of usual string functions to parse URLs.
Additionally expanded tests to cover the new valid URL cases.

### DIFF
--- a/comfy_cli/command/models/models.py
+++ b/comfy_cli/command/models/models.py
@@ -1,8 +1,9 @@
+import contextlib
 import os
 import pathlib
 import sys
 from typing import Annotated, Optional
-from urllib.parse import unquote, urlparse
+from urllib.parse import parse_qs, unquote, urlparse
 
 import requests
 import typer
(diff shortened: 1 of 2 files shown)


Arcitec:
Yeah. :) Thank you so much for this project. Comfy-CLI is an amazingly useful project, and I appreciate your great work and quality mind. It was fun to collaborate with you and addressing your feedback on the previous PR. It became better when we settled on using custom environment names and I am glad you caught that design mistake, since it could have become a mess with time if we didn't standardize the envs. :+1: :smile_cat:

Issue #328, opened 2025-10-08
`comfy install` failed under Python 3.14


fengsi:
```python
$ comfy install
╭────────────────────────────────────────────────────────────────────────── Traceback (most recent call last) ──────────────────────────────────────────────────────────────────────────╮
│ /Users/si/Projects/ComfyUI/.venv/bin/comfy:10 in <module>                                                                                                                             │
│                                                                                                                                                                                       │
│    7 │   │   sys.argv[0] = sys.argv[0][:-11]                                                  ╭──────────── locals ─────────────╮                                                     │
│    8 │   elif sys.argv[0].endswith(".exe"):                                                   │ sys = <module 'sys' (built-in)> │                                                     │
│    9 │   │   sys.argv[0] = sys.argv[0][:-4]                                                   ╰─────────────────────────────────╯                                                     │
│ ❱ 10 │   sys.exit(main())                                                                                                                                                             │
│   11                                                                                                                                                                                  │
│                                                                                                                                                                                       │
│ /Users/si/Projects/ComfyUI/.venv/lib/python3.14/site-packages/comfy_cli/cmdline.py:35 in main                                                                                         │
│                                                                                                                                                                                       │
│    32                                                                                                                                                                                 │
│    33                                                                                                                                                                                 │
│    34 def main():                                                                                                                                                                     │
│ ❱  35 │   app()                                                                                                                                                                       │
│    36                                                                                                                                                                                 │
│    37                                                                                                                                                                                 │
│    38 class MutuallyExclusiveValidator:                                                                                                                                               │
│                                                                                                                                                                                       │
│ /Users/si/Projects/ComfyUI/.venv/lib/python3.14/site-packages/typer/main.py:325 in __call__                                                                                           │
│                                                                                                                                                                                       │
│ /Users/si/Projects/ComfyUI/.venv/lib/python3.14/site-packages/typer/main.py:308 in __call__                                                                                           │
│                                                                                                                                                                                       │
│ /Users/si/Projects/ComfyUI/.venv/lib/python3.14/site-packages/typer/main.py:347 in get_command                                                                                        │
│                                                                                                                                                                                       │
│ /Users/si/Projects/ComfyUI/.venv/lib/python3.14/site-packages/typer/main.py:329 in get_group                                                                                          │
│                                                                                                                                                                                       │
│ /Users/si/Projects/ComfyUI/.venv/lib/python3.14/site-packages/typer/main.py:474 in get_group_from_info                                                                                │
[...]

PR #329, opened 2025-10-08
Fix Python 3.14 compatibility and the built-in `list` redefinition

fengsi:
This addresses #328.

### DIFF
--- a/comfy_cli/command/models/models.py
+++ b/comfy_cli/command/models/models.py
@@ -2,7 +2,7 @@ import contextlib
 import os
 import pathlib
 import sys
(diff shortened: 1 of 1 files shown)


comfy_cli/command/models/models.py
```diff
@@ -394,6 +394,11 @@ def remove(
         typer.echo("Deletion canceled.")
 
 
+def list_models(path: pathlib.Path) -> list:
```
fengsi, line 397: No functional change here, but if we were to redefine the built-in `list`, we'll need to make sure that any possible references to `list` should happen before the `def list()` function.

PR #351, opened 2026-01-11
Fix ComfyUI repo URLs

fengsi:
Added the new ComfyUI URLs, and changed the `list` to `set` for better `in` performance.


bigcat88 (approved):
looks good

Issue #335, opened 2025-10-29
--pr cannot be used with --version or --commit


marchuk-vlad:
**Describe the bug**
The bug is [here](https://github.com/Comfy-Org/comfy-cli/blob/v1.5.2/comfy_cli/cmdline.py#L347).
According to operator precedence, the current condition means
`if (pr and version not in {None, "nightly"}) or commit`,
but it should be
`if pr and (version not in {None, "nightly"} or commit)`.
As a result, specifying --commit causes a crash.

**To Reproduce**
`comfy --skip-prompt --workspace ./confy2/ install --version nightly  --commit 93292bc450dd291925c45adea00ebedb8a3209ef  --manager-commit 3.31.12 --cpu`
--pr cannot be used with --version or --commit

**Expected behavior**
Guess :)

**Nice to have**
- [ ] Terminal output
- [ ] Screenshots

**Additional context**
I think no needs.

PR #365, opened 2026-02-17
fix: correct operator precedence in install command validation

Xyc2016:
Fixed issue #335: --pr cannot be used with --version or --commit

The condition "pr and version not in {None, \"nightly\"} or commit" was evaluated incorrectly due to operator precedence. Added parentheses to ensure the logic only triggers error when --pr is actually combined with conflicting flags, not when only --commit is specified.


bigcat88 (approved):
PR looks good - merging.

PR #438, opened 2026-04-24
feat(registry): resolve PEP 621 dynamic version from [tool.comfy.version].path

### LINKED ISSUE #294: Add support to `pyproject.toml`  dynamic values, i.e. "version"
The setuptools supports reading the version from the source code itself, avoiding repetition and sync problems, like this:

```
[project]
name = "image-misc"
description = """
Miscellaneous nodes for image manipulation.
Currently just download image with bypass, so you can create workflows including image examples.
No extra dependencies, just an internal module.
"""
dynamic = ["version"]
license = "GPL-3.0-or-later"
dependencies = ["seconohe"]

[project.urls]
Repository = "https://github.com/set-soft/ComfyUI-ImageMisc"

[tool.comfy]
PublisherId = "set-soft"
DisplayName = "Image Misc"

[tool.setuptools.dynamic]
version = {attr = "nodes.__version__"}
```

But when using it `comfy-cli` fails to find the version.

In the specs example it looks like the dependencies can be dynamic. So I think dynamic version should also be supported.

bigcat88:
Fixes #294

## Summary

Custom node authors using `dynamic = ["version"]` in `pyproject.toml` (PEP 621) previously got a silent failure — the version was read as `""` and POSTed to the registry anyway. This PR adds a comfy-native `[tool.comfy.version]` section that tells us where to read the version from, and a publish-time guard so empty versions never reach the registry.

```toml
[project]
name = "my-node"
dynamic = ["version"]

[tool.comfy.version]
path = "my_node/__init__.py"
```

The file is text-read and a single anchored regex extracts `__version__` or `VERSION`. No Python execution — this was the hard constraint agreed with @ltdrdata in the issue thread (env-sensitive, slow, breaks the Manager's "quick scan of installed packs" use case). The schema matches @Lex-DRL's proposal that got a 👍 in the thread.

## Behavior

- **Static `project.version` present** → used as-is (existing behavior, no change).
- **`"version"` in `project.dynamic`** → resolve via `[tool.comfy.version].path`.
- **Any resolution failure** (missing section, absolute path, path escaping project dir, missing file, no regex match) → warn via `typer.echo` and fall back to empty version. This keeps scanning contexts graceful — they never crash on malformed pack configs.
- **Publish-time** → `validate_node_for_publishing()` now exits `1` with a clear message if the resolved version is empty, so we never silently POST an empty version to the registry. This is the behavior change that actually fixes the bug.

## Regex

```python
^(?:__version__|VERSION)
    (?:[\t ]*:[\t ]*[^=\n]+)?   # optional PEP 526 annotation
    [\t ]*=[\t ]*
    (?:"(?P<dq>[^"\n]+)"|'(?P<sq>[^'\n]+)')
```

Start-of-line anchored (MULTILINE) so single-line comments are skipped. Horizontal whitespace only — no cross-line matching. Both quote styles. PEP 526 type annotation tolerated.

**Documented limitations** (unchanged from @Lex-DRL's proposal):
- Only static string literals work — computed `__version__` (f-strings, `get_distribution().version`, etc.) doesn't.
- Only `__version__` and `VERSION` names.
- Matching lines inside multiline string/docstring contexts can be picked up (first-match-wins). Keep version files trivial.

## Non-goals (punted to follow-ups)

- **`[tool.hatch.version]` passthrough**: zero-config for existing hatch users, ~10 LOC. Easy follow-up if the community asks for it; kept out of v1 to avoid coupling to a second schema.
- **`[tool.setuptools.dynamic] attr = "..."`**: requires Python module loading. Explicitly out of scope per the issue discussion.

## Test plan

New tests in `tests/comfy_cli/registry/test_config_parser.py` (12 cases, all using real `tmp_path` files):

- [x] Dynamic version resolved from `__version__` (double-quoted, single-quoted, with PEP 526 annotation).
- [x] Dynamic version resolved from `VERSION` name.
- [x] Commented `# __version__ = "x"` line correctly ignored (`^` anchor).
- [x] First match wins if the file has multiple `__version__` lines.
- [x] Static `project.version` beats `[tool.comfy.version]` when both present (no accidental resolution).
- [x] `dynamic = ["version"]` with no `[tool.comfy.version]` → warn + empty.
- [x] Absolute path → warn + empty.
- [x] Path traversal (`../../etc/passwd`) → warn + empty.
- [x] Missing file → warn + empty.
- [x] File exists but no match → warn + empty.

Plus in `tests/comfy_cli/command/nodes/test_publish.py`:

- [x] `comfy node validate` exits `1` with the helpful error when `project.version` is empty.

CI:
- [x] `pytest tests/comfy_cli/` — 687 passed, 6 skipped (+13 new).
- [x] `ruff check .` and `ruff format --check .` — clean.

## Risk / blast radius

- Static-version path untouched — pure regression test pass for existing users.
- Publish-time empty-version guard is new strictness, but only fires on configurations that were already broken (registry would have received `""`).
- No new dependencies. No API changes to `registry/types.py`.

### DIFF
--- a/comfy_cli/command/custom_nodes/command.py
+++ b/comfy_cli/command/custom_nodes/command.py
@@ -975,6 +975,17 @@ def validate_node_for_publishing():
     # Perform some validation logic here
     typer.echo("Validating node configuration...")
     config = extract_node_configuration()
+    if config is None:
+        raise typer.Exit(code=1)
+
+    if not config.project.version:
+        # Escape `[` chars so rich doesn't parse `[tool.comfy.version]` and
+        # `["version"]` as markup tags; `]` doesn't need escaping.
+        print(
+            "[red]Error: project version is empty. Set `project.version` in pyproject.toml, "
+            r'or configure `\[tool.comfy.version].path` if using `dynamic = \["version"]`.[/red]'
+        )
+        raise typer.Exit(code=1)
 
     # Run security checks first
     typer.echo("Running security checks...")
--- a/comfy_cli/registry/config_parser.py
+++ b/comfy_cli/registry/config_parser.py
@@ -1,4 +1,5 @@
 import os
+import pathlib
 import re
 import subprocess
 from urllib.parse import urlparse, urlunparse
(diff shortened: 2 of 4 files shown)


Lex-DRL:
Thank you! 🙏🏻

Lex-DRL:
Though, when it will be merged and documented, I **strongly** suggest encouraging node authors to keep a version in its own file (`path = "__version.py"` or `path = "__meta.py"`, NOT right there in `__init__.py`) - for the exact reason of this file contents being as simple as possible, with just module-level constants aka static strings there. Again, build tools for PyPI recommend the same.

bigcat88:
> Though, when it will be merged and documented, I **strongly** suggest encouraging node authors to keep a version in its own file (`path = "__version.py"` or `path = "__meta.py"`, NOT right there in `__init__.py`) - for the exact reason of this file contents being as simple as possible, with just module-level constants aka static strings there. Again, build tools for PyPI recommend the same.

Good point!

PR #455, opened 2026-05-17
feat(run): add --json NDJSON output mode for agents

bigcat88:
- Add `--json` flag to `comfy run` that switches stdout to NDJSON (one event per line, terminated with `completed` or `failed`).
- 8 event types and 17 stable `error.kind` values; every event carries `schema_version: 1`.
- ASCII-safe (`ensure_ascii=True`), per-line flushed, no Rich/ANSI on stdout.
- `queued` includes a `nodes` manifest so piped consumers can render per-node UI without parsing the workflow file.
- `node_progress` carries `class_type` and `title` so stateless consumers don't need to track a prior `node_executing`.
- `executed_node_ids` aggregates every node the executor touched, not just nodes that emit the server's `executed` WS message.
- Bounded timeouts on the pre-flight probe, `ws.connect`, `/prompt` POST, and `ws.recv` so the terminal-event guarantee holds even when the server hangs.
- Spec at `docs/json-output.md`; contract tests in `tests/comfy_cli/command/test_run_json.py`.

### DIFF
--- a/comfy_cli/cmdline.py
+++ b/comfy_cli/cmdline.py
@@ -459,7 +459,7 @@ def run(
         typer.Option(help="The port where the ComfyUI instance is running, e.g. 8188."),
     ] = None,
     timeout: Annotated[
-        int | None,
+        int,
         typer.Option(
             help=(
                 "Per-event timeout in seconds: bails out if the server is silent "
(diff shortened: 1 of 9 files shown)


bigcat88:
Additional findings from the Agent that does not have clue about this feature(clear memory):


```
1. local_path: null — still a real gap; agents must GET the URL.
2. First-run telemetry prompt swallows --json. Still a real gap.
3. Discoverability: the --json help block doesn't mention that UI workflows are accepted. Worth a sentence in --help and in docs/json-output.md.
```


Telemetry(point 2) is critical, agent tried 3 times with constant failures..

PR #470, opened 2026-06-13
feat: agent CLI — JSON envelopes, cloud run/jobs, fragments, projects, CQL, structured errors

skishore23:
## Summary

Turns `comfy-cli` into an **agent-first CLI** for ComfyUI — letting an agent (or a human) build, validate, run, and review image/video/audio workflows on a local server or Comfy Cloud, entirely from the terminal.

Three ideas hold it together:

1. **One machine contract** — every command emits the same versioned JSON envelope, and every error carries a registered code plus an actionable hint (an error is a *navigation signal toward the fix*, not just a failure). Agents never parse prose.
2. **A compile model for workflows** — author small typed *fragments* (or `decompose` an existing workflow into them) and `compose` them with YAML *blueprints* into one graph. The folders are source; the compiled workflow is a build artifact you never hand-edit.
3. **Async by default** — submit returns immediately, a detached watcher tracks the job, and any later process resumes from a state file.

## What's in it

- **Machine output** (`output/`) — JSON/NDJSON envelopes, a schema registry, `comfy discover`; auto-JSON on non-TTY. Errors inherit their code's registered hint automatically.
- **Composition** (`fragments.py`, `command/workflow_fragments.py`) — fragments + blueprints, `foreach` fan-out, and the `$asset`/`$item`/`$var`/`$alias` reference algebra. `comfy workflow decompose` is the inverse of `compose`: it projects any workflow into self-documenting fragment source (records where it came from and how to edit it).
- **Validation** (`cql/`) — a pure-Python engine validates a graph against the live server's `object_info` before spending. Combo option types are preserved so the published schema matches what the backend accepts, and enum rejections carry the full, untruncated valid-options list.
- **Execution** (`command/jobs.py`, `jobs_state.py`) — async run, detached watcher, resumable job state, item-named outputs.
- **Project convention** (`project.py`) — the `project/1` layout, `assets push` + content-addressed lock, run journal, and `--where` routing (OAuth-first credentials).
- **`comfy preview`** — render a previewable image from any media: image → thumbnail, video → contact sheet, audio → waveform.
- **Skills** (`skills/`) — bundled `SKILL.md` files that teach agents to drive the CLI (operate · build · debug · present).
- **Install robustness** — `comfy update` / `comfy install` no longer assume the workspace interpreter ships pip; a pip-less uv-managed venv is bootstrapped via uv or `ensurepip` (no-op when pip is present).

## Compatibility

No breaking changes to existing human-facing commands — pretty (TTY) output is unchanged. Everything new is additive: gated behind `--json`, new subcommands, or strictly-more-correct schema/error output.

## How the contract stays honest

AST-scan ratchet tests fail the build if: a command emits without a registered schema; an error code is raised-but-unregistered, registered-but-never-raised, or missing a navigation hint; or credentials are read outside the single resolver.

## Notes for reviewers

Large branch — it lands the full agent-CLI surface, so it's best reviewed by area (the package list above maps to the sections of work). The core domain logic (`fragments.py`, `cql/engine.py`) is pure value-in/value-out and unit-tested in isolation; `command/` shells are thin I/O wrappers.

## Verification

- Full test suite green: **2306 passed, 35 skipped**.
- CI runs build, test, ruff (lint + format), CodeQL, Socket, Codecov, and cross-platform (ubuntu / macOS / windows) + GPU e2e.


tests/e2e/verify_tracking_live.py
luke-mino-altherr, line 1: should this live in the `tests/e2e` dir?

luke-mino-altherr (approved):

tests/e2e/verify_tracking_live.py
skishore23: Good call — moved to `tests/e2e/verify_tracking_live.py` in 52bdb18. Kept the non-`test_` prefix so pytest doesn't collect it: it's a manual, opt-in, env-gated check (real PostHog round-trip, never in CI), run directly via `python tests/e2e/verify_tracking_live.py`.

PR #484, opened 2026-07-02
refactor: extract module-local _emit_http_error helper in models/search.py (BE-2143)

mattmillerai:
## ELI-5

Two spots in `comfy models` (the `list-folders` and `search` commands) had the
*exact same* copy-pasted block for turning an HTTP failure into a nice error:
pick a cloud-vs-local error code, and stuff a truncated, decoded copy of the
server's response body into the error details. This PR lifts that duplicated
block into one small helper next to them, so the idiom lives in one place.

## What changed

- New module-local helper `_emit_http_error(e, *, renderer, target, message, hint)`
  in `comfy_cli/command/models/search.py`. It reproduces the shared idiom
  verbatim: `code = cloud_http_error if target.is_cloud else server_not_running`,
  and `details={"status": e.code, "body": (e.read() or b"")[:1000].decode("utf-8","replace")}`.
  It's annotated `-> NoReturn` (it always ends in `raise typer.Exit(code=1) from e`).
- `list_folders_cmd` and `search_cmd` HTTPError handlers now call the helper.
  `message`/`hint` are the only per-caller differences, so they stay at the call
  sites.
- Added regression tests covering both extracted paths — cloud and local
  branches for `list-folders`, plus the `search` cloud path — asserting the
  error code and the truncated/decoded `details.body`. No such tests existed
  before.

## Scope (deliberately narrow)

Per the ticket's corrected recommendation (downgraded from the original,
overstated finding), this is a **module-local** extraction shared by the **two**
sites that actually share the body-decode idiom. The other HTTPError handlers do
NOT share it and are intentionally left untouched:

- `list-folder` (404-special-cased, `details={status, folder}`, no body decode)
- `show` (fixed `cloud_http_error`, `details={status}`, no body decode)
- the sibling `URLError`/`OSError`/`JSONDecodeError` handlers (no body decode)

No cross-module `emit_http_error` was built — the `transfer.py` / `project.py` /
`run/execution.py` sites cited in the original finding don't share the idiom and
would be forced through parameters they don't want.

## Verification

- `pytest tests/comfy_cli` — 2218 passed, 12 skipped.
- `ruff check` + `ruff format --check` clean on both changed files.

## Judgment call

- Pure refactor, behavior-preserving; no functional change to emitted errors.
- Followed the repo's own PR/commit convention of the `(BE-####)` suffix
  (matches recent history, e.g. `(BE-2142)`, `(BE-992)`).

### DIFF
--- a/comfy_cli/command/models/search.py
+++ b/comfy_cli/command/models/search.py
@@ -29,7 +29,7 @@ import re
 import urllib.error
 import urllib.parse
 import urllib.request
-from typing import Annotated, Any
+from typing import Annotated, Any, NoReturn
 
 import typer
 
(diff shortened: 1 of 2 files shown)


bigcat88 (approved):
Tested. Looks good!

PR #487, opened 2026-07-02
refactor: consolidate refuse-redirect handlers into shared comfy_cli/http.py (BE-2152)

mattmillerai:
## ELI-5

Five different files each had their own private copy of the same little safety class that says *"if a server tries to redirect this authenticated request somewhere else, refuse — don't hand over the `Bearer` token / `X-API-Key` to wherever it points."* Five identical copies means a future security fix could land in one and be forgotten in the other four. This PR replaces all five with **one** audited definition in a new `comfy_cli/http.py`, so there's a single source of truth.

## What changed

- **New module `comfy_cli/http.py`** — one `NoRedirectHandler` (stdlib-only, so no import cycles). It refuses to follow any 30x redirect and raises a clear `HTTPError` instead. The refusal `message` is a constructor arg so call sites keep their own wording.
- **Swapped in at all 5 sites**, deleting each local copy:
  - `comfy_client.py` (`_NoRedirectHandler`)
  - `cql/engine.py` (`_NoRedirectHandler`)
  - `cql/loader.py` (`_NoRedirect`)
  - `cloud/oauth.py` (`_NoRedirect`)
  - `command/transfer.py` (`_NoRedirectHandler`) — passes `"redirect refused (auth leak prevention)"` to preserve its original message byte-for-byte.
- **New tests** `tests/comfy_cli/test_http.py` — assert every redirect status (301/302/303/307/308) raises `HTTPError` with the code preserved, the default `"redirect refused"` message, and a custom message passed through.

## Deliberately NOT touched

- `command/transfer.py`'s `_DownloadRedirectHandler` / `_DOWNLOAD_OPENER` — that one *intentionally follows* redirects while stripping auth headers (for cloud `/api/view` → signed-GCS 302s). It is a distinct handler, not a duplicate. Left entirely alone.
- Did not consolidate the separate `_LOOPBACK_HOSTS` duplication — out of scope for this ticket.

## Behavior

Pure refactor — no behavior change. All four non-transfer sites already used the message `"redirect refused"` (now the default); transfer's longer message is preserved via the constructor arg. No import removals were needed (every file that kept `import urllib.error` still uses it elsewhere).

## Verification

- `ruff check` + `ruff format --check` clean (ruff v0.15.15, matching the repo's pre-commit pin).
- Full `pytest` suite: **2324 passed, 36 skipped**.

## Judgment calls

- Dropped the per-site rationale comments/docstrings in favor of the one consolidated docstring in `http.py` (the ticket calls for deleting the local classes; the shared docstring captures the auth-leak/SSRF reasoning).

### DIFF
--- a/comfy_cli/cloud/oauth.py
+++ b/comfy_cli/cloud/oauth.py
@@ -52,6 +52,7 @@ from comfy_cli.cloud import (
     CLIENT_NAME,
     get_base_url,
 )
+from comfy_cli.http import NoRedirectHandler
 
 # ---------------------------------------------------------------------------
 # Error types — caller maps these to renderer.error(code=...) codes.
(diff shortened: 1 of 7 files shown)


bigcat88 (approved):
Tested, LGTM.

PR #525, opened 2026-07-17
refactor: remove 8 dead symbols + vestigial where.py re-export (BE-3268)

mattmillerai:
## ELI-5

There were 8 functions/classes in the codebase that nothing ever called, plus one leftover "re-export" line — like a light switch wired to nothing. This PR deletes them. Because nothing referenced them, nothing changes at runtime; the code just gets smaller and less confusing. Matches the merged precedent of #499/#504.

## What & why

Deletes 8 zero-reference symbols plus a vestigial `where.py` re-export line. Each was verified dead by full-repo grep (definition-only hits — no decorators, no string/dynamic references, no `__all__` exports):

| Symbol | File |
|---|---|
| `get_version_from_pyproject` | `comfy_cli/update.py` |
| `install_conda_package` | `comfy_cli/utils.py` |
| `known_providers` | `comfy_cli/auth/store.py` (1-line wrapper; real callers use `SUPPORTED_PROVIDERS` directly) |
| `get_os_details` | `comfy_cli/command/install.py` |
| `prompt_autocomplete` | `comfy_cli/ui.py` |
| `active_spec_path` | `comfy_cli/command/generate/spec.py` (unused wrapper over `_select_spec_path`) |
| `validate_comfyui` | `comfy_cli/cmdline.py` (undecorated — not a typer callback; distinct from `custom_nodes` `validate_comfyui_manager`) |
| `OAuthCancelled` | `comfy_cli/cloud/oauth.py` (never raised or caught anywhere) |
| `get_token` re-export | `comfy_cli/where.py:27` (all callers use `comfy_cli.cancellation` directly) |

## Coupled cleanups (kept the tree honest / green)

- **Orphaned imports** removed as a direct consequence of the deletions (ruff-enforced): `metadata` (update.py), `Iterable` (store.py), and `subprocess` / `typer` / the `rprint as print` shim (utils.py — all three were used only by `install_conda_package`).
- **`oauth_cancelled` registry entry** removed from `comfy_cli/error_codes.py`. The `OAuthCancelled` class attribute `code = "oauth_cancelled"` was the *only* source that made that code count as "raised"; deleting the class left it dead, and the repo's own `test_every_registered_code_is_raised` requires dead entries be deleted. This also drops it from the `comfy --json discover` contract — safe, because the code was never actually raised, so no consumer could ever have received it.

## Judgment calls / flags

- **`OAuthCancelled` (flagged per ticket):** never raised or caught today. If a branch in flight plans to use it for user-abort during login, it (and the `oauth_cancelled` error code) can be re-added at that point.
- Two deleted symbols live in auth modules (`known_providers`, `OAuthCancelled`), but deleting unreferenced code changes no auth behavior.

## Testing

- `ruff check` + `ruff format --check`: clean.
- Full `pytest` suite: **2573 passed, 37 skipped**.
- Zero behavioral risk by construction (no references).

### DIFF
--- a/comfy_cli/auth/store.py
+++ b/comfy_cli/auth/store.py
@@ -17,7 +17,6 @@ import json
 import os
 import re
 import secrets
-from collections.abc import Iterable
 from dataclasses import asdict, dataclass
 from datetime import datetime, timezone
 from pathlib import Path
@@ -209,10 +208,6 @@ def remove(provider: str) -> bool:
         del data["providers"][provider]
         _write_all(path, data)
     return True
-
-
-def known_providers() -> Iterable[str]:
-    return SUPPORTED_PROVIDERS
 
 
 # ---------------------------------------------------------------------------
--- a/comfy_cli/cloud/oauth.py
+++ b/comfy_cli/cloud/oauth.py
@@ -86,10 +86,6 @@ class OAuthRefreshError(OAuthError):
     code = "oauth_refresh_failed"
 
 
-class OAuthCancelled(OAuthError):
-    code = "oauth_cancelled"
-
-
 class OAuthTimeout(OAuthError):
     code = "oauth_timeout"
 
--- a/comfy_cli/cmdline.py
+++ b/comfy_cli/cmdline.py
@@ -1103,12 +1103,6 @@ def run_cli(
     )
 
 
-def validate_comfyui(_env_checker):
-    if _env_checker.comfy_repo is None:
-        rprint("[bold red]If ComfyUI is not installed, this feature cannot be used.[/bold red]")
-        raise typer.Exit(code=1)
-
-
 @app.command(help="Stop background ComfyUI")
 @tracking.track_command()
 def stop():
--- a/comfy_cli/command/generate/spec.py
+++ b/comfy_cli/command/generate/spec.py
@@ -420,7 +420,3 @@ def write_cache(yaml_text: str) -> Path:
     load_raw_spec.cache_clear()
     _registry.cache_clear()
     return _USER_CACHE
-
-
-def active_spec_path() -> Path:
-    return _select_spec_path()
--- a/comfy_cli/command/generate/spec.py
+++ b/comfy_cli/command/generate/spec.py
@@ -522,3 +522,7 @@ def write_cache(yaml_text: str) -> Path:
     load_raw_spec.cache_clear()
     _registry.cache_clear()
     return _USER_CACHE
+
+
+def active_spec_path() -> Path:
+    return _select_spec_path()
--- a/comfy_cli/command/install.py
+++ b/comfy_cli/command/install.py
@@ -29,12 +29,6 @@ workspace_manager = WorkspaceManager()
 console = Console()
 
 
-def get_os_details():
-    os_name = platform.system()  # e.g., Linux, Darwin (macOS), Windows
-    os_version = platform.release()
-    return os_name, os_version
-
-
 def _pip_install_torch(python: str, index_args: list[str]) -> subprocess.CompletedProcess:
     """Install torch, torchvision, and torchaudio with the given index arguments."""
     return subprocess.run(
--- a/comfy_cli/error_codes.py
+++ b/comfy_cli/error_codes.py
@@ -309,11 +309,6 @@ REGISTRY: tuple[ErrorCode, ...] = (
         "run `comfy cloud login` to sign in again",
     ),
     ErrorCode(
-        "oauth_cancelled",
-        "OAuth flow was cancelled by the user.",
-        "re-run `comfy cloud login` to retry sign-in",
-    ),
-    ErrorCode(
         "oauth_timeout",
         "Timed out waiting for browser callback during OAuth login.",
         "re-run `comfy cloud login` and complete the sign-in in your browser",
--- a/comfy_cli/ui.py
+++ b/comfy_cli/ui.py
@@ -38,26 +38,6 @@ def show_progress(iterable, total, description="Downloading..."):
 ChoiceType = str | Choice | dict[str, Any]
 
 
-def prompt_autocomplete(
-    question: str, choices: list[ChoiceType], default: ChoiceType = "", force_prompting: bool = False
-) -> ChoiceType | None:
-    """
-    Asks a single select question using questionary and returns the selected response.
-
-    Args:
-        question (str): The question to display to the user.
-        choices (List[ChoiceType]): A list of choices the user can autocomplete from.
-        default (ChoiceType): Default choice.
-        force_prompting (bool): Whether to force prompting even if skip_prompting is set.
-
-    Returns:
-        Optional[ChoiceType]: The selected choice from the user, or None if skipping prompts.
-    """
-    if workspace_manager.skip_prompting and not force_prompting:
-        return None
-    return questionary.autocomplete(question, choices=choices, default=default).ask()
-
-
 def prompt_select(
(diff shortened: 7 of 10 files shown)


mattmillerai:
## Resolver pass: no code changes needed — both red checks are pre-existing repo-wide breaks

Went through this PR for conflicts, CI, and review feedback. Summary: **nothing actionable on the diff itself**; the two failures are upstream/infra problems that also break `main`.

### Merge conflict
None — `mergeable: MERGEABLE` against `main`.

### Review feedback
No unresolved review threads. Cursor consolidated panel: *"✅ No high-signal findings"* (6/8 reviewers contributed). CodeRabbit: skipped (no matching review label).

### ❌ `build` — pre-existing, caused by an upstream tomlkit release
9 failures, all `ValueError: Comment cannot contain line breaks`, raised from `tomlkit/items.py:504` inside `comfy_cli/registry/config_parser.py` — **a file this PR does not touch**.

Root cause: `tomlkit` is unpinned (`pyproject.toml:49`), and CI now resolves **tomlkit 0.15.1**, which forbids `\n` in `.comment()`. Two call sites pass multi-line hint blocks (`config_parser.py:75` and `:250`).

Reproduced on a clean `origin/main` worktree, with no part of this PR present:

```
$ uv run --extra dev --with 'tomlkit==0.15.1' pytest tests/comfy_cli/registry/test_config_parser.py -q
8 failed, 73 passed
```

With the locally-resolved tomlkit 0.13.3, the same tests pass (81 passed). So: upstream dependency regression, not this diff.

### ❌ `test` (Windows Specific Commands) — pre-existing infra failure
Fails on **10 of the last 10 runs across many unrelated branches**. It dies in `Install Dependencies`, not in any test:

```
error: failed to remove file `...\venv\Lib\site-packages\pydantic_core/_pydantic_core.cp312-win_amd64.pyd`: Access is denied. (os error 5)
ImportError: cannot import name '__version__' from 'pydantic_core' (unknown location)
```

`comfy install` shells out to `uv pip install` against the same venv the running interpreter is loaded from; Windows locks the loaded `.pyd`, the partial removal corrupts the venv, and the next `comfy` call fails at import.

### Verification of this PR
- Full suite in an isolated worktree: **2573 passed, 37 skipped** — matches the PR description.
- `ruff check` / `ruff format --check`: clean in CI (`ruff_check` ✅).
- Re-verified every deleted symbol still has **zero** references repo-wide (incl. non-Python files). The only `validate_comfyui` hits are `validate_comfyui_manager`, a genuinely distinct symbol, as described.
- The `oauth_cancelled` registry deletion is required by the repo's own `test_every_registered_code_is_raised`, which mandates dead entries be removed.

### Follow-ups proposed
Both CI breaks are out of scope here (this is a pure dead-code deletion) and block many branches at once, so they should land independently on `main` rather than be conflated into this diff. I've recorded both for follow-up tickets: the tomlkit pin/rework (well-specified) and the Windows workflow venv-locking issue (needs a workflow-design decision first).

This PR is ready to merge on its own merits once the tomlkit break is fixed on `main`.

mattmillerai:
## Resolver re-check: still no changes needed — and both CI breaks now have fixes in flight

Re-verified this PR from scratch in an isolated worktree. Conclusion is unchanged: **the diff is clean and nothing here is actionable.** Updating my earlier comment with one new fact — the two red checks are *already being fixed* by other PRs, so no follow-up tickets are needed.

### Verification of this diff (re-run, not carried over)
- Full suite: **2573 passed, 37 skipped** — matches the PR description.
- Re-grepped every deleted symbol repo-wide: `get_version_from_pyproject`, `install_conda_package`, `known_providers`, `get_os_details`, `prompt_autocomplete`, `active_spec_path`, `OAuthCancelled`, `oauth_cancelled` — **0 hits each**. A word-boundary search for `validate_comfyui` also returns 0, confirming `validate_comfyui_manager` is a genuinely distinct symbol and is untouched.
- No merge conflict (`MERGEABLE`). No unresolved review threads. Cursor panel: no high-signal findings.

### Both red checks are pre-existing and reproduce on PRs that don't touch the code
Confirmed identical failures on #543, #544, #545, #547.

**`build`** — upstream tomlkit 0.15.1 forbids `\n` in `.comment()`; `tomlkit` is unpinned at `pyproject.toml:49`. Reproduced against an unmodified tree:
```
--with 'tomlkit==0.15.1'  -> 8 failed, 73 passed
lockfile tomlkit 0.13.3   -> 81 passed
```
Already addressed by → #536 (BE-3290) / #533 (BE-3285), with #537 (BE-3292) pinning CI to `uv.lock`.

**`test` (Windows)** — dies in `Install Dependencies`, before any test: `comfy install` shells out to `uv pip install` against the venv its own interpreter is loaded from, Windows locks the in-use `_pydantic_core\*.pyd`, and the partial removal corrupts the venv. Already addressed by → #535 (BE-3289) / #529 (`UV_LINK_MODE=copy`).

### Net
No follow-ups filed — filing any would duplicate the five PRs above. This PR needs **no changes**; it should go green on its own once the tomlkit fix lands on `main` and this branch is updated.

bigcat88:

skishore23 (approved):
Approving. Genuinely zero-risk as claimed, and I re-verified the deadness against **current** `main` rather than trusting the July audit — which matters here more than usual, since a dead-symbol audit has a shelf life.

## Verified on the merged tree (`origin/main` merged in, clean)

Re-ran the zero-reference check across `*.py`, `*.md`, and `*.json` for every symbol:

```
get_version_from_pyproject   refs=0
install_conda_package        refs=0
known_providers              refs=0
get_os_details               refs=0
prompt_autocomplete          refs=0
validate_comfyui             refs=0
OAuthCancelled               refs=0
oauth_cancelled              refs=0
```

- **`ruff check` — all checks passed**, which is the real guard on the coupled import cleanups. Confirmed `platform` is still used in `install.py` (one remaining call site) so dropping `get_os_details` didn't orphan it.
- **The `where.py` re-export is safe to drop** — the only surviving `from comfy_cli.where import …` anywhere is `cloud_preflight_or_exit` (`jobs.py:41`). Nothing imported `get_token` from there.
- **`oauth_cancelled` is correctly out of the contract** — `comfy --json discover` now reports 134 error codes with `oauth_cancelled` absent. Your reasoning is right: since `OAuthCancelled.code` was the only thing marking it as raised, leaving the entry would have failed `test_every_registered_code_is_raised`, and no consumer could ever have received a code that was never raised.
- **382 passed** across `tests/comfy_cli/output`, `auth`, `cloud`, and `command/generate/test_spec.py`. CI green.

## One description fix before squash: it's 7 symbols, not 8

The table still lists `active_spec_path` (`comfy_cli/command/generate/spec.py`), but **that deletion is not in this PR** — `spec.py` isn't among the changed files. It looks like you already caught and dropped it on a rebase, which was the right call, because it stopped being dead:

```
e073b17  2026-07-22  fix: comfy generate refresh fetches /openapi with
                     /openapi.yml fallback (BE-3391) (#560)
```

That PR added `assert spec.active_spec_path() == cache` at `tests/comfy_cli/command/generate/test_spec.py:38`, five days after this branch opened. I confirmed the function is still defined at `spec.py:527` on the merged tree and its test passes (16 passed).

So no action on the code — just drop that row from the table so the squash message doesn't claim a deletion the diff doesn't make.

## Note on `OAuthCancelled`

Good call flagging it rather than silently removing it. Worth a quick check with whoever owns the login flow that nothing in flight plans to raise it for user-abort — re-adding the class *and* its `error_codes.py` entry together is cheap, but it's cheaper still to not delete it twice.

PR #529, opened 2026-07-17
ci: set UV_LINK_MODE=copy in the Windows workflow

mattmillerai:
## ELI-5

The Windows CI job installs packages with `uv`, and `uv` saves time by hardlinking files instead of copying them. On Windows it cannot overwrite a `.pyd` that some other process still has open, so the install blows up with "Access is denied" before any test runs. Telling `uv` to just copy the files makes the problem go away.

## What

Adds `UV_LINK_MODE: copy` to `jobs.test.env` in `.github/workflows/test-windows.yml`, alongside the existing `PYTHONIOENCODING`.

## Why

The "Windows Specific Commands" workflow fails on every recent PR across all branches, in the `Install Dependencies` step, before any test runs:

```
error: failed to remove file `...\venv\Lib\site-packages\pydantic_core\_pydantic_core.cp312-win_amd64.pyd`: Access is denied. (os error 5)
```

uv then exits non-zero and the half-removed `.pyd` produces a downstream `ImportError: cannot import name '__version__' from 'pydantic_core'`.

`comfy install --fast-deps` shells out to `python -m uv pip install` (`comfy_cli/uv.py`), and uv's default link mode on Windows is hardlink/clone — it cannot replace a `.pyd` that is loaded/locked by another process on the runner. Copy mode writes a fresh file instead, which is the standard remediation for this error. Setting it at job level (rather than on the step) means it also reaches the uv subprocess that `comfy install` spawns; `comfy_cli/uv.py` never passes an explicit `env=`, so the child inherits it.

## Verification / judgment calls

- **This PR cannot verify itself.** The workflow only triggers on `paths: comfy_cli/**`, so a workflow-only change does not run it. Verify by re-running "Windows Specific Commands" on any open PR that touches `comfy_cli/**` (e.g. #517) once this lands, and confirm `Install Dependencies` and the subsequent `comfy launch --quick-test-for-ci` both pass.
- I deliberately did **not** add `.github/workflows/test-windows.yml` to the `paths` filter (which would make the workflow self-testing) — that is a separate behavior change and out of scope here. Flagging it as a reasonable follow-up.
- No Python source is touched, so there is no test or lint surface for this diff; the YAML was parsed to confirm `jobs.test.env` resolves to `{PYTHONIOENCODING: utf8, UV_LINK_MODE: copy}`.
- Tradeoff: copy mode is marginally slower/more disk than hardlinking. On a fresh CI runner that is negligible next to a job that currently always fails.

### DIFF
--- a/.github/workflows/test-windows.yml
+++ b/.github/workflows/test-windows.yml
@@ -14,6 +14,9 @@ jobs:
     runs-on: windows-latest
     env:
       PYTHONIOENCODING: "utf8"
+      # uv hardlinks by default, which cannot replace a .pyd that is still
+      # loaded by another process on the runner (os error 5). Copy instead.
+      UV_LINK_MODE: copy
 
     steps:
       - name: Check out code


mattmillerai:
Reviews-resolver pass: **no changes needed on this PR** — no unresolved review threads, no merge conflict (\`MERGEABLE\`), and the diff is still exactly the intended 3 lines in \`.github/workflows/test-windows.yml\`. Verified \`jobs.test.env\` resolves to \`{PYTHONIOENCODING: utf8, UV_LINK_MODE: copy}\`.

### The red `build` check is pre-existing and unrelated to this PR

`build` fails on `tests/comfy_cli/registry/test_config_parser.py` (7 tests) and `tests/comfy_cli/command/nodes/test_node_init.py::test_node_init_strips_credentials` with:

```
ValueError: Comment cannot contain line breaks
  tomlkit/items.py:504
```

Root cause: `comfy_cli/registry/config_parser.py:75-77` attaches a **multi-line** comment to `[tool.comfy].includes`. tomlkit **0.15.0+** rejects line breaks in `Item.comment()`. CI installs via `pip install -e .` (unpinned, `pyproject.toml:49`), so it resolves tomlkit 0.15.1, while `uv.lock` still pins 0.13.3 — which is why it reproduces in CI but not under `uv run`.

Evidence this is not caused by this PR:
- With `tomlkit==0.13.3` → **81 passed**; with `tomlkit==0.15.1` → **8 failed, 73 passed**.
- Reproduced against **`main`'s own code** (same 8 failures), so it predates this branch and affects every open PR.
- This PR touches no Python and does not affect `build-and-test.yml` at all.

This is also real user-facing breakage, not just tests: `create_comfynode_config()` backs `comfy node init`, so a fresh install crashes today.

Fixing product code inside a Windows-CI env-var PR would be scope creep, so I've **deferred it to a follow-up ticket** with a verified fix (`comfy.add(tomlkit.comment(...))` preserves the original rendering under 0.15.1) plus a recommendation to make CI installs honor `uv.lock`. That fix lands on `main` and unblocks this PR and everything else.

No action needed here beyond a re-run once the tomlkit fix merges.

mattmillerai:
Correction to my previous comment: the tomlkit fix is **already in flight as #533** ("fix(registry): emit pyproject hint blocks as single-line tomlkit comments", BE-3285), so no new follow-up ticket is needed and I have not filed one.

#533 covers the same root cause and is in fact broader than what I sketched — it fixes **both** multi-line `.comment()` sites (the `requires-comfyui` hint *and* the `classifiers` block in `initialize_project_config`), emitting one comment item per line via a shared `_append_hint()` helper.

So the state of this PR is unchanged: **nothing to fix here**. Once #533 merges to `main`, re-run `build` on this PR and it should go green. Still worth considering separately (not filing unless someone wants it) is that CI's `pip install -e .` ignores `uv.lock`, which is why tomlkit 0.15.1 reached CI while local `uv run` stayed on the locked 0.13.3 — that gap is what let this land unnoticed.

bigcat88 (approved):
Approving, with one correction: **the failure this fixes is no longer happening**, so this lands as hardening rather than a fix.

The description says the Windows job "fails on every recent PR across all branches." That was true when you opened this (2026-07-17), but the last 12 `test-windows.yml` runs are **12/12 success**, including branches from this morning:

```
2026-07-30T07:33  success  matt/be-4357-git-utils-cwd
2026-07-30T07:13  success  matt/be-3275-dedupe-auth-header-selection
2026-07-30T06:27  success  matt/be-3266-cloud-http-error-envelope
… (9 more, all success)
```

The `pydantic_core` `Access is denied (os error 5)` breakage was resolved by #572's `mixpanel<5` pin (2026-07-22), which stopped the version churn that had uv replacing a loaded `.pyd` mid-install.

I'm still approving because the underlying condition hasn't gone away — uv's default link mode on Windows remains hardlink/clone, and it still cannot replace a `.pyd` another process holds open. The pin removed today's trigger, not the mechanism. `UV_LINK_MODE: copy` costs a little install time on a CI runner and removes that whole failure class. uv's hardlink fallback is live and visible even here on Linux:

```
warning: Failed to hardlink files; falling back to full copy.
         If this is intentional, set `export UV_LINK_MODE=copy` … to suppress this warning.
```

**Your job-level placement reasoning checks out.** I verified the claim rather than assuming: `comfy_cli/uv.py` has **zero** `env=` arguments across its `subprocess` calls (`_check_call` is a bare `subprocess.check_call(cmd, cwd=cwd)`), so the uv child that `comfy install --fast-deps` spawns inherits the job environment. Putting it on the step instead would have missed exactly that subprocess.

Three lines, CI-only, no runtime code touched. Worth updating the description so the next reader doesn't go looking for a red Windows job that isn't there any more.

PR #536, opened 2026-07-17
fix(registry): emit pyproject hints as standalone comments for tomlkit 0.15

mattmillerai:
## ELI-5

`comfy node init` writes a starter `pyproject.toml` with a couple of commented-out
"here's what you could put here" hints. It built those hints by handing tomlkit a
multi-line string. tomlkit 0.15.1 started rejecting that, so the command crashed —
and since `tomlkit` was unpinned, CI picked up the new release and the `build` check
went red on `main` and therefore on every open PR.

This writes each hint line as its own comment instead, which tomlkit is happy with,
and pins `tomlkit` so a future release can't break us silently again.

## What was broken

`Item.comment()` now validates its argument and raises
`ValueError("Comment cannot contain line breaks")` on any `\n` ([tomlkit/items.py:504][1]).
Two call sites passed multi-line strings, so **every** call to `create_comfynode_config()`
raised — `comfy node init` was broken at runtime, not just in tests.

Reproduced on clean `origin/main` @ a732f5c with tomlkit 0.15.1: **9 failed** (8 in
`test_config_parser.py`, 1 in `test_node_init.py`), matching the report exactly.

[1]: https://github.com/python-poetry/tomlkit/blob/master/tomlkit/items.py

## What changed

- **`_add_commented_hint()`** emits a hint one standalone `tomlkit.comment()` per line.
  The helper takes the block and splits it, so the hint text stays verbatim and the
  shape that caused the bug is now structurally impossible to reintroduce.
- **The classifiers hint moved into `create_comfynode_config`.** `initialize_project_config`
  reparses that file and only rewrites values, so this is the only point where the hint
  can be placed under `license` *and before* the `[project.urls]` sub-table.
- **Pinned `tomlkit>=0.13,<0.16`** (defense-in-depth, as the ticket suggested).
- Two regression tests covering both hints: rendered position, inertness, and that
  uncommenting actually parses.

Rendered output is **byte-identical** to the pre-break version except two dropped
trailing spaces and one blank line (artifacts of the old inline attachment). Verified by
diffing this branch's `comfy node init` output against `origin/main`'s under tomlkit 0.13.2.

## Judgment calls (flagging for review)

1. **I did not use the fix the ticket proposed, because it doesn't hold up.** The ticket
   suggested `includes.comment('"requires-comfyui" = ...')`. That renders *inline*:
   `includes = [] # "requires-comfyui" = ...` — but the stated intent (and the code's own
   comment, "uncommentable hint") is a hint *beneath* the field that a user can uncomment
   in place. Inline, it can't be. A standalone comment preserves the original behavior
   exactly; I went with that.

2. **The ticket's root cause was incomplete — there is a second call site.** It named only
   `config_parser.py:74-76`, but that one is in `create_comfynode_config`; the 9 failing
   tests actually die at **line 250** in `initialize_project_config`, a 17-line commented-out
   `classifiers` block. Fixing only the reported lines would have left CI red. Both are fixed.

3. **Placement is load-bearing, and the obvious fix is wrong.** Naively appending the
   classifiers comment to the `project` table renders it *after* `[project.urls]` — so
   uncommenting it would define `classifiers` under `[project.urls]`, the wrong table. The
   test asserts the ordering; I verified it fails against that mutation.

4. **Verified across the whole pinned range**, not just the version I had installed — the
   affected tests pass under tomlkit 0.13.0 / 0.13.2 / 0.14.0 / 0.15.1. (At 0.13.0 blank
   hint lines render `# ` vs `#`; cosmetic only, and the tests are not overfit to it.)

5. **Left alone (pre-existing, out of scope):** the `# Used by Comfy Registry` comment
   renders *inside* `[project.urls]` rather than at the top of the file. Same class of
   placement bug, but it predates this change — I confirmed identical output on `origin/main`
   — so it isn't part of this fix. Happy to file it separately.

6. **Did not regenerate `uv.lock`.** `uv lock` wants to pull in 286 lines of unrelated churn
   (main's lock is already stale w.r.t. the `bench` extra). CI installs via `pip install -e .`
   and doesn't enforce the lock, so this is a no-op for the build; keeping the diff scoped.

## Verification

- `ruff check .` — pass
- `ruff format --check .` — pass (235 files)
- `pytest` — **2575 passed, 37 skipped** (the 9 previously-failing tests now green)
- End-to-end: ran `initialize_project_config()` in a scratch git repo and diffed the real
  emitted `pyproject.toml` against `origin/main`'s.

### DIFF
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -50,7 +50,7 @@ dependencies = [
   "rich",
   "ruff",
   "semver~=3.0.2",
-  "tomlkit",
+  "tomlkit>=0.13,<0.16",
   "typer>=0.12.5",
   "typing-extensions>=4.7",
   "uv>=0.11.15",
--- a/uv.lock
+++ b/uv.lock
@@ -267,7 +267,7 @@ requires-dist = [
     { name = "ruff", marker = "extra == 'dev'" },
     { name = "semver", specifier = "~=3.0.2" },
     { name = "term-image", marker = "extra == 'preview'", specifier = ">=0.7,<0.8" },
-    { name = "tomlkit" },
+    { name = "tomlkit", specifier = ">=0.13,<0.16" },
     { name = "typer", specifier = ">=0.12.5" },
     { name = "typing-extensions", specifier = ">=4.7" },
     { name = "uv", specifier = ">=0.11.15" },


comfy_cli/registry/config_parser.py
```diff
@@ -44,6 +45,46 @@
...
+    """Append `hint` to `table` as standalone comment lines.
+
+    tomlkit's `Item.comment()` takes a single line by contract — it raises
+    `ValueError` on line breaks as of tomlkit 0.15.1 — and renders inline, which
+    would leave the hint impossible to uncomment in place. So emit one standalone
+    comment per line instead. `tomlkit.comment()` re-adds the `#` marker, hence
+    the strip. Hints must be added while `table` is still being built, before any
+    sub-table: a comment appended after one renders *inside* that sub-table.
+    """
+    for line in hint.strip("\n").splitlines():
```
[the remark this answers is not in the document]
mattmillerai: Fixed in 7932c01 — switched to `hint.strip()` so a CRLF checkout no longer leaves a leading `\r` that `splitlines()` turns into a stray empty `#` comment. Added a CRLF regression test (`test_add_commented_hint_tolerates_crlf_source`).

pyproject.toml
```diff
@@ -46,7 +46,7 @@ dependencies = [
   "rich",
   "ruff",
   "semver~=3.0.2",
-  "tomlkit",
+  "tomlkit>=0.13,<0.16",
```
bigcat88, line 53: Changing the requirement to `tomlkit>=0.13,<0.16` without regenerating `uv.lock` leaves the lockfile's `requires-dist` out of sync with pyproject — `uv lock --check` fails, and once #537's `--locked` CI lands (approved today) every job on this branch fails at the install step. Needs `uv lock` + committing the lockfile (or drop the bound change from this PR).

comfy_cli/registry/config_parser.py
```diff
@@ -44,6 +45,49 @@
...
+
+    tomlkit's `Item.comment()` takes a single line by contract — it raises
+    `ValueError` on line breaks as of tomlkit 0.15.1 — and renders inline, which
+    would leave the hint impossible to uncomment in place. So emit one standalone
+    comment per line instead. `tomlkit.comment()` re-adds the `#` marker, hence
+    the strip. Hints must be added while `table` is still being built, before any
+    sub-table: a comment appended after one renders *inside* that sub-table.
+    """
+    # `strip()` (not `strip("\n")`) so a CRLF checkout of this file — the repo has
+    # no `.gitattributes` and `core.autocrlf=true` is the Windows default — doesn't
```
bigcat88: This premise is false — CPython reads source files in universal-newlines mode, so a CRLF checkout can never leave `\r` inside a triple-quoted literal. Verified:

```
$ printf 'X = """\r\nline1\r\n# line2\r\n"""\r\n' > crlf_test.py
$ python -c "import crlf_test; print('\r' in crlf_test.X)"
False
```

`strip()` is harmless, but the comment documents protection against something that can't happen — please drop or reword it.

comfy_cli/registry/config_parser.py
```diff
@@ -44,6 +45,49 @@
...
+    # leave a leading `\r` that `splitlines()` would turn into a stray empty `#`.
+    for line in hint.strip().splitlines():
+        table.add(tomlkit.comment(line.removeprefix("#").removeprefix(" ")))
```
bigcat88: For the blank separator lines (`#`), `removeprefix` yields `""` and `tomlkit.comment("")` renders `"# "` — trailing whitespace in the file we hand to users (confirmed in the rendered output, and it's what the repo's own pre-commit `trailing-whitespace` hook would flag). #533 sidesteps this with `Comment(Trivia(comment_ws="", comment="#"))`, which renders a bare `#`.

bigcat88 (changes requested):
This is a second, independent fix for the same tomlkit-0.15 breakage as #533 (also yours), which I approved yesterday. I ran the full `comfy node init` flow on both branches — the generated pyproject.toml files are equivalent apart from whitespace, so functionally either works. But as-is this one can't merge:

1. **uv.lock is not relocked** for the new `tomlkit>=0.13,<0.16` bound (inline comment below).
2. Two PRs fixing the same thing can't both land — whichever merges second conflicts. Since #533 is already approved and its output is slightly cleaner (no trailing whitespace on the `#` separator lines — this branch renders them as `"# "`), my suggestion: **close this one in favor of #533**, and if you want to keep the `tomlkit<0.16` upper bound, land it as a tiny follow-up with the relock included.
3. The CRLF rationale for the last commit doesn't hold up empirically (inline comment below).

pyproject.toml
mattmillerai: Fixed in 4a5dd5f — ran `uv lock` and committed the refreshed `uv.lock`; `uv lock --check` now passes, so this branch won't fail the `--locked` install step once #537 lands.

comfy_cli/registry/config_parser.py
mattmillerai: You're right — CPython reads source in universal-newlines mode, so a CRLF checkout can't leave `\r` in the literal. Reworded the comment (and the test's docstring) in 4a5dd5f: `strip()` is really about dropping the blank lines the triple-quoted literal carries around its `#` body, not CRLF.

comfy_cli/registry/config_parser.py
mattmillerai: Fixed in 4a5dd5f — blank separator lines now render as a bare `#` via `tomlkit.items.Comment(Trivia(comment_ws="", comment="#"))`, matching #533. Verified the generated pyproject.toml has zero trailing-whitespace lines, and added an assertion locking that in.

mattmillerai:
Thanks @bigcat88 — addressed all three inline points in 4a5dd5f:

1. **uv.lock relocked** for the `tomlkit>=0.13,<0.16` bound; `uv lock --check` passes now.
2. **CRLF rationale dropped** — you're right that CPython's universal-newlines reading makes the `\r` scenario impossible; reworded the comment/test to the real reason (stripping the literal's surrounding blank lines).
3. **Trailing whitespace fixed** — blank separator lines now render as a bare `#` via `Trivia(comment_ws="", comment="#")`, matching #533. Confirmed zero trailing-whitespace lines in the generated file.

On **close-vs-keep**: that's a call for the author, not something I'll decide here. The two PRs do overlap, but this one additionally pins the `tomlkit<0.16` upper bound (with the relock now included, per your own suggestion). If #533 is the preferred vehicle, close this; if the upper bound is wanted, this branch is now technically clean to land. Leaving the decision to @mattmillerai.

mattmillerai:
Since #533 has merged (it landed the equivalent hint-rendering fix, confirmed against `origin/main` — same `_append_hint`/bare-`#`-separator approach), reworked this branch down to exactly the tiny follow-up you suggested: only the `tomlkit>=0.13,<0.16` pin in pyproject.toml + a relocked uv.lock. Dropped the now-redundant config_parser.py/test changes and rebased clean onto main — no more conflict. `uv lock --check` and the full test suite pass.

skishore23 (approved):
Approving. @bigcat88's three blockers all look addressed — this is now exactly the "tiny follow-up with the relock included" they suggested.

## State check

The code fix landed on `main` via #533, in @bigcat88's preferred form — `config_parser.py:75-79` splits the hint per line and uses `Comment(Trivia(comment_ws="", comment="#"))` for blank separators, so no `"# "` trailing whitespace. `tomlkit` is still **unpinned** on `main` (`pyproject.toml:53`), so the pin is still worth landing.

This diff is now **only** the pin plus the lock line — the duplicated code fix and the disputed CRLF commit are gone.

## Verified

- **Merges cleanly** with current `origin/main`.
- **`uv lock --check` passes** — resolves 73 packages with no error. That was blocker #1; the locked `tomlkit 0.13.3` already satisfies `>=0.13,<0.16`, so no churn was needed beyond the `specifier` line. Good that this avoided the 286 lines of unrelated relock drift.
- `test_config_parser.py` + `test_node_init.py` — **87 passed**.
- **The fixed code is structurally immune to the 0.15 breakage.** Rather than trust the version I happen to have installed (0.13.3), I monkeypatched tomlkit's `Item.comment` / `tomlkit.comment` to emulate 0.15.1's `ValueError("Comment cannot contain line breaks")` and ran `create_comfynode_config()` end to end:
  ```
  create_comfynode_config() under emulated 0.15 validation: OK
  hint lines rendered: 2
  any trailing-whitespace comment lines: False
  ```
  So no `\n` ever reaches `comment()`, and the output is clean. The pin is genuine defense-in-depth rather than the thing doing the work.

## One note on the bound itself (non-blocking)

`<0.16` will need a bump when 0.16 ships, and since the code is now structurally safe, the bound is insurance rather than necessity. Worth keeping given this exact breakage just cost the repo a red `build` on every open PR — but it's the kind of ceiling that goes stale quietly, so it may be worth a Dependabot/renovate rule to test-and-raise it rather than leaving it to be discovered.

Note my approval won't clear @bigcat88's outstanding CHANGES_REQUESTED — they'll need to re-review to unblock. Flagging @bigcat88: points 1 (relock), 2 (duplicate of #533), and 3 (CRLF rationale) all appear resolved by the rescope.

bigcat88 (approved):
Addressed — approving and merging. All three of my objections are resolved, and the PR is now exactly the shape I suggested: the tiny follow-up pin, with the relock included. Sorry it sat 18 days — you fixed it the next morning and I was filtering my queue on `reviewDecision`, which never clears without a re-review.

| my objection | status |
|---|---|
| `uv.lock` not relocked for the new bound | fixed — `uv lock --check` resolves cleanly, and `requires-dist` now carries `{ name = "tomlkit", specifier = ">=0.13,<0.16" }` |
| duplicate of #533, can't both land | resolved — #533 merged 07-23, and this branch is now **2 files, 2 lines**: the pin plus its relock. Nothing overlaps |
| the CRLF rationale didn't hold | moot — that content is gone; the branch touches no `registry/` file at all |

## I tested the bound rather than just the lock

The lock resolves **tomlkit 0.13.3**, so neither CI nor the suite ever exercises the top of the range this pin admits. If `<0.16` let through a version that breaks the thing the pin exists for, nothing here would catch it. So I built an isolated env at **tomlkit 0.15.1** (current latest) with the package installed, and ran the paths that broke originally:

- `tests/comfy_cli/registry/` — **110 passed**
- a real `comfy node init` in a fresh git repo — exit 0, and the multi-line hint block emits correctly as separate single-line comments:

```
#     "Operating System :: OS Independent",
#     "Environment :: GPU :: NVIDIA CUDA",    # NVIDIA CUDA support
#  Used by Comfy Registry https://registry.comfy.org
# "requires-comfyui" = ">=1.0.0"  # ComfyUI version compatibility
```

- the generated file re-parses and re-dumps through tomlkit 0.15.1 without raising

So the range is right: 0.15.1 is admitted and works, which is what #533 was for. The `<0.16` half is preemptive — 0.16 doesn't exist yet — and after being burned by a formatting behaviour change in a *minor* release, capping is a reasonable call for a dependency this output-sensitive.

## Everything else

- Full suite: **4773 passed, 31 skipped**. The one failure is `test_non_fast_deps_uses_global_python`, which fails identically on plain `main` here.

## One nit for later

Nothing records *why* the upper bound is there, so the next person to see `tomlkit>=0.13,<0.16` in `pyproject.toml` has no way to tell a considered cap from a stale one — and the usual fate of an undocumented cap is to sit unexamined for years. A trailing comment (`# <0.16: 0.15 changed multi-line comment handling, see BE-3290 / #533`) would fix that. Not worth another round-trip on a 2-line PR; fold it in if you touch the file again.

PR #540, opened 2026-07-17
fix(workflow): detect oversize responses in _http_request instead of silently truncating (BE-3297)

mattmillerai:
## ELI-5

When the CLI asks Comfy Cloud for a workflow, it reads at most 64 MiB of the reply. If the reply was bigger, the CLI quietly chopped it off. The chopped-off JSON was then broken, so parsing failed — and that failure was caught and turned into "no body at all." The user got the same result as an empty response, with no clue their workflow had been truncated.

Now the CLI reads one byte past the limit. If that extra byte exists, the body was too big, and it says so loudly with a `workflow_too_large` error instead of pretending nothing was there.

## What changed

- Added a module-level `_HTTP_MAX_BYTES = 64 * 1024 * 1024`, replacing the inline `64 * 1024 * 1024` magic number in `_http_request`.
- `_http_request` now reads `_HTTP_MAX_BYTES + 1` and raises `_ResponseTooLarge` when the body exceeds the cap — mirroring the sibling `_userdata_request`, which has done exactly this since it was written.
- `_handle_cloud_http_error` gained a `_ResponseTooLarge` branch emitting a `workflow_too_large` envelope with `details.limit_bytes`.
- All four `_http_request` call sites (`list`, `get`, `save`, `delete`) added `_ResponseTooLarge` to their `except` tuple, so the new exception routes to an envelope rather than an uncaught traceback.
- Generalized the `_ResponseTooLarge` docstring, which named `/userdata` specifically and now covers both surfaces.

## Judgment calls / deviations from the ticket

**The ticket's premise about `_handle_cloud_http_error` was wrong, and I fixed it rather than working around it.** The ticket states the cloud handler "already knows how to map `_ResponseTooLarge` to a `limit_bytes` envelope error." It does not — that is `_handle_local_http_error`, the *local* `/userdata` sibling. The cloud handler would have dropped `_ResponseTooLarge` into its generic `else` branch and emitted a vague `cloud_http_error: get failed: ` with an empty exception string and a misleading "check network" hint. So step 3 of the plan required adding the branch, not just wiring the `except` tuples. This is the one substantive addition beyond the ticket's four steps.

**Separate constant rather than reusing `_USERDATA_MAX_BYTES`.** Both are 64 MiB and could share one name. I kept them distinct because they cap two different surfaces (local ComfyUI `/userdata` vs. the cloud API) that may reasonably diverge later; collapsing them would couple the two. Same value today, so no behavior difference either way.

**`details` includes `workflow_id`** in the cloud too-large envelope (the local one omits it). It is `None` for `list`/`save`, which is accurate and matches the other cloud error branches.

## Verification

`uv run -p 3.10 --extra dev pytest tests/comfy_cli/command/ -q` → **1165 passed, 2 skipped**. `ruff format --check` and `ruff check` clean on both touched files.

Six new tests:
- `TestGet::test_response_over_cap_refuses_to_truncate` — end-to-end envelope assertion, including `details.limit_bytes`.
- `TestHttpRequestCap::test_oversize_body_raises_rather_than_returning_none` — the exact regression the ticket names: oversize must raise, not return `(status, None)`.
- `test_body_exactly_at_cap_still_parses` — boundary guard: `len(raw) == cap` is a *complete* body, not a truncated one, and must still parse.
- `test_empty_body_still_returns_none` — guards the unchanged `(status, None)` contract for a genuinely empty body.
- `test_every_call_site_routes_oversize_to_envelope[list|get|save|delete]` — parametrized over all four call sites, so an untouched `except` tuple fails the suite rather than surfacing a traceback to users.

**The two regression tests were confirmed to fail against pre-fix behavior**, not just pass against the new code: reverting `_http_request` to the truncating read (`resp.read(_HTTP_MAX_BYTES)`, no overflow check) reproduces `DID NOT RAISE _ResponseTooLarge` and a red `test_response_over_cap_refuses_to_truncate`. The three boundary/contract tests pass in both directions by design — they guard against over-correction.

## On the new error path

This adds a failure path, so worth stating plainly: it does **not** deny a capability that works today. Fetching an oversize workflow is already broken on `main` — I ran the pre-fix `_http_request` against a truncated body and it returns `(200, None)`, which for `get` then surfaces as a misleading "unexpected response shape" error naming the wrong cause. The change converts an already-failing silent path into a loud, correctly-coded one; it does not remove a working path. Behavior for every in-cap response (the entire real-world case) is byte-for-byte unchanged.

## Scope

`_http_request` has exactly four call sites, all in `workflow.py`; neither it nor `_ResponseTooLarge` is imported anywhere else in the repo (verified by grep), so the new exception cannot escape to an unprepared caller.

Deferred from a review thread on #530.


mattmillerai:
## Review resolution (f91cab6)

The cursor-review panel's judge crashed with a `parse_error` before it could post findings, but its truncated output named four concrete issues. I treated those as real review signal and worked each one. Three held up; details below.

### Fixed

**1. The `workflow_too_large` hint was hardcoded to the `get` wording.** The panel was right, and this was on code this PR introduced. `_handle_cloud_http_error` serves all four verbs, but the hint read *"the saved workflow is unexpectedly large"* — meaningless for `list`, where an oversize response means *too many workflows*, not one large one. Replaced with a per-operation `_TOO_LARGE_HINTS` table.

**2. `save` can succeed server-side yet report `workflow_too_large`.** Also correct, and a better catch than it first looks: `save`/`delete` have already sent their request by the time the response is read, so an oversize response says nothing about whether the write landed. The old hint implied a clean failure, which would invite a user to retry a mutation that already succeeded. Their hints now say the change *may still have landed* and point at `workflow list` to confirm.

**3. `UnicodeDecodeError` escapes `_http_request`.** Confirmed empirically — `json.loads()` on non-UTF-8 **bytes** raises `UnicodeDecodeError`, which is a `ValueError` but *not* a `JSONDecodeError`, so it escaped the `except` and no call site caught it. A non-UTF-8 body surfaced as a **traceback**, not an envelope. Now named alongside `JSONDecodeError`.

**4. The `HTTPError` body read was unbounded.** `(e.read() or b"")[:1000]` pulled an arbitrarily large error body into memory before slicing. `e.read(1000)` is byte-for-byte equivalent, verified including the `fp=None` edge case.

3 and 4 are pre-existing rather than introduced here, but both sit in the two functions this PR rewrites and are squarely on its theme (don't read unbounded; don't let a malformed body produce a confusing failure), so they're fixed rather than deferred.

### Tests

Four new tests, each confirmed to **fail against the pre-fix code** — the hint test fails for `list`/`save`/`delete` and passes for `get`, which is exactly the shape of the bug. `1169 passed, 2 skipped`; ruff clean.

### Deferred

`_handle_local_http_error` (workflow.py:548) has the **identical** hardcoded-hint bug on the local `/userdata` surface. Not fixed here: this PR is scoped to the cloud `_http_request`, the local path is untouched pre-existing code no reviewer looked at on this PR, and its hints need different wording ("on the server" vs "cloud UI"), so it's a separate table rather than a reuse. Filed as a follow-up with a full plan.

### On the two red checks

Both are **pre-existing repo-wide breakage, unrelated to this diff** (which touches only `workflow.py`):
- `build` — 9 failures in `registry/test_config_parser.py` + `nodes/test_node_init.py`, all `tomlkit` "Comment cannot contain line breaks" (tomlkit 0.15). Fix in flight: #533 / #536.
- `test` — dependency install fails on `pydantic_core` import. Fix in flight: #535.

Neither touches any file in this PR, and both pass locally. This PR's own suite is green.

**Note:** CodeRabbit never actually reviewed this PR — it hit a Fair-Usage rate limit and posted only the limit notice, though its check reports "pass". Re-triggering below.

mattmillerai:
@coderabbitai review

mattmillerai:
- <https://linear.app/comfyorg/issue/BE-3311|BE-3311> — Give the local /userdata workflow_too_large error per-operation hints

mattmillerai:
## Second-pass review resolution (812274e)

All review threads on this PR were already resolved by the previous pass (f91cab6). This pass found one thing that pass introduced, and re-verified the rest.

### Fixed: dropped an unrelated 287-line `uv.lock` regeneration

`f91cab6` accidentally committed a full lockfile regen — adding `anthropic`, `pydantic`, `jiter` and friends to a PR that only touches `_http_request`. The PR description never mentions it, which is what flagged it.

The drift is **real but pre-existing, not contamination**: #490 added `optional-dependencies.bench = [ "anthropic>=0.40" ]` to `pyproject.toml` without re-locking, so `uv lock --check` fails on `main` today. I verified this directly — main's lock reports *"needs to be updated"*, this branch's passes.

The trap is that a bare `uv run --extra dev pytest ...` (this repo's own documented test command) **silently re-locks as a side effect**. That's how it got picked up, and it re-dirtied the file on me mid-pass too. Reverted to main's copy and committed without touching `uv` in between.

Behavior-neutral: no CI job consumes `uv.lock` (only `ci-cursor-review.yml` mentions it, and only to *exclude* it from the diff-size gate). Carrying it here would have invited conflicts with #533/#535/#536, all in flight on dependency code. Diff is now 2 files / 120 lines.

Deferred to a follow-up with a full re-lock plan → tracked by the fleet.

### Re-verified the substantive diff

- **`_TOO_LARGE_HINTS["list"]` points at `--limit` and `--name`** — confirmed both flags actually exist on `list_cmd`, so the hint is actionable rather than inventing options.
- **`_ResponseTooLarge` is raised after the `with urllib.request.urlopen(...)` block exits**, so the response is closed before unwinding.
- **All four call sites** (`list`/`get`/`save`/`delete`) carry `_ResponseTooLarge` in their `except` tuple; the parametrized test fails the suite if one is missed.
- **Negative-claim check:** this adds a failure path, so — it denies nothing that works today. Oversize fetch is already broken on `main` (returns `(200, None)`, surfacing as a misleading "unexpected response shape"). This converts an already-failing silent path into a correctly-coded loud one. Every in-cap response is byte-for-byte unchanged.

`1169 passed, 2 skipped`; `ruff format --check` and `ruff check` clean.

### The one red check is not this PR

`test` (windows) fails in **Install Dependencies**, before any test runs:

```
error: failed to remove file `...\venv\Lib\site-packages\pydantic_core\_pydantic_core.cp312-win_amd64.pyd`: Access is denied. (os error 5)
ImportError: cannot import name '__version__' from 'pydantic_core' (unknown location)
```

That's the pre-existing repo-wide breakage #535 (BE-3289) is fixing — a locked-DLL upgrade during `comfy install`, on a code path this PR does not touch. This PR's own suite is green.

mattmillerai:
## Third-pass review resolution — no code change

No unresolved review threads, no merge conflict (`MERGEABLE`, branch is current with `main`). The two prior passes (f91cab6, 812274e) resolved every finding the cursor panel surfaced. This pass re-verified the diff independently and diagnosed the **newly-red `build` check**, which is not this PR.

### The `build` red is a repo-wide dependency break, already fixed by #537

`build` fails on **every open PR** (#540, #541, #543, #544) with the same 9 failures in `test_config_parser.py` / `test_node_init.py` — files this PR does not touch:

```
ValueError: Comment cannot contain line breaks
```

Root cause, reproduced locally: `tomlkit` **0.15.1** shipped today and rejects comments containing line breaks. `uv.lock` pins `tomlkit==0.13.3`, but `pytest.yml` installs via `pip install -e .`, which re-resolves from scratch and never sees the pin. Confirmed both directions:

- `uv run --frozen --extra dev pytest tests/comfy_cli/registry/test_config_parser.py` → **81 passed** (locked 0.13.3)
- `uv run --extra dev --with 'tomlkit>0.13.3' pytest ...` → **8 failed** (0.15.1) — the exact CI signature

**#537 (BE-3292) already fixes this** by installing from `uv.lock`; its own `build` job is green. Nothing to do here — this PR goes green once #537 lands. The `test` (windows) red is the pre-existing `pydantic_core` locked-DLL failure tracked by #535 (BE-3289), also unrelated.

### Re-verified the diff independently

- **Every branch of `_handle_cloud_http_error` reaches `return typer.Exit(code=1)`** — the new `_ResponseTooLarge` branch is an `if` ahead of the `HTTPError` `elif`, so it cannot fall through and return `None` into a `raise ... from`.
- **`_ResponseTooLarge` is not an `HTTPError`**, so ordering the `isinstance` check first is required, not stylistic.
- **`raise _ResponseTooLarge()` sits outside the `with urlopen(...)` block** — the response is closed before unwinding.
- **All four call sites** carry `_ResponseTooLarge` in their `except` tuple; the parametrized test fails the suite if one is dropped.
- **`e.read(1000)`** is equivalent to the old `(e.read() or b"")[:1000]` without the unbounded read.

`1169 passed, 2 skipped`; `ruff format --check` and `ruff check` clean. Working tree clean — `uv.lock` untouched (used `uv run --frozen` throughout to avoid the silent re-lock trap #812274e documented).

No new follow-ups: the deferrals from earlier passes are already filed as BE-3311 and BE-3312.

mattmillerai:
## Fourth-pass review resolution — no code change

Re-verified end to end in an isolated worktree. State: `MERGEABLE`, branch current with `main` (3 ahead / 0 behind `a732f5c`), **zero unresolved review threads**. Nothing to rebase, nothing to resolve.

### Both red checks are pre-existing dependency drift, not this PR

The PR diff is 2 files (`workflow.py` + its test); it touches neither failing test file.

- **`build`** — 9 failures in `test_config_parser.py` / `test_node_init.py` with `ValueError: Comment cannot contain line breaks`. Reproduced locally: `tomlkit` **0.15.1** rejects multi-line `.comment()`; `uv.lock` pins `0.13.3` but `pytest.yml` installs via `pip install -e .` which re-resolves fresh. Confirmed both directions — `uv run --frozen … pytest` → **81 passed**, fresh resolve → **8 failed** (exact CI signature). Fixed by **#537 (BE-3292)** (installs from `uv.lock`), still open. Not this PR.
- **`test` (windows)** — `ImportError: cannot import name '__version__' from 'pydantic_core'` during *Install Dependencies*, a locked-DLL upgrade on a path this PR doesn't touch. Tracked by **#535 (BE-3289)**. Not this PR.

Full run with the frozen lock (what #537 makes CI do): the PR suite + both currently-red files → **132 passed**. `ruff check` / `ruff format --check` clean.

### Diff re-verified independently

`_ResponseTooLarge` is raised only after the `urlopen` block closes; the one-byte-over read distinguishes full from truncated; all four call sites (`list`/`get`/`save`/`delete`) carry it in their `except` tuple; the `isinstance(_ResponseTooLarge)` branch precedes the `HTTPError` `elif` and can't fall through.

### One new deferral (out of scope, follow-up filed by the fleet)

The Cursor panel's judge (the panel comment itself failed to post via `parse_error`, so no thread exists) surfaced a real *pre-existing* latent issue this PR only makes consistent: `_http_request` returns `(status, None)` for a non-empty **unparseable** 200 body, so `list_cmd` reports `count:0, ok:true` and `save_cmd` reports `ok:true, workflow_id:null` rather than a malformed-response error. Predates this PR (the `JSONDecodeError` path already did this) and is separate from the oversize/byte-cap work → deferred to an `agent-ok` follow-up (distinguish empty-vs-unparseable in `_http_request`, raise a `_ResponseUnparseable` mapped to a `workflow_bad_response` code).

This PR is code-complete and correct; it goes green once #537 (and #535 for the Windows leg) land.

mattmillerai:
- <https://linear.app/comfyorg/issue/BE-3334|BE-3334> — workflow cloud: surface an error when a 200 response body is unparseable instead of reporting empty/success

mattmillerai:
## Review resolution (reviews loop)

No unresolved review threads on this PR. The Cursor panel's judge crashed (`parse_error`) both runs without posting structured findings; the four concrete issues its truncated output named were already fixed in f91cab6. Re-reading the latest crash (`pullrequestreview-4721620756`), its output began articulating a **fifth** concern before cutting off — worth recording:

### Deferred (out of scope → follow-up filed)

- **`list`/`save` silently succeed on an *undecodable* cloud response.** `_http_request` returns `(status, None)` for both an empty and an undecodable body. `get_cmd` guards against this (non-dict `body` → `cloud_http_error`), but `list_cmd` (`rows = (body or {}).get("data") or []` → `count:0, ok:true`) and `save_cmd` (`workflow_id = (resp or {}).get("id")` → `workflow_id:null, ok:true`) do not. This is **pre-existing** (the `except: return status, None` predates this PR — #540 only added `UnicodeDecodeError` to the same catch) and orthogonal to this PR's *oversize*-detection theme, which is correctly handled via `_ResponseTooLarge`. Deferred to a follow-up rather than expanding this PR's scope.

### CI status

The `build` (pytest) and `test` (Windows) checks are red, but **not** because of this PR — it only touches `workflow.py`. The 9 `build` failures are all in untouched files (`registry/test_config_parser.py`, `nodes/test_node_init.py`) with `ValueError: Comment cannot contain line breaks`, and the **identical** failures appear on other open PRs (e.g. `matt/be-3326`, run 29593447664). Root cause is a newer `tomlkit` that CI resolves fresh (local `0.13.3` passes) now rejecting multi-line comments produced by `registry/config_parser.py`. The Windows `test` failure is a `pydantic_core` install flake in the dependency step. Both are repo-wide/environmental and must be fixed on `main`; a follow-up is filed. No code change made here.

bigcat88 (approved):
Verified and approving:

- The cap-detection fix is correct and mirrors the `_userdata_request` precedent: `read(_HTTP_MAX_BYTES + 1)` distinguishes exactly-at-cap (parses, pinned by `test_body_exactly_at_cap_still_parses`) from oversize (raises). `workflow_too_large` was already registered (`error_codes.py:78`), so no registry churn.
- The two adjacent fixes riding along are both real: bytes that aren't valid UTF-8 previously raised `UnicodeDecodeError` out of `json.loads` — a `ValueError` that none of the call-site `except` tuples caught, i.e. an uncaught traceback on main; and `e.read(1000)` bounds the error-body read that `e.read()[:1000]` didn't.
- The per-operation hints acknowledging that `save`/`delete` may have already landed server-side are a thoughtful touch.
- Full unit suite on this branch merged with current main: 2555 passed. Mutation check: reverting `_http_request` to the truncating read fails 6 tests (both regression tests plus the four parametrized call-site routes) — the suite pins the fix, not just exercises it.

bigcat88:
The red `build`/`test` checks are the pre-existing Windows pydantic-core DLL lock (branch CI last ran July 17, before #572's `mixpanel<5` pin landed on main). Update from main and they should go green — full suite passes locally on this branch merged with current main.

PR #545, opened 2026-07-17
chore: re-lock uv.lock to match the bench optional-dependency extra

mattmillerai:
## ELI-5

`pyproject.toml` gained an optional `bench` extra a while back (in #490), but the lockfile was never regenerated to match. So the lock and the manifest disagreed: `uv lock --check` failed on `main`, and any bare `uv run` — like the one in our own test instructions — would silently re-lock and leave a ~287-line `uv.lock` diff in your working tree. That has already been committed into an unrelated PR by accident once, and reverted to keep that PR's diff clean. This regenerates the lockfile so the two agree again.

## What changed

`uv.lock` only, regenerated with `uv lock` (uv 0.11.26) from a clean checkout of `main`:

- Adds the `anthropic` dependency tree pulled in by the `bench` extra: `anthropic`, `pydantic`, `pydantic-core`, `jiter`, `docstring-parser`, `annotated-types`, `typing-inspection`.
- Records `bench` in `provides-extras`.

## Risk surface

The diff is 285 additions and **2 deletions**. Everything else is purely additive. Both deletions:

1. `provides-extras = ["dev", "preview"]` → `["bench", "dev", "preview"]` — this is the manifest/lock disagreement being fixed.
2. `exceptiongroup`'s dependency on `typing-extensions` lost its `python_full_version < '3.13'` marker. This is a semantic no-op: `exceptiongroup` is only ever reachable under `python_full_version < '3.11'` (via `anyio` and `pytest`), so a `< '3.13'` marker is tautologically true everywhere it is installed, and uv simplified it away.

**No existing package version changed** — no other lines were removed. `requires-python = ">=3.10"` and the lockfile revision are unchanged.

## Verification

- `uv lock --check` → exits 0 (fails on `main` today).
- `uv run -p 3.10 --extra dev pytest tests/comfy_cli/command/ -q` → 1157 passed, 2 skipped.
- `uv run -p 3.10 --extra dev pytest tests/comfy_cli/ -q` → 2471 passed, 13 skipped.
- `ruff check .` reports 14 pre-existing `UP038` errors — identical on a pristine `origin/main` checkout, so unrelated to this change. `ruff format --check .` → 235 files already formatted.
- `uv run --extra bench python -c "import anthropic, pydantic"` → installs the 7 new packages and imports cleanly (anthropic 0.117.0 / pydantic 2.13.4), confirming the newly-locked tree actually resolves rather than just looking plausible.
- After a bare `uv run`, the working tree shows no new lock churn and `uv lock --check` still exits 0 — the silent re-dirty behavior is gone.

## Judgment calls

- **Landed without waiting for #533/#535/#536.** The plan suggested landing after those merge to avoid a lockfile conflict. None of them touch `uv.lock`, and only #536 touches `pyproject.toml` — so there is no conflict to wait for and this can land in any order.
- **#536 will re-introduce the drift.** #536 pins `tomlkit>=0.13,<0.16`. Applying that change on top of this branch makes `uv lock --check` fail again, so whichever of the two merges second needs a re-lock. Cheapest fix is for #536 to run `uv lock` and include the result.
- **The CI `uv lock --check` gate is deliberately not included here.** The hardening is worth doing — #536 above is concrete proof the drift recurs — but it is a repo-wide CI policy change that would immediately fail an in-flight PR, and it touches nothing this PR touches. It is better as its own PR that can be weighed on its own merits than bundled into a mechanical re-lock. Recommending it as a follow-up.


mattmillerai:
The failing `build` check is **not caused by this PR** — it is an environmental break on `main` that landed today, and it is already fixed by #536.

## Root cause

`tomlkit` **0.15.1** was released today (2026-07-17 01:48 UTC). It makes `Item.comment()` reject line breaks, which breaks `create_comfynode_config()` in `comfy_cli/registry/config_parser.py:75` and `:250`. Bisected locally across versions:

| tomlkit | `Item.comment(multiline)` |
|---|---|
| 0.13.3 | OK |
| 0.15.0 | OK |
| 0.15.1 | `ValueError: Comment cannot contain line breaks` |

Timeline lines up exactly: `main` last passed `pytest.yml` on 2026-07-13 (a732f5c); tomlkit 0.15.1 published 01:48 today; this PR's run was 10:32 today.

## Why this PR cannot be the cause

`pytest.yml` installs with `pip install pytest pytest-cov jsonschema` + `pip install -e .`. It never reads `uv.lock`, and `uv.lock` is the *only* file this PR touches (285+/2-). `pip` resolves the unpinned `"tomlkit"` from `pyproject.toml` to whatever is newest, so it picked up 0.15.1. Any PR opened against `main` today fails the same 9 tests.

## The fix is #536, not here

#536 (`fix(registry): emit pyproject hints as standalone comments for tomlkit 0.15`) pins `tomlkit>=0.13,<0.16` and emits standalone comments instead — its `build` check is green. Duplicating that fix into this lockfile-only PR would just conflict with it.

## Merge-order note

#536 changes `pyproject.toml` but **not** `uv.lock`, so the two interact:

- **This PR merges first** → #536 then needs a re-lock, otherwise `uv lock --check` drifts again.
- **#536 merges first** → this branch needs a rebase + `uv lock` to pick up the `tomlkit` pin, which also turns this PR's CI green.

Verified the lockfile is still correct as-is: `uv lock --check` exits 0 on this branch and fails on a pristine `main` clone — the drift this PR fixes is still real and still unfixed on `main`.

Leaving the code untouched. Happy to rebase + re-lock the moment #536 lands.

mattmillerai:
🤖 Reviews-resolver re-checked this PR — status unchanged and confirmed:

- **Lockfile still correct & still needed.** `uv lock --check` exits 0 on this branch; `main` still has `provides-extras = ["dev", "preview"]` (missing `bench`), a 289-line drift. No new base conflict — PR is `MERGEABLE`.
- **The red `build` check is not caused by this PR.** It's the environmental tomlkit-0.15.1 break in `config_parser.py` (multi-line `Item.comment()`), and `pytest.yml` installs via `pip install -e .` — it never reads `uv.lock`, the only file this PR touches. Same 9 tests fail on any PR opened against `main` today.
- **Fix belongs in #536, not here.** #536 emits standalone comments + pins `tomlkit`. I deliberately did **not** duplicate that fix into this lockfile-only PR — it would just conflict with #536 on the same lines.

⚠️ **Blocker update:** #536 is currently a **draft** and has a failing `test` check, so it can't land yet. This PR is therefore blocked on #536 being de-drafted and made green. Recommended order: land #536 first, then rebase + `uv lock` here (which also turns this PR's build green). Happy to do that re-lock the moment #536 merges.

bigcat88:
This is byte-identical to #543 (verified: identical diffs), which just merged — so this branch's delta vs main is now empty. Leaving open for you to close. Same for #534.

PR #547, opened 2026-07-17
fix: escape square brackets in Typer help strings so choice lists render

mattmillerai:
## ELI-5

The CLI's `--help` was supposed to tell you which values an option accepts — `[remote|local|cache]`, `[all|comfy|cli]`. It didn't. Rich (the library that draws those pretty help panels) reads square brackets as *styling markup*, the same way `[bold]` means "make this bold". So it looked at `[remote|local|cache]`, decided `remote|local|cache` must be a style name, and quietly deleted the whole thing. You got an empty cell.

This escapes the opening bracket at the 16 places it happened, so Rich prints the brackets instead of eating them.

## What changed

One character per site — `help="[a|b|c]"` → `help="\\[a|b|c]"` — across 16 argument/option help strings:

- `comfy_cli/command/custom_nodes/command.py` (15): the `--mode` `[remote|local|cache]` help on 11 `node` subcommands, plus the `nodes` argument help on `update` / `disable` / `enable` / `fix`.
- `comfy_cli/cmdline.py` (1): the `target` argument help on `comfy update`.

No runtime behavior changes — these are help strings only. Only the opening `[` needs escaping; the closing `]` is already literal.

## Verification

Since no test asserts on these strings, the real check is rendering the actual help. **All 16 sites confirmed rendering**, not just the 3 named in the issue:

```
$ comfy node show --help
│ --mode           TEXT  [remote|local|cache]          │   # was: empty cell

$ comfy node update --help
│ *  nodes  NODES...  [all|List of custom nodes to update] [required]

$ comfy update --help
│    target  [TARGET]  [all|comfy|cli] [default: comfy]
```

All 11 `[remote|local|cache]` sites verified by rendering each owning command (`show`, `simple-show`, `install`, `reinstall`, `uninstall`, `update`, `disable`, `enable`, `fix`, `install-deps`, `deps-in-workflow`) — 11/11 render.

- `pytest tests/comfy_cli -q` → **2471 passed, 13 skipped** (matches baseline exactly)
- `ruff format` → clean; `ruff check` → **14 errors, identical to the pre-change baseline on `main`** (all pre-existing, none in touched lines)

## Judgment calls / notes

**1. Two additional bracket sites exist that the issue's grep (`help="\[`) can't see — both deliberately left alone.** That pattern anchors to a *leading* bracket; a broader `help=("|')[^"']*\[` sweep found two more:

- `cmdline.py:1133` — `help="Launch ComfyUI: ?[--background] ?[-- <extra args ...>]"`. **Not a bug — verified by rendering.** `comfy launch --help` prints its brackets intact, because Rich only treats `[...]` as a tag when the contents can parse as a style name; `--background` starts with `-`, so Rich leaves it literal. This is also *why* the 16 sites in this PR do break: `[remote|...]`, `[all|...]` start with a letter. I nearly "fixed" this one before checking.
- `cmdline.py:611` — `@app.command(help="Update ComfyUI Environment [all|comfy|cli]")`. This one **is** genuinely stripped on `main` today (renders as `Update ComfyUI Environment`), but it is the exact site #520 already fixes, so touching it here would collide. Left to #520.

**2. Verified this does not conflict with #520**, which is open and touches both of these files with hunks landing within a few lines of two of mine. Test-merged #520 into this branch locally: **merges cleanly, no conflicts**, and both fixes render correctly side by side (`update` panel help *and* `--mode` both show their choice lists). No coordination needed on merge order.

**3. Escape spelling.** Written `"\\[..."` (normal string → the string holds a literal `\[`), not `"\["`. The latter renders identically but is an invalid escape sequence — it trips ruff `W605` (+16 errors over baseline) and is a future `SyntaxError`. Worth noting since `"\["` is the tempting shorter form.

**4. No Linear ref in this title/body**, per the public-repo convention for this repo; the ticket autolinks via the branch name. Flagging because sibling PRs (#490, #520) do carry `(BE-…)` suffixes, so this is an inconsistency someone may want to settle either way.

**5. Not done (explicitly out of scope per the issue):** converting these to a real `enum` / `click.Choice` so Typer renders the values itself. Several duplicate an existing `autocompletion=` choice list, so that dedup is real — but it's a behavior change and should be decided separately.

### DIFF
--- a/comfy_cli/cmdline.py
+++ b/comfy_cli/cmdline.py
@@ -613,7 +613,7 @@ def install(
 def update(
     target: str = typer.Argument(
         "comfy",
-        help="[all|comfy|cli]",
+        help="\\[all|comfy|cli]",
         autocompletion=utils.create_choice_completer(["all", "comfy", "cli"]),
     ),
 ):
--- a/comfy_cli/command/custom_nodes/command.py
+++ b/comfy_cli/command/custom_nodes/command.py
@@ -515,7 +515,7 @@ def show(
     ] = None,
     mode: str = typer.Option(
         None,
-        help="[remote|local|cache]",
+        help="\\[remote|local|cache]",
         autocompletion=mode_completer,
     ),
 ):
@@ -540,7 +540,7 @@ def simple_show(
     ] = None,
     mode: str = typer.Option(
         None,
-        help="[remote|local|cache]",
+        help="\\[remote|local|cache]",
         autocompletion=mode_completer,
     ),
 ):
@@ -595,7 +595,7 @@ def install(
     ] = False,
     mode: str = typer.Option(
         None,
-        help="[remote|local|cache]",
+        help="\\[remote|local|cache]",
         autocompletion=mode_completer,
     ),
 ):
@@ -664,7 +664,7 @@ def reinstall(
     ] = None,
     mode: str = typer.Option(
         None,
-        help="[remote|local|cache]",
+        help="\\[remote|local|cache]",
         autocompletion=mode_completer,
     ),
 ):
@@ -709,7 +709,7 @@ def uninstall(
     ] = None,
     mode: str = typer.Option(
         None,
-        help="[remote|local|cache]",
+        help="\\[remote|local|cache]",
         autocompletion=mode_completer,
     ),
 ):
@@ -748,7 +748,7 @@ def update_node_id_cache():
 def update(
     nodes: list[str] = typer.Argument(
         ...,
-        help="[all|List of custom nodes to update]",
+        help="\\[all|List of custom nodes to update]",
         autocompletion=node_or_all_completer,
     ),
     channel: Annotated[
@@ -769,7 +769,7 @@ def update(
     ] = None,
     mode: str = typer.Option(
         None,
-        help="[remote|local|cache]",
+        help="\\[remote|local|cache]",
         autocompletion=mode_completer,
     ),
 ):
@@ -785,7 +785,7 @@ def update(
 def disable(
     nodes: list[str] = typer.Argument(
         ...,
-        help="[all|List of custom nodes to disable]",
+        help="\\[all|List of custom nodes to disable]",
         autocompletion=node_or_all_completer,
     ),
     channel: Annotated[
@@ -798,7 +798,7 @@ def disable(
     ] = None,
     mode: str = typer.Option(
         None,
-        help="[remote|local|cache]",
+        help="\\[remote|local|cache]",
         autocompletion=mode_completer,
     ),
 ):
@@ -812,7 +812,7 @@ def disable(
 def enable(
     nodes: list[str] = typer.Argument(
         ...,
-        help="[all|List of custom nodes to enable]",
+        help="\\[all|List of custom nodes to enable]",
         autocompletion=node_or_all_completer,
     ),
     channel: Annotated[
@@ -825,7 +825,7 @@ def enable(
     ] = None,
     mode: str = typer.Option(
         None,
-        help="[remote|local|cache]",
+        help="\\[remote|local|cache]",
         autocompletion=mode_completer,
     ),
 ):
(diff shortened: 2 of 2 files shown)


mattmillerai:
## Review pass: no code changes needed — both CI failures are pre-existing and unrelated

Re-reviewed this PR. No merge conflict (`MERGEABLE`), no open review threads, Cursor's panel reported ✅ no high-signal findings, and CodeRabbit never actually reviewed (it hit a Fair-Usage rate limit — its green check is `Review skipped: no matching review label`, not an approval).

The two red checks are **not caused by this PR**. This branch touches only Typer `help=` strings — 2 files, 16 single-character escapes, zero runtime behavior — and neither failure involves that code.

### `build` — 9 failures, upstream `tomlkit` 0.15 regression

`ValueError: Comment cannot contain line breaks` from `config_parser.py`, a file this PR does not touch (byte-identical to `main`). Isolated it to the dependency:

- `tomlkit` is unpinned (`pyproject.toml:49`); `uv.lock` pins `0.13.3` → `test_config_parser.py` **81 passed** locally.
- Force-installing `tomlkit==0.15.1` into the same venv reproduces **CI's exact 9 failures**.
- Minimal repro: `tomlkit.array().comment('\n# x = 1\n')` → `ValueError` on 0.15.1, fine on 0.13.3.

tomlkit 0.15 started rejecting comments containing line breaks; `config_parser.py:75` and `:250` both pass multi-line strings to `.comment()`. CI installs via `pip install -e '.[dev]'`, which ignores `uv.lock` and resolves the newest tomlkit — so CI is red while uv-locked local runs are green. `main` was last green 2026-07-13, before that release; the first red runs are today.

### `test` (Windows) — infra, fails on every PR

Dies in **Install Dependencies**, before any repo test runs: `failed to remove file ..._pydantic_core.cp312-win_amd64.pyd: Access is denied. (os error 5)`, which corrupts the install and yields `ImportError: cannot import name '__version__' from 'pydantic_core'`. Windows file-locking on a loaded `.pyd`. Fails on **8/8** recent runs across unrelated branches.

Both are repo-wide and block every open PR, so I've proposed them as separate follow-ups rather than smuggling a dependency-compat fix into a help-string PR, where it would muddy review and revert.

### Verification of this PR's own diff

- **All 16 sites render**, checked individually (11 × `[remote|local|cache]`, 4 × `nodes` argument help, 1 × `update` `target`).
- **Counterfactual confirmed** — reverting the two files to `origin/main` renders those cells *empty*; with the fix they show the choice lists. The fix is real, not cosmetic.
- `pytest tests/comfy_cli` → **2471 passed, 13 skipped** (matches baseline).
- `ruff format` clean; `ruff check` → 14 errors, **diffed against `main` and byte-identical — zero new**.

mattmillerai:
**Correction to my comment above:** both failures already have open fix PRs — no new follow-ups needed, and they independently corroborate the diagnosis:

- **`build` / tomlkit 0.15** → #533 and #536 both rewrite the `config_parser.py` hint blocks as single-line comments; #537 (`ci: install test dependencies from uv.lock instead of resolving fresh`) targets the exact mechanism I traced — CI resolving a fresh tomlkit because `pip install -e '.[dev]'` ignores `uv.lock`.
- **`test` / Windows `pydantic_core`** → #535 (`build telemetry providers lazily so comfy install can upgrade pydantic_core`) matches the root cause in the traceback: the locked `.pyd` is held via `tracking.py:13 → from mixpanel import Mixpanel → pydantic`. #529 (`ci: set UV_LINK_MODE=copy in the Windows workflow`) is the standard remedy for the `Access is denied (os error 5)` replace failure.

This PR needs no changes and stays ready to merge; its own CI should go green once those land. I did not merge anything.

bigcat88 (approved):
Verified by rendering: on main `comfy node show --help` shows an empty --mode cell; on this branch (with main merged in) it renders `[remote|local|cache]`, and `comfy update --help` shows `[all|comfy|cli]` — no stray backslashes. Full unit suite green (2476 passed). The failing build/test checks are the pre-existing tomlkit 0.15 CI breakage fixed by #569, unrelated to this diff. Nice catch on the launch-help false positive (Rich only eats bracket contents that parse as style names).

PR #550, opened 2026-07-17
feat(jobs): surface text/STRING node outputs in local jobs status as text_outputs (BE-3346)

mattmillerai:
## ELI-5

When a ComfyUI workflow has a node that produces **text** — like a Gemini node describing an image, or a ShowText node — that text is stored by ComfyUI but `comfy jobs status` never showed it. You could only get it by manually `curl`-ing the server's `/history` endpoint. This PR makes `comfy jobs status` include those text outputs, so tools built on comfy-cli (like comfy-local-mcp) can read captions/descriptions/VLM tags without shelling out to `curl`.

## What changed

- **New pure helper** `extract_text_outputs(record)` in `comfy_client.py`, next to `extract_output_entries`. For each `node_id` in the history record's `outputs`, when `node_output["text"]` is a list, it keeps the `str` items and returns `{node_id: [text, ...]}`, omitting nodes with no text. Dict-shape tolerant (non-dict `outputs`/node → skipped, never raises).
- **Wired into the local `jobs status` snapshot** (`_snapshot`) as an always-present additive key `"text_outputs"`: `extract_text_outputs(body)` on the history branch, `{}` on the queued/running branch. All existing keys (`outputs`, `outputs_by_node`, `outputs_by_item`, `error`, `status`) are unchanged — additive only.
- **Pretty renderer** shows a bounded per-entry preview (first line, ~120 chars, `[node_id] <preview>`). The `--json` path emits the full untruncated strings, exactly like the existing `error` payload.

## Why

Text outputs (`GeminiNode` image descriptions, `ShowText`, anything emitting `ui.text`) live in `GET /history/<prompt_id>` under `outputs[node_id]["text"]` as bare strings, but the media-key flatten (`images`/`gifs`/`videos`/`audio`/`files`, each requiring a dict with `filename`) drops them. This blocks local LLM/vision workflows (captioning, VLM tagging, prompt expansion) through the comfy-cli → comfy-local-mcp toolchain.

Because comfy-local-mcp's `job_status` returns comfy-cli's envelope `data` verbatim, this additive key flows through the MCP with **zero code changes** there (a follow-up will just update the MCP docstring + version floor).

## Scope / judgment calls

- **Out of scope (v1), per the ticket:** the cloud snapshot (`_cloud_status_snapshot` — belongs to the cloud sibling ticket), `comfy download` materializing text to files, and other UI-output keys (`string`, etc.).
- `_wait_fetch_snapshot` (`jobs wait`) is intentionally untouched: when the server is up it reuses `_snapshot` (inheriting the new key); its on-disk fallback is a minimal shape and out of scope for this `jobs status` ticket.
- Empty strings under `text` are preserved as legitimate (empty) text outputs rather than silently dropped; only non-`str` items in the list are filtered out.

## Testing

- New unit tests for `extract_text_outputs`: text present / absent / non-list `text` / mixed-type items (only strings kept) / non-dict `outputs` & node / int node-id coercion.
- New `_snapshot` tests: `text_outputs` grouped-by-node with full untruncated strings on the history branch; `{}` on the no-text history branch; `{}` on the queued/running branch.
- New pretty-render test asserting the per-entry preview is truncated (first line only, `…` marker) while the full text is not rendered in pretty mode.
- `pytest` (2583 passed, 37 skipped), `ruff check .`, and `ruff format --check .` all pass (verified against the CI-pinned ruff `0.15.15`).


bigcat88:

mattmillerai:
Rebased onto latest `main` (merge commit 278ccc3, resolving the conflicts @bigcat88 flagged) and pushed a follow-up fix for the Cursor review findings that came through in the raw (unparsed) judge output on the text-output preview added by this PR:

- Preview cells were built as plain f-strings handed to a Rich table, so server-supplied node text containing `[...]` would be interpreted as markup instead of rendered literally — now built as `rich.text.Text` so it's always literal.
- The "first line" guard returned an empty string when text started with a blank line (`\n\nactual content` → blank preview) — now strips before splitting so the first *non-blank* line surfaces.
- The preview loop was unbounded despite the "bounded preview" comment (only per-entry chars were capped, not entry count) — now capped at `_TEXT_PREVIEW_LIMIT` (20) entries with an "N more" note.

Added regression tests for all three (`test_render_status_pretty_text_preview_skips_leading_blank_lines`, `_is_not_rich_markup`, `_bounds_entry_count`). Full local suite (3657 passed) and ruff check/format are green.

skishore23 (approved):
Approving. Clean additive feature. **One conflict to resolve before merge**, and it's an interesting one.

## The conflict is worth reading carefully

`comfy_cli/command/jobs.py` conflicts because `main` hardened the adjacent line:

```python
HEAD:        tbl.add_row("outputs", "\n".join(snap["outputs"]))
origin/main: tbl.add_row("outputs", "\n".join(sanitize_markup(o) for o in snap["outputs"]))
```

**Resolution: take `main`'s sanitized line, keep your `text_outputs` block after it.** I did that locally — `ruff check` clean, **204 passed** (the 3 failures are `test_ls_default_scopes_state_rows_to_local` / `test_ls_all_restores_the_union_view` / `test_ls_watch_mirrors_one_shot_scope`, which fail identically on plain `origin/main` — pre-existing, unrelated).

Do **not** resolve by keeping HEAD wholesale — that would silently revert `main`'s markup hardening on the `outputs` row.

## Nice: you'd already solved the same hazard independently

Your comment says the preview is built as `Text` (not markup strings) "since node ids / node text are server-supplied and may contain `[...]` that Rich would otherwise interpret as style markup." I verified that holds — rendered a hostile payload through the `Text` path:

```
input : node_id="[red]evil", text="[bold blink]PWNED[/] and [link=http://x]click[/link]"
output: [[red]evil] [bold blink]PWNED[/] and [link=http://x]click[/link]
```

Markup is **not** interpreted. So both halves of the merged function are safe — `main` via `sanitize_markup()`, yours via `Text()`.

**Follow-up worth filing:** after this lands, one function guards the same hazard two different ways. Both are correct, but converging on one (probably `sanitize_markup`, since it's the repo-wide helper) would make it obvious to the next person that the guard isn't optional.

## Verified

- **194 passed** as-authored (`tests/comfy_cli/jobs/test_jobs.py`, `test_transfer_download.py`); CI green.
- `extract_text_outputs` is pure and genuinely shape-tolerant — non-dict `outputs`, non-dict node entry, and non-list `text` all skip rather than raise, matching `extract_output_entries`'s posture. Node ids are coerced with `str(node_id)`, so int keys from the history JSON group correctly.
- The empty-string handling matches what you describe: `text: [""]` → `strings == [""]` → truthy → **kept**, while `text: []` → omitted. That's the right split — an empty caption is a result, an absent one isn't.
- Additive-only on the snapshot: `text_outputs` is always present (`{}` on the queued/running branch), so MCP consumers reading the envelope verbatim get a stable shape rather than a sometimes-missing key. That's what makes the zero-change flow-through to comfy-local-mcp actually true.
- Truncation lives only in the pretty path; `--json` emits full strings — consistent with how `error` already behaves.

Scoping the cloud snapshot, `comfy download` text materialization, and other `ui` keys out of v1 is the right call.

PR #554, opened 2026-07-17
fix(run): canonical credential guidance — hint cloud login / set-key, unhide set-key (BE-3362)

mattmillerai:
## ELI-5

When you try to run a workflow that needs a Comfy Cloud / partner-API key and you don't have one, the CLI used to tell you to run `comfy auth set comfy-cloud-api-key --key …`. That command only "works" as a side effect — `auth set` actually rejects the cloud provider and prints a "not a well-known provider; stored anyway" warning. The real, sanctioned command is `comfy cloud set-key` (or better, `comfy cloud login`). But `comfy cloud set-key` was `hidden=True`, so you couldn't even find it in `comfy cloud --help`.

This PR fixes the steering: the missing-credential hints now point at the canonical path (`comfy cloud login` → `COMFY_API_KEY` env → `comfy cloud set-key`), and `set-key` is now discoverable in the help output.

## What changed

1. **`comfy_cli/command/run/__init__.py`** — the `partner_node_requires_credential` hint now reads:
   `run: comfy cloud login   (or set COMFY_API_KEY in the environment, or persist a key with `comfy cloud set-key --key …`; cloud runs auto-inject via --where cloud)`. This is the hint comfy-local-mcp passes through verbatim to agents (BE-3344), so it's the primary agent-facing steering — it now names the canonical order.
2. **`comfy_cli/cloud/command.py`** — removed `hidden=True` from `set-key` so it appears in `comfy cloud --help`. Help text keeps the "canonical sign-in is `comfy cloud login`" / testing-path framing. **Not** added to `SUPPORTED_PROVIDERS` / `comfy auth list` (the store deliberately excludes cloud; OAuth-first per `credentials.py`).
3. **`comfy_cli/error_codes.py`** — the registry entry for the same error code, updated to the canonical order (dropped the `comfy auth set comfy-cloud-api-key` reference).
4. **`comfy_cli/help_json.py`** (judgment call) — the agent copy/paste example for `comfy auth set` was `comfy auth set comfy-cloud --key sk-…`, which is the exact path that ERRORS (`auth_use_login_for_cloud`). Changed to a real third-party provider (`civitai`), matching the sibling example on line 33. This is squarely the ticket's "steer away from the rejected back door" intent.
5. **`comfy_cli/credentials.py`** (judgment call) — dropped the now-stale word "hidden" from two docstring/comment references to `comfy cloud set-key` (accuracy only, tied to unhiding).

## Tests

- Added `test_cloud_set_key_visible_in_help` — asserts `set-key` appears in `comfy cloud --help`.
- The existing `test_auth_set_comfy_cloud_rejected` already asserts `auth set comfy-cloud` still errors with `auth_use_login_for_cloud` and hints `comfy cloud login` — left intact (deliberately unchanged; the rejection path is preserved).
- No test asserted the old hint string (grep of `tests/` for `comfy auth set comfy-cloud-api-key` was empty).
- Ran the auth / cloud / run / output / credentials / help_json suites — all green; `ruff check` + `ruff format --check` clean on the touched files.

## Notes / judgment calls

- Items 4 and 5 are beyond the literal grep in the ticket but directly serve its goal (canonical guidance + no undiscoverable/misleading back-door references). Flagged here for review.
- **Negative-claim check:** this change adds no capability denial — the `partner_node_requires_credential` and `auth_use_login_for_cloud` deny/reject paths already existed; the diff only improves the guidance *toward* working paths (`cloud login` / `set-key`), so there is no dead-end to falsify.
- Kept `set-key` out of `SUPPORTED_PROVIDERS` / `comfy auth list` per the ticket (OAuth-first, deprecation-path stance preserved).


skishore23 (approved):
LGTM

PR #560, opened 2026-07-18
fix: comfy generate refresh fetches /openapi with /openapi.yml fallback (BE-3391)

mattmillerai:
## ELI-5

`comfy generate refresh` downloads the latest list of partner AI models so the CLI knows about brand-new ones. It was asking the server for a file at `/openapi.yml` — but that address now returns **404 Not Found**. The real, up-to-date spec lives at `/openapi` (no extension, served as JSON). So `refresh` — the *only* way to update the model catalog — was completely broken. This points it at the working address (and keeps the old one as a fallback), and refuses to save anything that isn't a real spec so a bad download can't silently break `comfy generate` for the next 7 days.

## What changed

- **`_refresh()` (`app.py`)** now fetches `spec.base_url() + "/openapi"` first, and **falls back to `/openapi.yml` on a 404** — preserving any custom `COMFY_API_BASE_URL` deployment that still serves the `.yml` path.
- **Validate before caching:** the fetched body is parsed with the *same* loader `spec.load_raw_spec()` uses (`validate_spec_text()`, new helper in `spec.py`) and must contain a top-level `paths` mapping. JSON is a subset of YAML 1.2 and the custom loader only restricts bool resolution, so the JSON body served at `/openapi` parses cleanly. If the body doesn't parse or has no `paths`, `refresh` exits 1 **without writing the cache** — previously a `200`-with-garbage response would have been cached into `~/.comfy/openapi-cache.yml` and poisoned every subsequent `comfy generate` for `CACHE_TTL_SECONDS` (7 days).
- **`spec.write_cache()` is unchanged** — the cache file staying `.yml` is fine since the loader reads JSON.

## Verification

- **End-to-end against the real `api.comfy.org` (the acceptance criterion):**
  - `comfy generate refresh` → fetched `/openapi` (HTTP 200, ~973 KB JSON), wrote `~/.comfy/openapi-cache.yml`, printed `Refreshed model catalog`.
  - `comfy generate list` afterwards read the refreshed cache and rendered the model table (including `flux-2`, which is ahead of the vendored `spec/openapi.yml`).
  - Confirmed the old path is dead: `curl -so /dev/null -w '%{http_code}' https://api.comfy.org/openapi.yml` → `404`; `/openapi` → `200 application/json`.
- **Unit tests** (`tests/comfy_cli/command/generate/`):
  - (a) refresh hits `/openapi` and writes the cache on a valid JSON spec body;
  - (b) a 404 on `/openapi` falls back to `/openapi.yml`;
  - (c) an invalid body (non-parse **or** 200-with-no-`paths`) exits 1 and leaves a pre-existing cache byte-for-byte untouched;
  - (d) a cached JSON body round-trips through `spec.load_raw_spec()` and `spec.get_endpoint()` resolves an endpoint from it.
- `ruff format --check` + `ruff check` clean on the changed files; full suite green (`2573 passed, 37 skipped`).

## Notes / judgment calls

- **Chesterton's fence — changing the fetch URL is intentional, not a silent regression.** The old default (`/openapi.yml`) is empirically 404 today; the change is authorized by the ticket and the old path is retained as a 404 fallback, so custom deployments serving the `.yml` path keep working.
- **Test isolation:** added an autouse `_isolate_spec_caches` fixture to `test_app.py` that clears the module-level `load_raw_spec` / `_registry` lru_caches around each test. The new refresh tests monkeypatch `spec._USER_CACHE` to a temp file, and `base_url()` loads that temp spec into the process-global caches; without the clear, a temp spec leaked into later tests (empty registry → spurious "No models match" failures). The fixture adds a small re-parse cost per test but keeps the suite deterministic.


bigcat88 (approved):
Verified end-to-end against the real api.comfy.org with main merged in: /openapi.yml is 404 and /openapi is 200 as claimed; main's _refresh fails with the 404, this branch's refresh fetches /openapi, validates, and writes the ~980KB cache, and comfy generate list then renders the refreshed catalog including flux-2 (ahead of the vendored spec — proving the refresh is real). Generate suite (177) and full unit suite (2493) green. The 404-only fallback, validate-before-cache via the same loader load_raw_spec uses, and the lru_cache isolation fixture are all sound. Two housekeeping notes: the uv.lock hunk in the diff is the #543 relock already on main (drops out on rebase), and #517 covers the same fix without the fallback/validation — worth closing it in favor of this. Failing build/test checks are the pre-existing tomlkit 0.15 CI issue fixed on main's CI by #569 (root-cause code fix reviewed in #533).

PR #563, opened 2026-07-18
fix(run): close the run WebSocket on all exit paths of local --wait (BE-3404)

mattmillerai:
## ELI-5

When you run `comfy run --wait` against a local ComfyUI server, the CLI opens a WebSocket to watch the job. On the happy path that socket was never explicitly closed — it lingered open (holding a server-side ws session) until the Python interpreter tore down at exit. This adds a best-effort close in the function's existing `finally:` block so the socket is shut on **every** exit path — success, error, or interrupt.

## What & why

On the local `comfy run --wait` success path, the WebSocket opened by `execution.connect()` was only closed via the SIGINT cancellation token (`token.on_cancel(lambda: _safe_close(execution))`). After `watch_execution()` returned and the terminal envelope was emitted, the socket stayed open until interpreter teardown — a resource-hygiene defect surfaced during BE-3354.

The fix adds `_safe_close(execution)` to the `execute()` `finally:` block (which already stops the Rich progress):

```python
finally:
    if progress is not None:
        progress.stop()
    _safe_close(execution)
```

## Why it's safe

- `execution` is constructed before the `try:`, so it is always bound when `finally` runs.
- `_safe_close` (in `execution.py`) already guards `ws is None` and swallows exceptions, so it is a **no-op** on the async (no-`--wait`) path where `connect()` never ran, and **idempotent** with the SIGINT-token close (which is left untouched — it's what breaks the blocking `recv()` on Ctrl-C).
- Early exits (workflow-load errors, server-not-running, `--print-prompt`) occur before `execution` is constructed and before any ws is opened, so there is nothing to leak there.

## Tests

- `test_successful_execution` now asserts `mock_exec.ws.close.assert_called_once()` after a successful `--wait` run.
- New `test_websocket_closed_on_watch_failure` asserts `close()` is also called when `watch_execution` raises (`WebSocketTimeoutException`).
- Full run-module suite green: `pytest tests/ -k run` → 314 passed, 2 skipped.

## Verification note

The unit tests assert `close()` is invoked on both the success and watch-failure paths (non-vacuous: absent the fix, `close()` is never called on the success path since the only other wiring is the SIGINT `on_cancel`, which doesn't fire without a signal). The live `lsof`/server-log confirmation in the ticket's Verification section requires a running ComfyUI server and was not run in this unattended environment.


bigcat88 (approved):
Verified locally with main merged in: full unit suite green (2494 passed). Confirmed in source that execution is bound before the try, _safe_close guards ws-None and swallows exceptions (no-op on the async path, idempotent with the SIGINT close), and both new assertions fail against main's run/__init__.py — non-vacuous coverage. Failing build/test checks are the pre-existing tomlkit 0.15 CI issue fixed by #569.

PR #568, opened 2026-07-18
fix(hardware): detect Apple Silicon under Rosetta 2 in GPU probe (BE-3435)

mattmillerai:
## ELI-5

If you run an Intel-flavored Python on a modern Apple Mac (via Apple's Rosetta 2 translator), the machine *lies* and says it's an old Intel chip. Our hardware probe believed the lie and reported "no GPU" — even though the Mac has a perfectly capable Apple GPU. This teaches the probe to ask macOS one extra question ("am I being translated?") so it correctly recognizes the Apple chip underneath and reports its unified-memory GPU.

## What & why

`comfy_cli/hardware.py:_detect_gpu` keyed Apple Silicon unified-memory detection on `machine == 'arm64'`. An x86_64 Python running under **Rosetta 2** on an Apple Silicon Mac reports `platform.machine() == 'x86_64'`, so the unified-memory block was skipped and the machine fell through to the NVIDIA/AMD probes (which find nothing) — a capable GPU reported as `gpu: null`.

Fix: a small `_is_apple_silicon(machine)` helper keeps the native `arm64` fast path and, when `machine != 'arm64'`, falls back to `sysctl -n sysctl.proc_translated` (via the existing `_run` helper). A value of `'1'` means the process is translated, which only happens on Apple Silicon, so the apple/unified block is used.

## Notes for the reviewer

- **Chesterton's Fence — this deliberately broadens an existing check.** The original `machine == 'arm64'` guard was introduced (intentionally minimal) in the base PR #562 (BE-3399); this ticket (BE-3435) is the scoped follow-up to broaden it. The old and new conditions differ **only** in the newly-matched state — `machine != 'arm64'` **and** `proc_translated == '1'`, i.e. exactly Apple Silicon under Rosetta. On a genuine Intel Mac the `sysctl.proc_translated` OID is absent, `_run` returns `None`, `None == '1'` is False, so Intel behavior is **unchanged** (the existing `test_intel_mac_has_no_apple_gpu` still passes).
- **Never-raise contract preserved:** `_is_apple_silicon` only calls `_run` (which never raises and is `timeout`-bounded) plus a string compare, and it runs inside `_detect_gpu`'s existing `try/except`.
- Under Rosetta, `arch` is still honestly reported as `x86_64` (that's what the process is); only the `gpu` block is corrected. The `model` comes from `machdep.cpu.brand_string`, which returns the real Apple chip name even under translation.
- New unit test mocks `_run` to return `'1'` for the `sysctl.proc_translated` key and asserts the unified block.

## Stacking

Opened **stacked on `matt/be-3399-hardware-block` (#562)** — `hardware.py` does not exist on `main` yet; it lands with #562. GitHub will retarget this PR to `main` when #562 merges; the diff here is just the net Rosetta change.

## Testing

- `pytest tests/comfy_cli/test_hardware.py` → 20 passed
- `ruff check` + `ruff format --check` on the touched files → clean

### DIFF
--- a/.github/workflows/build-and-test.yml
+++ b/.github/workflows/build-and-test.yml
@@ -39,30 +39,30 @@ jobs:
       - name: Check out code
         uses: actions/checkout@v7
 
-      - name: Set up Python
-        uses: actions/setup-python@v6
+      - name: Install uv
+        uses: astral-sh/setup-uv@11f9893b081a58869d3b5fccaea48c9e9e46f990 # v8.3.2
         with:
+          enable-cache: true
           python-version: ${{ matrix.python-version }}
 
+      # `--locked` fails the job if uv.lock has drifted from pyproject.toml, so test
+      # dependency versions can only change via a deliberate lockfile bump.
       - name: Install Dependencies
-        run: |
-          python -m pip install --upgrade pip
-          pip install pytest
-          pip install -e .
+        run: uv sync --extra dev --locked
 
       - name: Test CUDA auto-detection (skips if no GPU)
         run: |
-          pytest tests/comfy_cli/test_cuda_detect_real.py -v
+          uv run --locked --extra dev pytest tests/comfy_cli/test_cuda_detect_real.py -v
 
       - name: Test e2e
         env:
           PYTHONPATH: ${{ github.workspace }}
           TEST_E2E: true
         run: |
-          pytest tests/e2e
+          uv run --locked --extra dev pytest tests/e2e
 
       - name: Test torch backend compilation
         env:
           TEST_TORCH_BACKEND: "true"
         run: |
-          pytest tests/uv/test_torch_backend_compile.py -xvs
+          uv run --locked --extra dev pytest tests/uv/test_torch_backend_compile.py -xvs
--- a/.github/workflows/ci-cursor-review.yml
+++ b/.github/workflows/ci-cursor-review.yml
@@ -29,7 +29,7 @@ jobs:
     # SHA-pinned. Bump this SHA to pick up upstream changes; keep
     # `workflows_ref` matching so prompts/scripts load from the same commit as
     # the workflow definition.
-    uses: Comfy-Org/github-workflows/.github/workflows/cursor-review.yml@047ca48febe3a6647608ed2e0c4331b491cb9d6a # github-workflows#9
+    uses: Comfy-Org/github-workflows/.github/workflows/cursor-review.yml@ece4c3eaa71812a5a665f48d5d1f4bb0ff5f655e # github-workflows main (ece4c3e)
     with:
       # Overriding diff_excludes replaces the reusable default wholesale, so
       # this restates the generated/vendored defaults and adds this repo's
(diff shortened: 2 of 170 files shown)


bigcat88:

skishore23 (approved):
Approving. I was able to verify this one on **real affected hardware** rather than mocks — my review machine is exactly the target scenario (Apple M3 Max, x86_64 Python under Rosetta 2).

#562 is merged and `hardware.py` is on `main`, so the stacking note is moot.

## Before/after on genuine Rosetta hardware

```
Machine : Apple M3 Max   (hw.optional.arm64 = 1)
Python  : Mach-O 64-bit executable x86_64  → running translated
platform.machine() → 'x86_64'              → the lie this PR describes

BEFORE (origin/main):
  arch: x86_64 | cpu: Apple M3 Max
  gpu : None                                    ← the bug

AFTER (this PR):
  _is_apple_silicon('x86_64') -> True
  arch: x86_64 | cpu: Apple M3 Max
  gpu : {'vendor': 'apple', 'model': 'Apple M3 Max',
         'vram_bytes': None, 'unified_memory': True}
```

So the reported failure is real, reproduces on `main`, and this fixes it. Worth noting the bug is live for anyone on an Apple Silicon Mac running an x86_64 Python — which includes the default x86_64 miniconda install, a very common setup. This is more impactful than "Rosetta edge case" suggests.

## One thing I checked carefully and want to record

`sysctl.proc_translated` reports whether **the calling process** is translated, so an obvious worry is that the `sysctl` *child* might execute natively (arm64) and honestly report `0` for itself — which would make the probe silently useless in exactly the case it targets. It doesn't:

```
in-process sysctlbyname()      : 1
subprocess ['sysctl','-n',...] : 1
```

A translated parent's children inherit translation, so the subprocess answer matches the in-process one. (My first probe *did* return `0` — because I ran it from the native arm64 shell, not from Python. Easy trap; the code is correct.)

If you ever want to drop the subprocess entirely, `ctypes.CDLL("libc.dylib").sysctlbyname(b"sysctl.proc_translated", ...)` gives the same answer in-process — but the current form correctly reuses the never-raises, timeout-bounded `_run`, which is the better fit for this module's contract. Not worth changing.

## Other verification

- **34 passed** in `tests/comfy_cli/test_hardware.py` after merging `origin/main` (clean merge).
- **Intel Macs are genuinely unaffected**, and the short-circuit is real — instrumented `_run` and confirmed `sysctl.proc_translated` is **never invoked** on `system='Linux'`. Python's `and` short-circuits before `_is_apple_silicon`, so the only platform paying for the extra sysctl is non-arm64 Darwin. On a real Intel Mac the OID is absent, `_run` returns `None`, `None == "1"` is False — and `test_intel_mac_has_no_apple_gpu` still passes.
- Never-raise contract holds: the helper only calls `_run` plus a string compare, inside `_detect_gpu`'s existing `try/except`.
- Leaving `arch` as the honest `x86_64` while correcting only the `gpu` block is the right call — the process really is x86_64, and conflating the two would break anything keying off `arch` for wheel selection.

PR #570, opened 2026-07-19
docs: add CLAUDE.md shim importing AGENTS.md (BE-3537)

mattmillerai:
## ELI-5

Claude Code only looks for a file called `CLAUDE.md` for its instructions — it never reads `AGENTS.md` and doesn't fall back to it. This repo keeps its agent instructions in `AGENTS.md` (the cross-agent standard), so Claude Code sessions were getting **zero** instructions. This PR adds a tiny `CLAUDE.md` that just `@`-imports `AGENTS.md`, so Claude Code now picks up the same instructions every other agent already reads.

## What changed

- Added root `CLAUDE.md` containing exactly the org-standard shim (an HTML comment explaining it's a shim + `@AGENTS.md`), per the 2026-07-15 "Pillar 1: Agentic Engineering" guide and the org-wide audit (BE-3241).

## What did NOT change

- `AGENTS.md` is untouched (the single source of truth stays canonical).
- No `.cursorrules` added/removed (none exists; modern tools read `AGENTS.md`).
- No nested `AGENTS.md` files exist, so no nested shims were needed.

## Verification

- `CLAUDE.md` content is byte-for-byte the canonical shim (em-dash preserved as UTF-8).
- Relevant pre-commit hooks pass on the new file: `end-of-file-fixer` ✅, `trailing-whitespace` ✅ (ruff/toml/uv hooks skip — no matching files).
- Doc-only change; no Python/CLI behavior touched, so pytest/ruff/mac/windows suites are unaffected.

Closes BE-3537.


bigcat88 (approved):
Approving. I checked the premise against the official Claude Code docs rather than assuming, and this PR implements the **documented** pattern verbatim.

From `code.claude.com/docs/en/memory`, under a heading literally called "AGENTS.md":

> Claude Code reads `CLAUDE.md`, not `AGENTS.md`. If your repository already uses `AGENTS.md` for other coding agents, create a `CLAUDE.md` that imports it so both tools read the same instructions without duplicating them.
> ```markdown CLAUDE.md
> @AGENTS.md
> ```

Confirmed on the repo side too: `AGENTS.md` exists (65 lines) and there was no `CLAUDE.md` at all before this, so Claude Code sessions really were getting zero repo instructions.

Three mechanics I verified against the docs, all of which this shim gets right:

1. **The import fires.** Import parsing skips code spans and fenced blocks — `` `@AGENTS.md` `` in backticks would stay literal. Yours is bare, so it imports. Relative paths resolve against the file containing the import, and max depth is four hops; one hop here.
2. **Your HTML comment costs zero context.** "Block-level HTML comments in CLAUDE.md files are stripped before the content is injected into Claude's context… Use them to leave notes for human maintainers without spending context tokens on them." So the "don't add content here — edit AGENTS.md" note reaches humans reading the file and never reaches the model. Exactly the intended use, and better than a plain-markdown note would have been.
3. **No approval dialog.** External imports (paths resolving outside the working directory) trigger a one-time approval prompt. `@AGENTS.md` is repo-relative, so teammates get it silently.

The docs also mention `ln -s AGENTS.md CLAUDE.md` as an alternative, but explicitly warn that symlinks need Administrator privileges or Developer Mode on Windows and recommend the `@AGENTS.md` import there. Given this repo's CI runs `windows-latest`, the import form is the right choice — worth knowing you picked the portable one.

Mechanically clean: 2-line pure addition, `ruff check` and `ruff format --diff` clean under the CI-pinned 0.15.15, nothing in the codebase references `CLAUDE.md`, and packaging is unaffected (`pyproject.toml` ships only `README.md` plus `comfy_cli.skills`/`comfy_cli.schemas` markdown, so a root `CLAUDE.md` isn't swept into the wheel).

PR #574, opened 2026-07-22
fix(cql): expand dynamic-combo sub-inputs in slot extraction and set-slot/vary write path (BE-3371)

mattmillerai:
## ELI-5

ComfyUI has "shape-shifter" dropdowns: pick a model in the dropdown and the node grows extra knobs (size, width, height…) that belong to that model. The CLI didn't know these dropdowns existed, so when it counted a node's knobs it skipped the dropdown *and* its extra knobs — every knob after it got the wrong label, and `set-slot` turned the wrong knob (setting the seed actually overwrote the model choice, silently). Now the CLI reads which model is picked, counts that model's knobs in their real positions, and when you switch models it swaps the knob set for the new model's defaults and tells you it did.

## Problem

**Silent graph corruption** on the pristine `api_bytedance_seedream_5_0_pro_t2i` template. Measured
against `origin/main` @ `578e2c8` in a separate clean worktree (necessary — `python script.py`
resolves `comfy_cli` via the editable install rather than the checkout, which makes both sides
look identical):

|                                         | `main` (`578e2c8`) | this PR |
| --                                      | --                 | --      |
| `widget_order()` → `'seed'` index       | **1**              | 2       |
| `widget_order_for_node(…, lite values)` | not present        | seed at 7 |

On `main`, `'seed'` resolves to index **1** — and `widgets_values[1]` is the *model selector*
(`'seedream 5.0 pro'`). So `set-slot 1.seed=42` overwrites the model choice, with no warning.

**Root cause: `widget_order` is value-INDEPENDENT**, while the frontend inlines the *selected
option's* widget sub-values into the flat positional `widgets_values` right after the selector — and
the options carry different sub-input counts (`pro` → 4, `lite`/`4-5`/`4-0` → 6), so no single
static order can be correct for all of them. A value-independent order is off by two for three of
the four options, misaligning every widget after the combo.

Reading the selector from its own positional slot is therefore the only thing that *can* be right,
which is what `widget_order_for_node` does.

> Scope note (corrected in review): `main` already parses `COMFY_DYNAMICCOMBO_V3` as a widget and
> already expands *some* sub-inputs, so the earlier framing of this section — "no `1.model` /
> `1.model.*` slots at all" — is superseded. The **misalignment** is the half that still
> reproduces, and it is what this diff fixes.

## Fix

- **Parsing**: a `COMFY_*COMBO*` type is a widget port. Its enum = the option `key` strings; the raw option dicts are retained on `Port.dynamic_options` for expansion. `_is_scalar_choice` still rejects dict options for ordinary combos.
- **`Graph.widget_order_for_node(class_name, widgets_values)`**: value-aware order — at a dynamic combo it reads the selector from its own positional slot and expands the matching option's widget-like sub-inputs as `<name>.<sub>` (same widget-vs-link test as top level; nested dynamic combos recurse with a depth cap; connection subs like `COMFY_AUTOGROW_V3` contribute no slot; `control_after_generate` markers follow control-flagged inputs, sub-inputs included). `widget_order(class_name)` is kept for callers without node context.
- **Slots**: `_node_widget_slots` emits `N.model` (type `COMFY_DYNAMICCOMBO_V3`, `enum` = option keys, current = selector) plus `N.model.<sub>` slots with correct types/enums/current values. Slots with enum choices now carry an additive `enum` key.
- **Write path** (`set-slot` / `vary`):
  - `N.model.<sub>=v` writes the correct position, validated against the sub-spec.
  - `N.model=<other key>` rebuilds `widgets_values` for that node: values before the combo kept, new selector written, the new option's widget sub-inputs filled from schema defaults (sub-spec `default`, first enum option for combos), trailing values (seed/marker/watermark/thinking) re-aligned — and emits a `dynamic_combo_roster_rebuilt` warning describing the rebuild.
  - `N.model.<sub>` that doesn't exist under the *current* selector → `unknown_dynamic_sub_input` warning, **no write**, listing the valid addresses and how to switch rosters. This denial is schema-grounded, not assumed: under the `pro` selector the frontend serializes no `max_images` slot (the vendored pristine template's 8-value layout confirms it), so the write has no position to land in — the warning redirects to the valid addresses instead of dead-ending.
  - `N.model=<unknown key>` → hard error listing valid options (an unknown selector has no roster to build; writing it would corrupt the positional alignment of everything after — deliberate deviation from ordinary combos, which warn-and-write).
- `_resolve_proxy_value` (curated subgraph proxy reads) migrated to the value-aware order as well.

## Judgment calls

- **Shared vs duplicated widget-ness rules**: the ticket suggested importing the converter's expansion helper (`workflow_to_api`). I kept them separate — the converter operates on raw `INPUT_TYPES` specs, the engine on parsed `Port`s, and a cross-module import would couple the CQL engine to the converter's internals. The rules are aligned (`_is_dynamic_combo_type` mirrors `_is_widget_input`'s `COMFY_*COMBO*` test) and both sides are pinned by tests on the same fixtures.
- **Selector-roster rebuild fills from schema defaults** (as the ticket specifies) rather than carrying over same-named sub-values — options can have divergent enums (lite's `size_preset` list ≠ pro's), so defaults are the safe deterministic choice; the warning tells the agent what was reset.
- Updated one existing test (`test_slot_address_with_dotted_input_name`) to monkeypatch `widget_order_for_node` instead of `widget_order` — its intent (first-dot address split) is unchanged.

## Tests

- `tests/comfy_cli/command/test_workflow_slots.py`: pristine-template slots (addresses/types/enums/currents, `1.seed` current is `0` not the model name), seed write doesn't clobber the selector, sub-input write hits index 2, selector change to lite rebuilds the roster (+2 slots, `max_images`/`fail_on_partial` appear, prompt/seed/watermark preserved, warning emitted, new roster addressable), unknown sub-address warns without mutating, unknown selector rejected, `vary` over a sub-input, autogrow subs not exposed.
- `tests/comfy_cli/cql/test_engine.py`: dynamic combo parses as widget with option-key enum; value-independent vs value-aware order; nested-combo recursion; unknown selector expands nothing; nested + outer selector-change rebuilds.

## Verification

- `uv run pytest tests/comfy_cli/command/test_workflow_slots.py tests/comfy_cli/cql/ tests/comfy_cli/test_workflow_to_api.py` — **326 passed** on a re-run against current `origin/main` (clean merge, no conflicts); 285 when this branch was opened. CI green.
- `ruff check` clean on all touched files; `ruff format` applied (the 15 repo-wide UP038 findings also pre-exist on `origin/main` under the local ruff)

> An earlier revision of this section carried a caveat about two full-suite failures
> (`test_validate_command.py::test_api_format_unchanged` and `::test_empty_dict_payload_unchanged`,
> pre-existing on `origin/main` via BE-3357's `prompt_no_outputs` hard error). **Both pass now** —
> #565 landed in that area — so the caveat was stale in this diff's favour and has been dropped
> rather than carried into the squash message.


bigcat88:

mattmillerai:
Rebased on \`main\` and resolved the conflict — thanks for flagging it, @bigcat88.

The conflict was semantic, not just textual: #617 (\`main\`) and this branch independently added dynamic-combo option expansion to \`Port\` for different consumers — #617's validate path (\`raw_spec\`) vs. this branch's slot-extraction/write path (\`dynamic_options\`). Kept both fields since they're used by disjoint code, and fixed the resulting arity break in \`_check_dynamic_combo_sub\`, which still unpacked \`_parse_input_spec\`'s old 4-tuple after this branch changed it to 5.

Also addressed the two real findings buried in the Cursor review's parse-error output:
- Missing implicit \`seed\`/\`noise_seed\` control_after_generate detection in the new widget-slot expansion (mirrors \`workflow_to_api._has_control_after_generate_companion\`, which already handles this for the converter).
- \`_write_dynamic_combo_selector\`'s roster-rebuild branch wasn't checking \`extend\` before padding.

Added regression tests for both; full suite (3665 passed) and ruff (pinned to the CI version, 0.15.15) are green.

skishore23 (approved):
Approving. I verified the reported corruption **still reproduces on current `main`**, and that this fixes it. Worth recording the evidence, because the bug is a silent data-corruption one and the "it looks fine" reading is easy to reach by accident (I reached it twice before isolating properly).

## The bug is real on today's `main`

Compared pristine `origin/main` (`578e2c8`) against this branch in two **separate clean worktrees** — necessary because `python script.py` resolves `comfy_cli` via the editable install rather than the checkout, which silently makes both sides look identical:

```
                                    main (578e2c8)   this PR
widget_order()        'seed' index        1              2
widget_order_for_node(…, lite values)   NOT PRESENT      7
```

On `main`, `'seed'` resolves to **index 1** — and `widgets_values[1]` is the **model selector** (`'seedream 5.0 pro'`). So `set-slot 1.seed=42` writes over the model choice, exactly as reported, with no warning. That's silent graph corruption on a pristine gallery template.

## Why value-awareness is genuinely required, not just tidier

The options carry **different sub-input counts**, so no single static order can be right:

```
'seedream 5.0 pro'   -> size_preset, width, height, images                                (4)
'seedream 5.0 lite'  -> size_preset, width, height, max_images, images, fail_on_partial   (6)
'seedream-4-5-…'     -> …same 6
'seedream-4-0-…'     -> …same 6
```

A value-independent order is off by two for three of the four options — every widget after the combo (seed, `control_after_generate`, watermark, thinking) misaligns. `widget_order_for_node` reading the selector from its own positional slot is the only thing that can be correct here, and it produces the right lite roster:

```
model.size_preset, model.width, model.height, model.max_images, model.fail_on_partial → seed at 7
```

## Verified

- **326 passed** (`test_workflow_slots.py`, `cql/`, `test_workflow_to_api.py`) after merging current `origin/main` — **clean merge**, no conflicts. CI green.
- The description's "2 pre-existing main failures" caveat is **stale in your favour** — `test_api_format_unchanged` / `test_empty_dict_payload_unchanged` pass now (#565 landed in that area). Drop it from the squash message.
- Note the "Problem" section is also partially stale: `main` already parses `COMFY_DYNAMICCOMBO_V3` as a widget and expands *some* subs (3 `DYNAMICCOMBO` references + the fixture are already there), so the `1.model` slots aren't entirely absent any more. The **misalignment** half is what still reproduces, and that's the part that matters. Worth rewriting so the commit message matches what the diff actually fixes.

## Design calls I agree with

- **Hard-erroring on an unknown selector** rather than warn-and-write (the ordinary-combo behavior) is right and worth the deviation: there's no roster to build, so writing it would corrupt positional alignment for everything downstream. Failing loudly beats writing a graph that silently misexecutes.
- **Filling the new roster from schema defaults** rather than carrying same-named values across a selector change — `lite`'s `size_preset` enum genuinely differs from `pro`'s, so carrying values over could produce an out-of-enum value that only fails server-side. The `dynamic_combo_roster_rebuilt` warning covers the discoverability cost.
- **Not importing the converter's expansion helper.** Agreed — the converter works on raw `INPUT_TYPES`, the engine on parsed `Port`s; coupling them across that boundary would be worse than two aligned implementations pinned by tests on shared fixtures. Just keep `_is_dynamic_combo_type` and `_is_widget_input` commented as mirrors of each other so a future edit to one prompts a look at the other.

mattmillerai:
@skishore23 both asks from your approval are applied to the description — which matters here because the description *is* the squash message, so the fleet merger would otherwise have written the stale text into `main`'s history.

**1. "Drop it from the squash message."** The `2 pre-existing main failures` caveat is gone. Your reading is confirmed: `test_api_format_unchanged` and `test_empty_dict_payload_unchanged` pass now that #565 has landed, so the caveat was stale in this diff's favour. The Verification section now reports your **326 passed** re-run against current `origin/main` alongside the 285 from when the branch opened, and the dropped caveat is recorded in a footnote rather than deleted without trace — so the history shows it was retired deliberately, not quietly.

**2. "Worth rewriting so the commit message matches what the diff actually fixes."** The Problem section is rewritten around the **misalignment**, which is the half that still reproduces, and it now leads with your measurement table (`'seed'` → index 1 on `578e2c8` vs 2 here; `widget_order_for_node` absent on main vs seed-at-7) and your point about the divergent sub-input counts — 4 for `pro`, 6 for the other three — which is the actual argument that no static order can be correct and value-awareness is required rather than merely tidier. The superseded framing ("no `1.model` / `1.model.*` slots at all") is called out as superseded, with the note that `main` already parses `COMFY_DYNAMICCOMBO_V3` as a widget and already expands some subs.

Thanks for the two-worktree isolation detail, incidentally — the editable-install-shadows-the-checkout trap is why the "it looks fine" reading is so easy to reach, and it now sits in the description where the next person to touch this will find it.

No code changed; description only, so CI is unaffected.

---

For the record on why this sat: the merger held it at **G12** — "approval body carries an unresolved ask" — and that was correct. An approval that says "LGTM, but drop X from the squash message" is not a green light to merge X. The gate can't tell a satisfied ask from an unsatisfied one, so it holds and names the fix; the resolution is to do the thing, which is what this comment records.

mattmillerai:
Applied — factored the three \`Port\`-from-spec builders into a shared \`_port_from_spec\` helper (used by \`_parse_inputs\`, \`_check_dynamic_combo_sub\`, and the dynamic-combo sub-port expansion) and deleted \`_sub_port_from_spec\`. Good catch on the drift — that gap was exactly the kind of thing the merge-conflict fixup two commits ago introduced.

PR #577, opened 2026-07-23
feat(generate): spend-gate — consent before credit spend (BE-4103)

mattmillerai:
## ELI-5

`comfy generate flux-pro --prompt "a cat"` costs real money (Comfy credits) the moment you run it — and until now it never asked. This PR makes it ask first. At a terminal you get a "this spends credits — proceed?" prompt; scripts and agents must pass `--yes` (or the user can persist "always proceed" once with `comfy generate consent always`); and a machine call (`--json` or no terminal) with neither errors out cleanly instead of spending silently or hanging on a question nobody can answer.

## What

The money interlock for `comfy generate` (launch blocker for exposing generation through the local MCP, parent BE-4102):

- **Interactive confirm** — TTY runs prompt (`default No`) before any credit-spending proxy call.
- **`--yes`** — explicit bypass for automation/evals (new meta flag, stripped like `--async`/`--json`).
- **`spend.auto_confirm` config** — persistent always-proceed, managed by a new `comfy generate consent [show|always|ask]` action; stored in `config.ini` via `ConfigManager` (see judgment calls).
- **Fail-closed non-interactive** — `--json` or no stdin TTY with neither consent source → error `spend_consent_required` (registered in `error_codes.REGISTRY`), exit 1, nothing spent, never a hanging prompt.

The gate sits in `_generate` **before every network side effect** — before OAuth refresh (`resolve_api_key`), before asset-upload transforms, before the generation request. Ungated because they spend nothing: `list` / `schema` / `refresh` / `upload` / `resume` (polls an already-paid job) / `--emit-workflow` (local artifact, no proxy call).

Consent failures emit `generate:error` with `error_kind="consent"`; the `consent` action emits `generate:consent`. Docs updated: top help, per-model schema help, and the agent-facing `skills/comfy/SKILL.md` (explicitly tells agents **not** to reflexively add `--yes` — it must represent real human approval).

## Acceptance criteria

- `comfy generate <model> --json` with no consent → errors, spends nothing ✅ (test + live CLI run: exits 1 with `spend_consent_required` before any network call)
- `--yes` or `spend.auto_confirm=true` → proceeds ✅ (tests for both, plus config-persistence-to-disk check)
- Interactive TTY → prompts before spending ✅ (accept / decline / bare-Enter-defaults-to-No tests)

## Judgment calls

- **`spend.auto_confirm` lives in `config.ini` (ConfigManager), not `secrets.json`.** The ticket says "same persistence that backs `comfy cloud set-key`", which literally is the auth-record store (`secrets.json`, provider→key schema). A boolean preference doesn't fit that schema; `config.ini` is the CLI's canonical settings persistence (same durability, same directory). Easy to move if the literal store was intended.
- **Added `comfy generate consent [show|always|ask]`** — the ticket asks for the config but no setter; without one, users would hand-edit `config.ini`. Small (~40 lines), makes the acceptance path actually usable, and gives the fail-closed error a concrete remediation to name.
- **Gate before auth resolution**: an unauthenticated non-consenting machine caller gets the consent error, not the auth error, and no OAuth token refresh happens pre-consent. Everything before the gate is local-only.
- **TTY check is stdin-only** (prompt reads stdin). `stdin=TTY, stdout=redirected` will prompt into the redirect — consistent with every other prompt in this CLI (`typer.confirm` defaults); the no-input-TTY hang case fails closed.
- **Breaking change by design**: previously-working unattended `comfy generate` invocations now require `--yes` — that is the ticket. Fallback for garbage `spend.auto_confirm` values is fail-closed.
- **Deny-path falsification**: the fail-closed error is a consent interlock, not a capability denial — every deny message names the working paths (`--yes`, `consent always`), and each of those paths is empirically proven by tests + a live CLI run.

## Tests

- New `tests/comfy_cli/command/generate/test_spend_gate.py` — 19 tests: fail-closed (json / pretty-non-TTY / async / pre-auth), bypasses (flag, config true/false/garbage), interactive prompt (accept/decline/default), ungated paths (emit-workflow, schema errors), lifecycle telemetry, and the `consent` action incl. on-disk persistence.
- New package `conftest.py` autouse fixture pre-authorizes the gate for the existing 177 generate tests (they exercise post-gate behavior; CliRunner has no TTY), overridden by name in the gate's own tests.
- Full suite: 2554 passed; the only failures are the 2 known `test_validate_command` reds from main (BE-3357×BE-3359, fix in #576). Ruff format clean; the 15 local `ruff check` UP038 hits are the known ruff-0.12.7-vs-CI-0.15.15 false flags in untouched files.

**Safety note (from the ticket): human review required before merge — this is the money interlock. Do not auto-merge.**

### DIFF
--- a/comfy_cli/command/generate/app.py
+++ b/comfy_cli/command/generate/app.py
@@ -2,19 +2,28 @@
 
 UX shape, modeled on fal-ai's genmedia but creative-user-first:
 
-    comfy generate <model> [--<param> value]... [--download P] [--async]
+    comfy generate <model> [--<param> value]... [--download P] [--async] [--yes]
     comfy generate list [--partner P] [--style S]
     comfy generate schema <model>
     comfy generate refresh
     comfy generate resume <model> <job_id> [--download P]
+    comfy generate consent [show|always|ask]
 
 The first positional is either a reserved action (``list``/``schema``/
-``refresh``/``resume``) or a model alias (``flux-pro``, ``ideogram-edit``, …).
-Anything not in the reserved set falls through to the generate path.
+``refresh``/``resume``/``consent``) or a model alias (``flux-pro``,
+``ideogram-edit``, …). Anything not in the reserved set falls through to the
+generate path.
+
+Spend gate: a generation call spends Comfy credits, so the proxy call sits
+behind a consent interlock (``_confirm_spend``) — an interactive TTY prompt,
+bypassed by ``--yes`` or the persisted ``spend.auto_confirm`` config, and
+fail-closed (error, no spend) when neither is present and no prompt is
+possible (``--json`` or no TTY).
 """
 
 from __future__ import annotations
 
+import sys
 import uuid
 from pathlib import Path
 from typing import Annotated
@@ -32,8 +41,9 @@ import typer
 from rich import print as rprint
 from rich.progress import Progress, SpinnerColumn, TextColumn, TimeElapsedColumn
 
-from comfy_cli import tracking, ui
+from comfy_cli import constants, tracking, ui
 from comfy_cli.command.generate import adapters, client, emit, output, poll, schema, spec, upload
+from comfy_cli.config_manager import ConfigManager
 from comfy_cli.output.renderer import get_renderer
 
 _HELP = "Generate images via ComfyUI partner nodes (Flux, Ideogram, DALL·E, Recraft, Stability, …)."
@@ -58,7 +68,7 @@ def register_with(parent: typer.Typer) -> None:
             str | None,
             typer.Argument(
                 help="A model alias (e.g. flux-pro, ideogram-edit, dalle) "
-                "or one of: list, schema, refresh, upload, resume.",
+                "or one of: list, schema, refresh, upload, resume, consent.",
             ),
         ] = None,
     ) -> None:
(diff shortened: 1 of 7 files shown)


bigcat88 (approved):
Reviewed as the money interlock deserves — code, tests, and live CLI runs (none of which spent anything). Approving:

- **Fail-closed verified live**: `comfy generate seedance --prompt x --json` with no consent → exit 1 with `spend_consent_required` — and crucially the *consent* error, not the auth error, proving the gate sits before `resolve_api_key` and any network side effect.
- **`--yes` verified live**: passes the gate and dies at the auth step on an unauthenticated box — gate passage without spend.
- **Persistence verified live**: `consent always` → `spend.auto_confirm = true` in config.ini → the same json call then reaches auth; `consent ask` reverts. `show` is a clean read.
- **Interactive default-No verified on a real TTY** (via `script`): prompt renders, bare Enter → "Canceled — no credits were spent."
- Fail-closed details hold up: `get_bool` returns `None` when unset (→ False) and raises `ValueError` on garbage (→ caught → False); closed-stdin `ValueError` from `isatty` → no prompt possible → fail closed.
- The autouse conftest fixture is safe: the suite-wide `_isolate_config_path` keeps the `cm.set("true")` writes in tmp dirs, and the teardown handles the singleton's in-memory state.
- Mutation check: neutralizing `_confirm_spend` fails 11 of the 19 spend-gate tests — the interlock is pinned from every direction.
- Full unit suite: 2562 passed. Two `test_generate_binary_response_*` failures I hit turned out to be my own environment: a fresh `~/.comfy/openapi-cache.yml` from live-testing #575 leaks into the test suite (`_USER_CACHE` has no test isolation — pre-existing, reproduces on main, disappears with the cache removed). Worth a small follow-up ticket: isolate the spec user-cache path in tests the way config/jobs-state already are.

Per the sweep protocol the merge decision stays with you — but this has had its human-directed empirical review.

PR #578, opened 2026-07-23
feat: comfy run-template — fetch, fill params, spend-gate, run to completion (BE-4131)

mattmillerai:
## ELI-5

`comfy templates` could only *browse* the gallery — you could list templates and fetch their JSON, but actually running one meant a manual fetch → edit → `comfy run` chain. This PR adds `comfy run-template <name>`: one command that fetches a gallery template, fills in any inputs you pass with `--param KEY=VALUE`, and runs it to completion on your local ComfyUI. If the template uses paid partner-API nodes (Flux Pro, Kling, Gemini, …), the command stops and asks for consent first — `--allow-spend` (or an interactive yes) is required before anything that burns Comfy credits is submitted.

## What

- New top-level verb **`comfy run-template <name>`** (`comfy_cli/command/templates.py`, registered in `cmdline.py`):
  - Resolves `<name>` against the gallery index (same close-matches affordance as `templates fetch`), pulls `templates/<name>.json`.
  - **`--param KEY=VALUE`** (repeatable) fills parameterized inputs via the existing CQL slot engine (`get_template_schema` / `apply_slots`). KEY is a slot address (`6.text`, `62/34.text`) or a unique slot *name* (`prompt`); VALUE is JSON-parsed with string fallback — identical semantics to `comfy workflow set-slot`. Unknown/ambiguous keys error with the candidate list so agents can self-correct.
  - **Spend gate**: before anything is submitted, the workflow's node types (including nodes inside UUID subgraph definitions) are checked against `object_info` (`api_node: true` / `partner/` category — same signals as `comfy run`'s partner detection), belt-and-suspendered with the gallery index's own paid markers (`API` tag, provider logos). Paid template + no `--allow-spend` → interactive confirm on a tty, hard error `spend_consent_required` (new registered error code) otherwise. Nothing is submitted, no credits spent.
  - Hands off to the **existing run path** (`command.run.execute`): UI→API conversion, partner-credential injection, preflight validation, execution, jobs state — nothing re-implemented. Runs to completion by default (`wait=True`); `--async` submits and returns. `--host/--port/--timeout/--verbose/--api-key/--json/--gallery/--refresh` mirror the surrounding commands.
- 14 new offline tests (`tests/comfy_cli/command/test_run_template.py`): resolution failures, param syntax/unknown-key/API-format rejection, the gate in all four postures (gallery-signal block, partner-node block without gallery signals, `--allow-spend` unblock, interactive decline), OSS pass-through, param filling by address and by name, wait/async handoff, temp-file cleanup.

## Judgment calls

- **BE-4103's shared spend-gate does not exist yet** (no branch, PR, or commit in this repo references it), so there was nothing to "route through." I implemented the gate inline in `run-template` with a registered error code and the `--allow-spend` consent flag — the same consent-before-credit contract the BE-4103 ticket describes for `comfy generate`. When BE-4103 lands a shared helper, this call site is a one-line adoption.
- **Gallery-signal calibration was checked against the live index** (528 templates): 0 of 224 OSS templates carry provider logos, so the provider signal cannot over-gate OSS runs; `openSource: false` alone is deliberately NOT a signal (60 local-model workflows like `sdxl_simple_example` carry it). Exactly one OSS template (`utility-gan_upscaler`) carries the `API` tag and would be conservatively gated — cost is one flag, spends nothing.
- **Missing-model behavior** is deferred to the existing run path: local model gaps surface through `comfy run`'s preflight/`/prompt` validation errors, which the local MCP already maps (`local-mcp-nomodel-001`); the per-template runnable/missing verdict is #557's scope (`comfy templates check`, in review). The command help points OSS users at `comfy model download`.
- The completion envelope comes from the run path and is labeled `command: "run"` — kept as-is rather than threading a label parameter through `execute()`'s stable contract.
- Residual under-gate window: an API-format paid template + unreachable `object_info` would skip node detection — but every current gallery template is frontend-format (verified), frontend conversion hard-fails without `object_info`, and credential injection would also be absent, so no spend can occur silently.

## Testing

- `pytest tests/comfy_cli` — 2550 passed; the only 2 failures (`test_validate_command.py::test_api_format_unchanged` / `test_empty_dict_payload_unchanged`) are pre-existing on `main` (fix is in-flight in #576).
- `comfy --json discover` verified to surface the new command + error code.

### DIFF
--- a/comfy_cli/command/templates.py
+++ b/comfy_cli/command/templates.py
@@ -751,7 +751,6 @@ def run_template_cmd(
     import tempfile
 
     from comfy_cli.command import run as run_module
-    from comfy_cli.config_manager import ConfigManager
     from comfy_cli.env_checker import check_comfy_server_running
 
     renderer = get_renderer()
@@ -819,20 +818,17 @@ def run_template_cmd(
         )
         raise typer.Exit(code=1) from e
 
-    # -- Resolve host/port exactly like `comfy run`'s local branch.
-    config = ConfigManager()
+    # -- Resolve host/port through the shared resolver, exactly like `comfy
+    # run`'s local branch (cmdline.py). This validates the host (rejecting
+    # URL-injection characters), brackets IPv6 literals, and honors
+    # config.background — behavior the old hand-rolled block lacked.
+    from comfy_cli.host_port import parse_host_port_arg, resolve_host_port
+
     if host:
-        s = host.split(":")
-        host = s[0]
-        if not port and len(s) == 2:
-            port = int(s[1])
-    if config.background:
-        host = host or config.background[0]
-        port = port or config.background[1]
-    host = host or "127.0.0.1"
-    if host == "0.0.0.0":  # wildcard bind, not a connect address
-        host = "127.0.0.1"
-    port = port or 8188
+        host, parsed_port = parse_host_port_arg(host)
+        if not port and parsed_port is not None:
+            port = parsed_port
+    host, port = resolve_host_port(host, port)
 
     if not check_comfy_server_running(port, host, timeout=timeout):
         renderer.error(
--- a/comfy_cli/command/templates.py
+++ b/comfy_cli/command/templates.py
@@ -524,3 +524,425 @@ def fetch_cmd(
     if renderer.is_pretty() and out:
         rprint(f"[green]✓[/green] wrote {len(body):,} bytes ({payload['node_count']} nodes) to {target_repr}")
     renderer.emit(payload, command="templates fetch")
+
+
+# ---------------------------------------------------------------------------
+# run-template — fetch → fill params → spend-gate → run via the run path
+# ---------------------------------------------------------------------------
+
+
+def _parse_param_value(raw: str) -> Any:
+    """Parse a ``--param`` value as JSON; fall back to the literal string.
+
+    Mirrors ``comfy workflow set-slot`` semantics so `--param seed=42` writes
+    an int and `--param prompt="a cat"` writes a string.
+    """
+    try:
+        return json.loads(raw)
+    except (json.JSONDecodeError, ValueError):
+        return raw
+
+
+def _workflow_node_types(workflow: Any) -> set[str]:
+    """Collect node class names from a workflow in either format.
+
+    Frontend format: ``nodes[].type`` plus every ``definitions.subgraphs[].nodes[].type``
+    (gallery templates routinely hide partner nodes inside UUID subgraphs).
+    API format: ``values()[].class_type``.
+    """
+    types: set[str] = set()
+    if not isinstance(workflow, dict):
+        return types
+    if isinstance(workflow.get("nodes"), list):
+        node_lists = [workflow.get("nodes") or []]
+        subgraphs = (workflow.get("definitions") or {}).get("subgraphs") or []
+        for sg in subgraphs:
+            if isinstance(sg, dict):
+                node_lists.append(sg.get("nodes") or [])
+        for nodes in node_lists:
+            for node in nodes:
+                if isinstance(node, dict) and isinstance(node.get("type"), str):
+                    types.add(node["type"])
+        return types
+    for node in workflow.values():
+        if isinstance(node, dict) and isinstance(node.get("class_type"), str):
+            types.add(node["class_type"])
+    return types
+
+
+def _detect_paid_nodes(workflow: Any, object_info: dict) -> list[str]:
+    """Sorted node class names in ``workflow`` that are partner-API (paid) nodes.
+
+    Same signals as ``comfy_cli.command.run``'s partner detection — the
+    authoritative ``api_node: true`` flag with a ``partner/`` category-prefix
+    fallback — but format-agnostic so it works on the frontend-format JSON
+    gallery templates ship as (run's detector only reads API format).
+    """
+    from comfy_cli.command.run import PARTNER_NODE_CATEGORY_PREFIXES
+
+    out: list[str] = []
+    for ct in _workflow_node_types(workflow):
+        info = object_info.get(ct) or {}
+        if not isinstance(info, dict):
+            continue
+        if info.get("api_node") is True:
+            out.append(ct)
+            continue
+        category = info.get("category")
+        if isinstance(category, str) and category.startswith(PARTNER_NODE_CATEGORY_PREFIXES):
+            out.append(ct)
+    return sorted(out)
+
+
+def _gallery_paid_signals(row: dict[str, Any]) -> list[str]:
+    """Gallery-index evidence that a template runs partner/API nodes.
+
+    Belt-and-suspenders for the spend gate: when the local server's
+    object_info is unavailable (fail-open fetch) node detection can miss, but
+    the curated gallery still marks paid templates with the ``API`` tag and
+    provider logos. Returns human-readable signal strings, empty = no signal.
+    """
+    signals: list[str] = []
+    for tag in row.get("tags") or []:
+        if isinstance(tag, str) and tag.lower() == "api":
+            signals.append("tag:API")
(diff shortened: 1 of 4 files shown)


comfy_cli/error_codes.py
```diff
@@ -149,6 +149,14 @@ class ErrorCode:
         "re-submit with `--where cloud` (the CLI auto-injects the credential there), or run "
         "`comfy auth set comfy-cloud-api-key --key …` so the local submit path can inject it too",
     ),
+    ErrorCode(
+        "spend_consent_required",
```
bigcat88: Collision with #577, which registers this same code name for the `comfy generate` gate. I merged the two branches locally: git auto-merges cleanly (the entries land in different REGISTRY regions), and the combined tree fails `tests/comfy_cli/output/test_error_code_registry.py::test_no_duplicate_codes` — so whichever PR merges second breaks main without a textual conflict to warn anyone. Please coordinate: land one entry with a description covering both surfaces (this one is currently `run-template`-specific), and have the second PR rebase to reuse it.

comfy_cli/command/templates.py
```diff
@@ -524,3 +524,425 @@ def fetch_cmd(
...
+        workflow = json.loads(body)
+    except json.JSONDecodeError as e:
+        renderer.error(
+            code="template_workflow_invalid_json",
+            message=f"upstream returned non-JSON for {name!r}: {e}",
+            hint="report at https://github.com/Comfy-Org/workflow_templates/issues",
+        )
+        raise typer.Exit(code=1) from e
+
+    # -- Resolve host/port exactly like `comfy run`'s local branch.
```
bigcat88: This comment is out of date — `comfy run` doesn't do this inline anymore; it goes through `parse_host_port_arg` + `host_port.resolve_host_port` (`cmdline.py:877-884`), which also validates the host and brackets IPv6 literals — this hand-rolled block does neither. More importantly, #571 (approved) routes `COMFY_LOCAL_URL` through `resolve_host_port`, so as written `run-template` would be the one local command that ignores the env var — on exactly the MCP boxes with non-default ports that motivated both features. Swapping this block for the shared resolver fixes the divergence and picks up #571 for free when it lands.

bigcat88 (changes requested):
I put this through a full live cycle against a real local ComfyUI and the feature itself is in great shape — the two blockers are coordination with your own in-flight PRs, both minutes to fix.

**What I verified live** (server on `:8189`, shared model library mounted):

- Full happy path: `comfy --json run-template sdxl_simple_example --port 8189` with **7 `--param` fills** (both `ckpt_name`s by address to a Lightning checkpoint, steps, prompt) → gate correctly silent for OSS → UI→API conversion → ran to completion → `ok: true` with the output PNG on disk. The one-command promise is real.
- Spend gate on `api_seedance2_0_r2v` without `--allow-spend` → exit 1, `spend_consent_required`, with **both** evidence channels populated: `partner_nodes: ["ByteDance2ReferenceNode"]` and `gallery_signals: ["tag:API", "provider:ByteDance"]` — so the `providers` field really is in the live index and the belt-and-suspenders design works.
- Unknown `--param` key → candidate address list; missing-model OSS run → the run path's per-node enum validation with suggestions. Both are exactly the self-correction surface an agent needs.
- Full unit suite on this branch merged with current main: 2557 passed.

**The two blockers** are inline: the `spend_consent_required` collision with #577 (empirically: the two branches auto-merge cleanly and the combined tree fails `test_no_duplicate_codes` — whichever lands second breaks main), and the hand-rolled host/port block that diverges from `comfy run`'s actual resolver and will leave `run-template` deaf to #571's `COMFY_LOCAL_URL` — on the very MCP boxes this command targets (BE-3956 × BE-4131).

One non-blocking note: the PR body's "BE-4103's shared spend-gate does not exist yet" is now stale — #577 exists and is approved. When you adopt its helper (or keep the inline gate), consider also aligning the consent flags (`--allow-spend` here vs `--yes` + `spend.auto_confirm` there); not honoring the persisted consent here is conservative and fine, but the naming split is UX debt worth a follow-up ticket.

comfy_cli/command/templates.py
mattmillerai: Fixed in b0dd039. Replaced the hand-rolled block with the shared `parse_host_port_arg` + `resolve_host_port` (host_port.py), exactly like `comfy run`'s local branch in cmdline.py — so run-template now validates the host, brackets IPv6 literals, and stays aligned with #571's `COMFY_LOCAL_URL` routing when it lands. Updated the stale comment too. (Note: this also drops the old `0.0.0.0`→`127.0.0.1` remap, matching `comfy run`, which doesn't remap either.)

comfy_cli/error_codes.py
mattmillerai: Good catch on the silent collision. I've broadened this entry's description to cover both surfaces (`comfy run-template` + `comfy generate`) in b0dd039 so the canonical text is ready.

The full dedup can't land on #578 alone right now: `test_error_code_registry` is bidirectional (every emitted code must be registered *and* every registered code must be emitted), so #578 can't drop its entry while its run-template source still emits the code — that would break #578's own CI before #577 is on main. Since #577 is APPROVED + MERGEABLE it'll merge first; the plan is then to rebase #578 onto main, delete this duplicate entry, and keep the broadened both-surfaces description on the surviving one. I've recorded that as a tracked follow-up so it isn't lost. Leaving this thread open until the reconciliation actually lands.

mattmillerai:
- [BE-4190](https://linear.app/comfyorg/issue/BE-4190) — Reconcile duplicate spend_consent_required error code after #577 merges (BE-4131/BE-4103)

mattmillerai:
- [BE-4193](https://linear.app/comfyorg/issue/BE-4193) — Align spend-consent UX flags across `comfy run-template` and `comfy generate`

mattmillerai:
@bigcat88 both blockers from your live cycle are addressed in b0dd039 — re-requesting review:

1. **Host/port** — the hand-rolled block is gone; `run-template` now goes through `parse_host_port_arg` + `resolve_host_port`, byte-for-byte the same path as `comfy run`'s local branch (cmdline.py:877-884). That gets host validation, IPv6 bracketing, and — the thing you flagged — it inherits #571's `COMFY_LOCAL_URL` support the moment #571 lands, on exactly the MCP boxes this targets. Thread resolved.

2. **`spend_consent_required` collision with #577** — the entry's description is broadened to cover both `run-template` and `generate` so the canonical text is ready. The full dedup can't land on #578 alone: `test_no_duplicate_codes` is bidirectional, so #578 can't drop its entry while its source still emits the code — that'd break #578's own CI before #577 is on main. Plan stands: once one of the two merges, the second rebases onto main and drops the duplicate entry (keeping the emit site). Left that thread open as the coordination tracker.

Non-blocking consent-flag naming note (`--allow-spend` vs #577's `--yes`/`spend.auto_confirm`) recorded as a follow-up for after both land. Full unit suite + registry/templates tests green locally, CI green.

comfy_cli/error_codes.py
mattmillerai: Reconciled now that #577 is on main (bc2b874). Merged current `origin/main` into this branch, which surfaced the duplicate (clean git auto-merge, no textual conflict — exactly as you flagged), then dropped this branch's entry and kept #577's canonical one, broadening its description to cover both surfaces (`comfy run-template` + `comfy generate`). So there's a single registration now.

Verified: `test_error_code_registry.py` green including `test_no_duplicate_codes` (6 passed), plus `test_templates` + `test_spend_gate` (30). Full suite 2770 passed / 36 skipped against current main.

Left the `--allow-spend` vs `--yes`/`spend.auto_confirm` naming split as-is per your non-blocking note; happy to file a follow-up for the UX alignment.

bigcat88 (approved):

jojodecayz (approved):

PR #583, opened 2026-07-24
fix(generate): skip deprecated partner endpoints in the openapi registry (BE-4298)

jojodecayz:
## Problem

`comfy_cli/command/generate/spec.py`'s `_registry()` built an `Endpoint` for every allowlisted `/proxy` op with **no deprecation check**, so a retired endpoint would surface through `comfy generate` (list + run) if it lingered in the allowlist or was deprecated upstream after a `generate refresh`. comfy-api flags deprecation in the operation **`summary`**, not the machine `deprecated` flag.

Mirrors the Cloud MCP partner-catalog generator's deprecation filter for cross-surface parity (Comfy-Org/comfy-cloud-mcp-server#805).

## Change

- `_is_deprecated_op(op)`: `op.deprecated is True` **OR** the **summary** matches `/deprecated/i`. Summary-only (not description) is deliberate — the live `stability/.../generate/sd3` endpoint is current (`"Stable Diffusion 3.5"`) yet its description mentions the older SD 3.0 API being deprecated; a description-wide scan would wrongly drop it.
- Skip deprecated ops in `_registry()`, same silent-skip contract as the existing missing-node case.

## Tests

- `_is_deprecated_op` cases incl. the **sd3 false-positive guard**.
- A registry test proving a deprecated allowlisted endpoint is dropped while its live sibling on the same partner survives.

All 18 tests in `test_spec.py` pass; `ruff check` + `ruff format` clean.

Part of BE-4298. Companion Cloud MCP PR: Comfy-Org/comfy-cloud-mcp-server#805.


jojodecayz:
Logic and tests look right — faithful port of the MCP-side rule, same `sd3` guard, and the registry-level test (deprecated endpoint dropped, live sibling on the same partner survives) is actually stronger than its counterpart in comfy-cloud-mcp-server#805.

One inconsistency worth closing: this drops silently.

```python
if _is_deprecated_op(op):
    # Never surface a deprecated endpoint through `comfy generate`.
    # Same silent-skip contract as the missing-node case above...
    continue
```

BE-4298 specifies *"Skip deprecated ops from the catalog/registry, **recorded visibly (never silent)**"*, and the MCP half honours that by pushing each drop into `skipped[]` as a stderr diagnostic. Since the entire ticket is about endpoints disappearing from a surface without anyone noticing, the CLI is the wrong half to stay quiet on — if a live endpoint is ever mis-flagged, this is the code path where nobody would find out.

Suggest matching #805: emit a diagnostic naming the skipped endpoint and the reason.

mattmillerai (changes requested):
@jojodecayz If you can accept the CLA (see link in comment above), I can approve this. Thanks!

jojodecayz:

PR #587, opened 2026-07-24
docs: document registry-install & --fast-deps, fix node-install example

### LINKED ISSUE #257: Add docs for registry-install and --fast-deps
1. The commands provided in the ComfyUI Registry use `comfy node registry-install`. This command is not documented and it's not clear to me how this command is different from `comfy node install`.

2. The example in the Readme: `comfy node install ComfyUI-Impact-Pack` doesn't work:
```
ERROR: An error occurred while installing 'ComfyUI-Impact-Pack'.
Node 'ComfyUI-Impact-Pack@unknown' not found in
```

3. There is no documentation for `--fast-deps`. And it's not clear to me why `registry-install` doesn't work with `--fast-deps`.

mattmillerai:
## ELI-5

If you told comfy-cli to install a node using its GitHub-style name
(`ComfyUI-Impact-Pack`), it failed with a confusing `Node
'ComfyUI-Impact-Pack@unknown' not found` error, because the command actually
wants the node's lowercase **Comfy Registry ID** (`comfyui-impact-pack`). Also,
comfy-cli has two different install commands (`install` and `registry-install`)
and a `--fast-deps` speed flag that were never explained. This PR is docs-only:
it explains all of that in the README so nobody hits the same wall.

## What changed

Documentation only (`README.md`) — no code changes.

1. **Node-install casing note** — added a note under the `comfy node install`
   example clarifying the argument is the node's lowercase Comfy Registry ID,
   not the GitHub repo name, and that the wrong casing produces the
   `Node '...@unknown' not found` error from the issue.
2. **`install` vs `registry-install`** — new section explaining that `install`
   delegates to ComfyUI-Manager's `cm-cli` (multi-node, channel/mode-aware,
   supports the dependency flags, needs the Manager) while `registry-install`
   downloads a single published node straight from the Comfy Registry API and
   runs its bundled install script (no Manager required — the command the
   Registry's own instructions use).
3. **`--fast-deps`** — new section documenting what it does, the commands that
   accept it (`comfy install`, `comfy node install`, `comfy node reinstall`),
   and why `registry-install` does not (it bypasses comfy-cli's dependency
   resolver entirely).

## Acceptance criteria

- [x] README node-install example corrected to a working command — the literal
  example was already lowercase on `main` (fixed in `0cb2cb8`, Mar 2026, after
  the issue was filed); this PR adds the missing *explanation* of why the
  capitalized form fails, which is the actual reporter confusion.
- [x] `registry-install` vs `install` documented (what each does, when to use
  which), verified against `comfy_cli/command/custom_nodes/command.py`.
- [x] `--fast-deps` documented, including which node commands accept it and the
  current reason `registry-install` does not, verified against the command
  signatures in source.
- [x] Follow-up bug: **not needed.** The `ComfyUI-Impact-Pack` failure is a
  case-sensitivity / stale-doc issue (node IDs are the lowercase registry IDs),
  not an install-resolution bug — the working form already ships on `main`. No
  resolver change is in scope (per the ticket's out-of-scope note).

## Judgment calls / notes

- `registry-install` is defined with `hidden=True` in the CLI. I documented it
  anyway (the Registry's install instructions use it and the issue explicitly
  asks for its docs) but left it hidden — unhiding is a behavior change and
  out of scope for a docs ticket.
- The `--fast-deps` "not accepted by registry-install" claim was verified by
  reading `registry_install`'s signature (it only exposes `--force-download`),
  and the docs redirect users to `comfy node install` for that capability
  rather than presenting a dead-end.
- The docs.comfy.org custom-nodes pages that surface these commands live in a
  **separate docs repo**, not this repo — those are out of scope for this
  comfy-cli README PR and would need a follow-up there if the same guidance is
  wanted on the docs site.

Closes #257

### DIFF
--- a/README.md
+++ b/README.md
@@ -226,6 +226,13 @@ comfy node [show|simple-show] [installed|enabled|not-installed|disabled|all|snap
 
   `comfy node install comfyui-impact-pack`
 
+  > **Note:** the argument is the node's **Comfy Registry ID**, which is
+  > lowercase (e.g. `comfyui-impact-pack`), not the GitHub repository name
+  > (`ComfyUI-Impact-Pack`). Passing the repo-name casing fails to resolve with
+  > an error like `Node 'ComfyUI-Impact-Pack@unknown' not found`. Find a node's
+  > ID on its [Comfy Registry](https://registry.comfy.org) page or via
+  > `comfy node show all`.
+
 - Managing snapshot:
 
   `comfy node save-snapshot`
@@ -241,6 +248,39 @@ comfy node [show|simple-show] [installed|enabled|not-installed|disabled|all|snap
 - Generate deps:
 
   `comfy node deps-in-workflow --workflow=<workflow .json/.png file> --output=<output deps .json file>`
+
+#### `install` vs `registry-install`
+
+comfy-cli offers two ways to install a custom node, backed by different
+mechanisms:
+
+- **`comfy node install <id>...`** delegates to
+  [ComfyUI-Manager](https://github.com/Comfy-Org/ComfyUI-Manager)'s `cm-cli`.
+  It accepts one or more node IDs, resolves them through the Manager's channel
+  database (`--channel`, `--mode`), and installs dependencies via the Manager
+  (so it also supports `--fast-deps`, `--no-deps`, and `--uv-compile`). This is
+  the recommended command for day-to-day node management, and it requires
+  ComfyUI-Manager to be present in the workspace.
+
+  `comfy node install comfyui-impact-pack`
+
+- **`comfy node registry-install <id> [version]`** talks directly to the
+  [Comfy Registry](https://registry.comfy.org) API. It downloads the published
+  archive for a single node (optionally pinned to a specific `version`),
+  extracts it into `custom_nodes/`, and runs the node's own install script. It
+  does **not** go through ComfyUI-Manager, so it does not require the Manager to
+  be installed — this is the command the Registry's own install instructions
+  use.
+
+  `comfy node registry-install comfyui-impact-pack`
+
+  `comfy node registry-install comfyui-impact-pack 1.0.0`  # install a specific version
+
+  Because `registry-install` bypasses the Manager's dependency machinery and
+  simply runs the node's bundled install script, it only accepts
+  `--force-download` — it does **not** accept `--fast-deps`, `--no-deps`, or
+  `--uv-compile`. If you want fast/unified dependency resolution, use
+  `comfy node install` instead.
 
 #### Unified Dependency Resolution (--uv-compile)
 
@@ -268,6 +308,26 @@ it can identify which node packs have incompatible dependencies and why.
 - Use `--no-uv-compile` to override the default for a single command:
 
   `comfy node install comfyui-impact-pack --no-uv-compile`
+
+#### --fast-deps
+
+`--fast-deps` swaps comfy-cli's dependency installation from `pip` to comfy-cli's
+built-in `uv`-based resolver (`DependencyCompiler`), which is significantly
+faster and only requires `uv` (no ComfyUI-Manager). On a dependency version
+conflict it prompts you interactively to pick a version.
+
+- Accepted by: `comfy install`, `comfy node install`, and
+  `comfy node reinstall`.
+
+  `comfy install --fast-deps`
+
+  `comfy node install comfyui-impact-pack --fast-deps`
+
+- **Not** accepted by `comfy node registry-install`: that command installs a
+  node directly from the Comfy Registry by running the node's own bundled
+  install script, so it never touches comfy-cli's dependency resolver (see
+  [`install` vs `registry-install`](#install-vs-registry-install) above).
+- Mutually exclusive with `--no-deps` and `--uv-compile`.
 
 #### --fast-deps vs --uv-compile
 
--- a/README.md
+++ b/README.md
@@ -293,17 +293,16 @@ mechanisms:
 
   `comfy node install comfyui-impact-pack`
 
-- **`comfy node registry-install <id> [version]`** talks directly to the
+- **`comfy node registry-install <id> [--version <v>]`** talks directly to the
   [Comfy Registry](https://registry.comfy.org) API. It downloads the published
-  archive for a single node (optionally pinned to a specific `version`),
-  extracts it into `custom_nodes/`, and runs the node's own install script. It
-  does **not** go through ComfyUI-Manager, so it does not require the Manager to
-  be installed — this is the command the Registry's own install instructions
-  use.
+  archive for a single node (optionally pinned with `--version`), extracts it
+  into `custom_nodes/`, and runs the node's own install script. It does **not**
+  go through ComfyUI-Manager, so it does not require the Manager to be
+  installed — this is the command the Registry's own install instructions use.
 
   `comfy node registry-install comfyui-impact-pack`
 
-  `comfy node registry-install comfyui-impact-pack 1.0.0`  # install a specific version
+  `comfy node registry-install comfyui-impact-pack --version 1.0.0`  # install a specific version
 
   Because `registry-install` bypasses the Manager's dependency machinery and
   simply runs the node's bundled install script, it only accepts
@@ -329,7 +328,9 @@ it can identify which node packs have incompatible dependencies and why.
 
   `comfy node uv-sync`
 
-- `--uv-compile` is mutually exclusive with `--fast-deps` and `--no-deps`.
+- `--uv-compile` is mutually exclusive with `--fast-deps` and `--no-deps` —
+  except on `restore-snapshot`, where `--fast-deps` selects the `--uv-compile`
+  path instead (see [--fast-deps](#--fast-deps) below).
 
 - To make `--uv-compile` the default for all commands, see
   [uv-compile default](#uv-compile-default) below.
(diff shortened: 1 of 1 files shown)


bigcat88 (changes requested):
Requesting changes — this is a docs PR whose whole point is to stop people hitting confusing walls, and one of its examples is itself a command that fails.

## Blocking: the `registry-install` version example doesn't work

The README documents a positional version, twice — in the section signature `comfy node registry-install <id> [version]` and in the example:

```
comfy node registry-install comfyui-impact-pack 1.0.0  # install a specific version
```

That errors. `version` is an *option*, not a positional:

```python
def registry_install(node_id: str, version: str | None = None, force_download: ... )
```

typer renders a defaulted `str | None` as `--version`, and the usage line agrees — `comfy node registry-install [OPTIONS] NODE_ID`, one positional only. Live:

```
$ comfy node registry-install comfyui-impact-pack 1.0.0
Error: Got unexpected extra argument (1.0.0)          # exit 2

$ comfy node registry-install comfyui-impact-pack --version 1.0.0
(parses; reaches the registry)                        # exit 0
```

Please use `--version 1.0.0` in both places.

## Blocking-ish: the `--fast-deps` list goes stale the moment #589 lands

The new section says `--fast-deps` is "Accepted by: `comfy install`, `comfy node install`, and `comfy node reinstall`" and is "Mutually exclusive with `--no-deps` and `--uv-compile`."

Both are accurate against `main` today — I checked every subcommand's `--help`:

```
main:        comfy node restore-snapshot : does NOT accept --fast-deps   ✓ matches README
#589 branch: comfy node restore-snapshot : ACCEPTS --fast-deps
```

#589 adds `--fast-deps` to `restore-snapshot`, and there it *implies* `--uv-compile` rather than excluding it. Since I'm merging #589, please add `comfy node restore-snapshot` to the accepted-by list and note that on that subcommand `--fast-deps` selects cm_cli's `--uv-compile` fast path (the two are synonyms there, not exclusive) — otherwise the README contradicts the shipped CLI on both points.

## What checks out

- **The casing fix is right**, which is the valuable part. Against the Registry API: `GET /nodes/comfyui-impact-pack` → **200**, `GET /nodes/ComfyUI-Impact-Pack` → **302**. Lowercase is canonical, so the guidance and the pointer to `comfy node show all` are sound. (I couldn't reproduce the exact `Node '...@unknown' not found` string here since it comes from cm-cli and ComfyUI-Manager isn't installed in this workspace — but it's from the linked issue and the underlying claim holds.)
- **The `install` vs `registry-install` split is accurate.** `registry-install --help` exposes exactly `--force-download`, `--version`, `--help` — no `--fast-deps`, `--no-deps`, or `--uv-compile`, exactly as documented.

## Unrelated bug I hit while checking (not yours — worth a ticket)

`comfy node registry-install` **exits 0 when it fails**:

```
$ comfy node registry-install comfyui-impact-pack --version 1.0.0
ERROR - Failed to install node: 404 - {"error":"","message":"Not found"}
Failed to download the custom node comfyui-impact-pack.
$ echo $?
0
```

A failed install reporting success is a real trap for scripts and agents. Pre-existing and out of scope here, but since this PR is about making node installation less confusing, it's a natural follow-up.

README.md
```diff
@@ -242,6 +249,39 @@ comfy node [show|simple-show] [installed|enabled|not-installed|disabled|all|snap
...
+  [Comfy Registry](https://registry.comfy.org) API. It downloads the published
+  archive for a single node (optionally pinned to a specific `version`),
+  extracts it into `custom_nodes/`, and runs the node's own install script. It
+  does **not** go through ComfyUI-Manager, so it does not require the Manager to
+  be installed — this is the command the Registry's own install instructions
+  use.
+
+  `comfy node registry-install comfyui-impact-pack`
+
+  `comfy node registry-install comfyui-impact-pack 1.0.0`  # install a specific version
```
bigcat88: This example fails. `version` is an option, not a positional:

```
$ comfy node registry-install comfyui-impact-pack 1.0.0
Usage: comfy node registry-install [OPTIONS] NODE_ID
Error: Got unexpected extra argument (1.0.0)     # exit 2
```

`registry_install(node_id: str, version: str | None = None, ...)` — typer renders a defaulted `str | None` as `--version`, and the usage line shows only one positional. Please change to:

```
comfy node registry-install comfyui-impact-pack --version 1.0.0  # install a specific version
```

The section signature on line 267 needs the same fix (`<id> [version]` → `<id> [--version <v>]`).

README.md
```diff
@@ -269,6 +309,26 @@ it can identify which node packs have incompatible dependencies and why.
...
   `comfy node install comfyui-impact-pack --no-uv-compile`
 
+#### --fast-deps
+
+`--fast-deps` swaps comfy-cli's dependency installation from `pip` to comfy-cli's
+built-in `uv`-based resolver (`DependencyCompiler`), which is significantly
+faster and only requires `uv` (no ComfyUI-Manager). On a dependency version
+conflict it prompts you interactively to pick a version.
+
+- Accepted by: `comfy install`, `comfy node install`, and
```
bigcat88: This list is correct against `main` today but goes stale as soon as #589 lands — I checked both trees with `--help`:

```
main:        comfy node restore-snapshot : does NOT accept --fast-deps
#589 branch: comfy node restore-snapshot : ACCEPTS --fast-deps
```

#589 is merging, so please add `comfy node restore-snapshot` here.

It's also an exception to the "Mutually exclusive with `--no-deps` and `--uv-compile`" bullet just below: on `restore-snapshot`, `--fast-deps` **selects** cm_cli's `--uv-compile` fast path. Upstream cm_cli (v4.2.2) has no `--no-deps` on that subcommand, and its `--uv-compile` calls `set_no_deps(True)` internally — so there the two are synonyms rather than exclusive, and `--fast-deps --no-uv-compile` is the pair that errors.

README.md
mattmillerai: Confirmed and fixed in f63d8aa — thank you, you're exactly right.

Reproduced your exit-2 verbatim against the branch:

```
$ comfy node registry-install comfyui-impact-pack 1.0.0
Usage: comfy node registry-install [OPTIONS] NODE_ID
Error: Got unexpected extra argument (1.0.0)
```

And `--help` confirms the shape — `node_id` is the only positional, `version` renders as an option:

```
Usage: comfy node registry-install [OPTIONS] NODE_ID
╭─ Arguments ─╮ *  node_id  TEXT  [required]
╭─ Options   ─╮ --version TEXT   |  --force-download
```

Both spots are corrected: the section signature is now `comfy node registry-install <id> [--version <v>]` and the example is `comfy node registry-install comfyui-impact-pack --version 1.0.0`. I also changed the prose "optionally pinned to a specific `version`" → "optionally pinned with `--version`" so the flag form is stated three-for-three.

README.md
mattmillerai: Good catch, and your timing was right — #589 merged this morning (68311e1), so this is now stale against `main` rather than merely about-to-be. I merged `origin/main` into the branch and fixed it in f63d8aa.

Verified on the merged tree via `--help` and by actually invoking the rejected pairs, rather than reading the code:

| command | `--fast-deps` | `--no-deps` | `--uv-compile` | rejected pair (actual stderr) |
|---|---|---|---|---|
| `comfy install` | yes | — | — | none; both are `No such option` |
| `comfy node install` | yes | yes | yes | `Cannot use --fast-deps and --no-deps together` / `...and --uv-compile together` |
| `comfy node reinstall` | yes | — | yes | `Cannot use --fast-deps and --uv-compile together` |
| `comfy node restore-snapshot` | yes | — | yes | `Cannot use --fast-deps with --no-uv-compile` |
| `comfy node registry-install` | — | — | — | all three are `No such option` |

Changes made:

- Added `comfy node restore-snapshot` to the **Accepted by** list, and to the `Scope` row of the `--fast-deps` vs `--uv-compile` table (same staleness, one section down).
- Replaced the blanket `Mutually exclusive with --no-deps and --uv-compile` bullet with a per-command table, since as you note that sentence was never true uniformly — `reinstall` and `comfy install` don't even have `--no-deps`.
- Wrote up your `restore-snapshot` exception explicitly: `--fast-deps` **selects** the uv-compile fast path there (cm_cli has no `--no-deps` on that subcommand and its `--uv-compile` calls `set_no_deps(True)` internally, so they're synonyms), and `--fast-deps --no-uv-compile` is the pair that errors.
- Also annotated the pre-existing `--uv-compile is mutually exclusive with --fast-deps and --no-deps` bullet in the section above with a pointer to that exception, since it had the same over-broad claim.

bigcat88 (approved):
Both blockers fixed — verified against the real CLI, not just the diff.

**1. The `registry-install` version example now works.** Corrected in both places (`<id> [--version <v>]` in the signature, `--version 1.0.0` in the example):

```
$ comfy node registry-install comfyui-impact-pack --version 1.0.0   -> exit 0 (parses, reaches the registry)
$ comfy node registry-install comfyui-impact-pack 1.0.0             -> Got unexpected extra argument (1.0.0)
```

The second line is why this mattered — the old text documented a form that errors.

**2. The `--fast-deps` list is now accurate post-#589**, and better than I asked for. You added `comfy node restore-snapshot` to the accepted-by list *and* documented the exception rather than leaving the "mutually exclusive" bullet to contradict it:

> except on `restore-snapshot`, where `--fast-deps` selects the `--uv-compile` …

Confirmed live: `comfy node restore-snapshot --help` now advertises `--fast-deps`, and the pair your table flags as invalid does error —

```
$ comfy node restore-snapshot … --fast-deps --no-uv-compile
Cannot use --fast-deps with --no-uv-compile
```

Still README-only, CI green. The casing guidance that motivated the PR remains correct (`GET /nodes/comfyui-impact-pack` → 200, `GET /nodes/ComfyUI-Impact-Pack` → 302).

The `registry-install` exit-0-on-failure bug I mentioned is still there (`Failed to download the custom node comfyui-impact-pack.` with `$? == 0`) — out of scope here, still worth a ticket.

Issue #217, opened 2024-12-04
Add "--fast-deps" (using uv) flag to restore snapshot functionality


anton-averich:
Now it's possible to run the "**comfy install**" command with the "**--fast-deps**" flag that significantly saves time on the installation process.

But restoring a snapshot process, especially with many nodes, still takes a lot of time, and adding this functionality can help to save years for users around the world.

It would be great to have the possibility to use it something like this:
`comfy node restore-snapshot snapshot.json --fast-deps`

polarathene:
@mattmillerai why was this issue not closed as stale? It was acknowledged by you when closing https://github.com/Comfy-Org/comfy-cli/pull/15#issuecomment-5064491174 (_during the [bulk closing of issues](https://github.com/Comfy-Org/comfy-cli/issues/206#issuecomment-5065095386)_)

**EDIT:** I've gone over the recent PR activity and see that you are indeed using AI tooling (_Claude agents + cursor_), so that explains the large amount of activity. More disclosure when delegating to AI tools to close issues would be appreciated. I'm still not sure why this issue did not receive the same treatment you (or an agent) decided to dish out to other issues.

PR #589, opened 2026-07-24
feat(node): support --fast-deps on restore-snapshot (BE-4276)

mattmillerai:
## ELI-5

`comfy install --fast-deps` uses the fast **uv** installer, but restoring a
snapshot with many custom nodes had no equivalent — so it stayed slow. This adds
a `--fast-deps` flag to `comfy node restore-snapshot` so the restore also uses
the fast uv dependency install. Requested in issue #217.

```
comfy node restore-snapshot snapshot.json --fast-deps
```

## What changed

- New `--fast-deps` option on `comfy node restore-snapshot`.
- When set, the restore forwards the **uv-compile fast path** to cm_cli.
- Default (flag omitted) behavior is **byte-for-byte unchanged**.
- Contradictory `--fast-deps --no-uv-compile` is rejected with a clear error.

## Why it maps to `--uv-compile` (the interesting bit / AC #3)

comfy-cli has two uv-based fast paths:
1. **`--fast-deps`** (install/reinstall): pass `--no-deps` to cm_cli, then run
   comfy-cli's own `DependencyCompiler`.
2. **`--uv-compile`**: pass `--uv-compile` to cm_cli, which batch-resolves with uv.

I verified against ComfyUI-Manager's actual cm_cli v4 source
(`cm_cli/__main__.py`, `manager-v4` branch): its `restore-snapshot` subcommand
accepts `--uv-compile` but **does not accept `--no-deps`** (only
`install`/`reinstall`/`fix`/`update` accept `--no-deps`). So path (1) is
**genuinely unavailable upstream** for snapshot restore — passing `--no-deps`
would make cm_cli error out.

The fast uv install that `restore-snapshot` *does* support is `--uv-compile`,
which internally does `cmd_ctx.set_no_deps(True)` + a batch uv resolve — i.e. it
already *is* the fast-deps equivalent. So `--fast-deps` forwards that path rather
than being silently ignored. On a ComfyUI-Manager older than v4.1 (no
`--uv-compile` on restore-snapshot), cm_cli surfaces its own "no such option"
error — same contract as the pre-existing `--uv-compile` flag.

## Acceptance criteria

- [x] `restore-snapshot <file> --fast-deps` accepts the flag and forwards the
  uv/fast path.
- [x] Restore uses the fast (uv) dependency install when `--fast-deps` is set;
  default behavior unchanged when omitted.
- [x] cm_cli honors it (v4.1+); never silently ignored, and the upstream
  limitation (no `--no-deps` on restore-snapshot) is documented above.

## Judgment calls

- **`restore-dependencies`** has the identical uv plumbing and could gain the
  same alias, but issue #217 and the ticket scope this to `restore-snapshot`
  only — left as an easy follow-up rather than expanding scope.
- `--fast-deps` maps to the `--uv-compile` mechanism here (not the
  `DependencyCompiler` mechanism it uses for `install`), because that is the only
  fast uv path cm_cli exposes for snapshot restore. The help text says so.

## Testing

- New `tests/comfy_cli/command/nodes/test_restore_snapshot_fast_deps.py` (5 tests):
  flag forwards `uv_compile=True`, default omitted stays `False`, `--uv-compile`
  still works, `--fast-deps --no-uv-compile` is rejected, pip extras compose.
- Full `tests/comfy_cli/command` + cm_cli suites: **1282 passed, 2 skipped**.
- `ruff check` + `ruff format --diff`: clean.

### DIFF
--- a/comfy_cli/command/custom_nodes/command.py
+++ b/comfy_cli/command/custom_nodes/command.py
@@ -218,6 +218,15 @@ def restore_snapshot(
         show_default=False,
         help="Restore for pip packages specified by local paths.",
     ),
+    fast_deps: Annotated[
+        bool,
+        typer.Option(
+            "--fast-deps",
+            show_default=False,
+            help="Use the fast (uv) dependency installer, matching `comfy install --fast-deps`. "
+            "For snapshot restore this runs the uv-compile fast path (requires ComfyUI-Manager v4.1+).",
+        ),
+    ] = False,
     uv_compile: Annotated[
         bool | None,
         typer.Option(
@@ -227,6 +236,16 @@ def restore_snapshot(
         ),
     ] = None,
 ):
+    # `--fast-deps` mirrors the `comfy install --fast-deps` UX (issue #217). cm_cli's
+    # `restore-snapshot` does NOT accept `--no-deps`, so the DependencyCompiler-based
+    # fast path used by `install`/`reinstall` is unavailable here; the fast uv path
+    # restore-snapshot *does* support is `--uv-compile` (it sets no-deps + batch-resolves
+    # with uv internally). So `--fast-deps` forwards the uv-compile fast path. Passing it
+    # alongside an explicit `--no-uv-compile` is contradictory.
+    if fast_deps and uv_compile is False:
+        typer.echo("Cannot use --fast-deps with --no-uv-compile", err=True)
+        raise typer.Exit(code=1)
+
     extras = []
 
     if pip_non_url:
@@ -238,8 +257,12 @@ def restore_snapshot(
     if pip_local_url:
         extras += ["--pip-local-url"]
 
+    effective_uv_compile = _resolve_uv_compile(uv_compile)
+    if fast_deps:
+        effective_uv_compile = True
+
     path = os.path.abspath(path)
-    execute_cm_cli(["restore-snapshot", path] + extras, uv_compile=_resolve_uv_compile(uv_compile))
+    execute_cm_cli(["restore-snapshot", path] + extras, uv_compile=effective_uv_compile)
 
 
 @app.command("restore-dependencies", help="Restore dependencies from installed custom nodes")
--- a/tests/comfy_cli/command/nodes/test_restore_snapshot_fast_deps.py
+++ b/tests/comfy_cli/command/nodes/test_restore_snapshot_fast_deps.py
@@ -1,0 +1,76 @@
+"""Tests for `comfy node restore-snapshot --fast-deps` (issue #217, BE-4276).
+
+cm_cli's `restore-snapshot` accepts `--uv-compile` but NOT `--no-deps`, so the
+DependencyCompiler-based fast path used by `install`/`reinstall` is unavailable
+there. `--fast-deps` therefore forwards the uv-compile fast path (which cm_cli's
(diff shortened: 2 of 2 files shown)


bigcat88 (approved):
Approving. I went and checked the premise this whole design rests on against the real ComfyUI-Manager source, because if it were wrong the flag would be wired backwards — and it holds up better than the description claims.

## The mapping isn't just "the available fast path" — it's the only one that works

Read from `ltdrdata/ComfyUI-Manager` at tag **4.2.2** (`cm_cli/__main__.py`, the module `execute_cm_cli` actually invokes as `python -m cm_cli`):

- `install` accepts **both** `--no-deps` and `--uv-compile`, and rejects them together: *"--uv-compile and --no-deps are mutually exclusive."*
- `restore-snapshot` accepts **only** `--uv-compile`. There is no `--no-deps` option on it at all.

So had `restore-snapshot --fast-deps` followed the sibling pattern and reached `execute_cm_cli(fast_deps=True)`, that helper's `elif fast_deps or no_deps: cmd += ["--no-deps"]` would have sent `--no-deps` to a command that doesn't define it — a hard `No such option` from upstream typer. Passing `uv_compile=True` instead (and deliberately *not* forwarding `fast_deps`) is the only wiring that runs.

Better still, the semantics converge rather than diverge. v4.2.2's `restore_snapshot` does:

```python
if uv_compile:
    cmd_ctx.set_no_deps(True)
```

`--uv-compile` **is** the no-deps path there; cm_cli just owns the batch resolve internally instead of comfy-cli's `DependencyCompiler` doing it. So `--fast-deps` means the same thing to a user in both places — "skip per-node dependency resolution, batch-resolve instead" — and only the transport differs. That's a stronger justification than the one in your description, and worth stating there.

## One surface inconsistency worth a look

The flag *combination* rules now differ between siblings, which I confirmed live:

```
$ comfy node install somepack --fast-deps --uv-compile
Cannot use --fast-deps and --uv-compile together        # error

$ comfy node restore-snapshot snap.json --fast-deps --uv-compile
(accepted — both simply mean uv_compile=True)

$ comfy node restore-snapshot snap.json --fast-deps --no-uv-compile
Cannot use --fast-deps with --no-uv-compile             # error, as designed
```

I don't think this blocks: it mirrors upstream's own asymmetry (install has two competing flags to arbitrate between, restore-snapshot has one), and `--fast-deps --uv-compile` here is redundant rather than contradictory, so accepting it is defensible. But an agent that has learned "`--fast-deps` and `--uv-compile` are mutually exclusive on `comfy node`" will find that rule silently not holding on one subcommand. Either rejecting the redundant pair here too, or a line in the help text saying they're synonyms for this subcommand, would close the gap.

## Verification

Full suite green on this branch merged with current `main`; `ruff check` + `ruff format --diff` clean under the CI-pinned 0.15.15. The `--fast-deps --no-uv-compile` guard errors with exit 1 and never reaches `execute_cm_cli`, which I confirmed by running it.

The test set covers the wiring properly — forwards `uv_compile=True`, default stays `False`, explicit `--uv-compile` unaffected, the contradictory pair rejected without dispatch, and pip extras still composing. Good that `test_fast_deps_forwards_pip_extras` asserts on `call_args.args[0]`, since the extras and the flag travel by different routes.

PR #594, opened 2026-07-24
refactor(install): dedupe manager-install fallback and normalize exit conventions (BE-4356)

mattmillerai:
## ELI-5

When you run `comfy install`, the tool installs ComfyUI-Manager and, if that
install fails, quietly turns the manager off so a later launch doesn't try to
use something that isn't there. That little "install-or-degrade" dance was
written out **twice** — once for the normal pip path and once for the
`--fast-deps` path — and the two copies had already started to drift. This PR
folds them into one shared helper so they can't drift again.

While in the file, it also makes the module exit the way the rest of the
codebase does: `raise typer.Exit(code=1)` instead of `sys.exit(...)`.

## What changed

1. **Dedupe the manager-install fallback.** New module-private helper
   `_install_manager_with_fallback(repo_dir, python, *, bootstrap_pip: bool)`
   holds the single copy of "install the manager; on failure disable the
   manager GUI mode and warn." Both call sites in `execute()` (the pip path and
   the `fast_deps` path) now call it. The only behavioral difference between the
   old copies — the `fast_deps` path calls `ensure_pip(python)` first because
   its uv-managed venv may ship no pip, while the pip path already bootstrapped
   pip earlier — is preserved via the `bootstrap_pip` flag (`False` for the pip
   path, `True` for `fast_deps`). Pure code motion; no behavior change.

2. **Normalize exit conventions.** The module's six `sys.exit(...)` calls are
   now `raise typer.Exit(code=1)`, matching the convention used broadly
   elsewhere in the codebase. All of these already propagate straight up to
   `main()`, which re-raises both `typer.Exit` and `SystemExit` identically, so
   the exit code is unchanged (1) for five of them.

## ⚠️ One intentional observable change

`sys.exit(-1)` on the "path exists but isn't a recognized ComfyUI repo" branch
previously produced shell exit code **255** (a negative arg wraps to 255); it
now produces **1**, like every other failure in the module. This is the only
user-visible change and is a deliberate normalization.

## Notes / judgment calls

- **Deferred as ticketed:** the full `_install_deps_pip` / `_install_deps_fast`
  two-strategy extraction is intentionally **not** done here — install is the
  highest-blast-radius command and the groom verdict downgraded scope to just
  the concrete dedupe + exit normalization. Defer that split until install next
  changes for feature reasons.
- **Test update:** `typer.Exit` is **not** a subclass of `SystemExit`, so the
  one existing test asserting `pytest.raises(SystemExit)` on
  `checkout_stable_comfyui` was updated to `pytest.raises(typer.Exit)`. I
  verified none of the six exit sites sit inside a broad `except Exception:` in
  their call path (which would newly swallow a `typer.Exit` where `SystemExit`
  escaped) — they all propagate to `main()` unchanged.
- **New coverage:** added `TestInstallManagerWithFallback` locking the helper's
  bootstrap-vs-not behavior and the degrade-on-failure config write.

## Testing

- Full unit suite green: `2670 passed, 13 skipped` (`uv run pytest tests/comfy_cli`).
- `ruff check` + `ruff format --check` clean on the touched files.

### DIFF
--- a/comfy_cli/command/install.py
+++ b/comfy_cli/command/install.py
@@ -81,7 +81,7 @@ def pip_install_comfyui_dependencies(
 
         if result and result.returncode != 0:
             rprint("Failed to install PyTorch dependencies. Please check your environment (`comfy env`) and try again")
-            sys.exit(1)
+            raise typer.Exit(code=1)
 
         # install directml for AMD windows
         if gpu == GPU_OPTION.AMD and plat == constants.OS.WINDOWS:
@@ -111,7 +111,7 @@ def pip_install_comfyui_dependencies(
     result = subprocess.run([python, "-m", "pip", "install", "-r", "requirements.txt"], check=False)
     if result.returncode != 0:
         rprint("Failed to install ComfyUI dependencies. Please check your environment (`comfy env`) and try again.")
-        sys.exit(1)
+        raise typer.Exit(code=1)
 
 
 def pip_install_manager(repo_dir, python=sys.executable):
@@ -141,6 +141,26 @@ def pip_install_manager(repo_dir, python=sys.executable):
     # Clear cache so find_cm_cli() picks up the newly installed module
     find_cm_cli.cache_clear()
     return True
+
+
+def _install_manager_with_fallback(repo_dir, python, *, bootstrap_pip: bool):
+    """Install ComfyUI-Manager, degrading gracefully when it fails.
+
+    On failure, disable the manager GUI mode so a later ``comfy launch`` doesn't
+    inject manager flags for a manager that isn't actually installed.
+
+    ``bootstrap_pip`` bootstraps pip first (no-op if already present): the
+    fast_deps path leaves a uv-managed venv that may ship no pip, whereas the
+    pip path has already bootstrapped it earlier in ``execute``.
+    """
+    if bootstrap_pip:
+        ensure_pip(python)
+    if not pip_install_manager(repo_dir, python=python):
+        # Manager installation failed - disable to prevent launch issues
+        from comfy_cli.config_manager import ConfigManager
+
+        ConfigManager().set(constants.CONFIG_KEY_MANAGER_GUI_MODE, "disable")
+        rprint("[yellow]Manager not installed. Launch will run without manager flags.[/yellow]")
 
 
 def execute(
@@ -189,7 +209,7 @@ def execute(
             checkout_stable_comfyui(version=version, repo_dir=repo_dir, url=url)
         except GitHubRateLimitError as e:
             rprint(f"[bold red]Error checking out ComfyUI version: {e}[/bold red]")
-            sys.exit(1)
+            raise typer.Exit(code=1) from e
 
     elif not check_comfy_repo(repo_dir)[0]:
         # Get actual remote URL for better error message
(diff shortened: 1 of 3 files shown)


bigcat88 (approved):
Approving — verified empirically, including the one claim in this PR that could have gone wrong quietly.

## The `sys.exit` → `typer.Exit` swap is the risky part, and it's clean

This is the change worth scrutinizing, because the two aren't interchangeable:

```
typer.Exit MRO: Exit → RuntimeError → Exception → BaseException
issubclass(typer.Exit, Exception)  -> True
issubclass(SystemExit, Exception)  -> False
```

So any `except Exception:` between the raise and `main()` would newly **swallow** an exit that `SystemExit` sailed straight through. Your description says you checked this; I checked it independently rather than take it on trust, with an AST walk that flags every enclosing `try` whose handlers catch `Exception` / `BaseException` / `RuntimeError` / bare `except:`.

All six changed sites are clean, and so is the whole path out: `install.py:85, 115, 214, 233, 615, 642`, the two call sites that reach them (`install.py:211, 247`), and both entry points in `cmdline.py:533, 598`. `tracking.track_command` is a pass-through wrapper with no handler, and `main()` re-raises. Confirmed live on both branches — `comfy install --version v0.0.0-nope` exits **1** with byte-identical output before and after, so the five "unchanged" sites really are unchanged.

## The 255 → 1 change is real, and exactly as documented

Reproduced against a directory that exists as a git repo but isn't ComfyUI:

```
main:        EXIT CODE = 255      # sys.exit(-1) wraps
this branch: EXIT CODE = 1
```

Nothing in `comfy_cli/`, `docs/`, or `.github/` keys on 255, so no caller can be relying on it. It's a nonsense code produced by a negative argument, every other failure in the module is 1, and you flagged it prominently — normalizing is right.

## The dedupe really is pure code motion

I diffed both original copies out of `origin/main` against the helper instead of trusting "no behavior change". They collapse exactly: identical `pip_install_manager` call, identical `ConfigManager().set(MANAGER_GUI_MODE, "disable")`, identical warning string. The only delta between the two copies was the `ensure_pip(python)` that the fast_deps path ran first, and that is faithfully preserved by `bootstrap_pip` (`False` for pip, `True` for fast_deps).

Nice touch that the function-local `from comfy_cli.config_manager import ConfigManager` is what lets the new tests patch `comfy_cli.config_manager.ConfigManager` at the source and have it take effect.

Full unit suite green on this branch merged with current `main`; `ruff check` and `ruff format --diff` clean under the CI-pinned **0.15.15** (worth using that pin locally — 0.15.13 disagrees with it on docstring formatting, which is currently making a sibling PR's CI red).

## One pre-existing bug I hit while tracing (not yours, not blocking)

The same AST sweep flagged `install.py:333`, which is untouched by this PR and identical on `main`:

```python
        if not pr_info:
            rprint(f"[bold red]PR not found: {pr_ref}[/bold red]")
            raise typer.Exit(code=1)          # <- line 333

    except Exception as e:                    # <- line 335 catches it
        rprint(f"[bold red]Error fetching PR information: {e}[/bold red]")
        raise typer.Exit(code=1)
```

The `except Exception` catches its own `typer.Exit`, and since `click.exceptions.Exit` carries no args, `{e}` renders empty — so `comfy install --pr <missing>` prints the correct `PR not found: …` and then immediately contradicts it with a blank-reason `Error fetching PR information: `. Exit code is 1 either way, so it's cosmetic. Worth `except typer.Exit: raise` or narrowing the handler next time this function is touched — and it's a good illustration of why the swap you did here needed checking.

PR #596, opened 2026-07-24
refactor(git_utils): shared git-error panel printer + cwd= over chdir guards

mattmillerai:
## ELI-5

Two functions in `git_utils.py` — `git_checkout_tag` and `checkout_pr` — each did the same two clunky things: (1) built a nearly-identical fancy red error box by hand, and (2) `cd`-ed the whole process into the repo folder to run git, then `cd`-ed back in a `finally`. This PR pulls the error box into one shared helper, and tells each `git` command *which folder to run in* (`cwd=`) instead of changing the whole process's working directory.

It also plugs a **real argument-injection hole**: a pull request author controls their own branch name, and we were handing that name straight to `git fetch` as a bare word. A branch literally named `--upload-pack=<some command>` would be read by git as an *option*, not a branch — and git would run that command. Adding `--` before the branch tells git "everything after this is a name, not a flag."

## Security fix: `--` end-of-options separator on `git fetch` (667887d)

`pr_info.head_branch` comes straight from the GitHub API (`data["head"]["ref"]`) and is fully controlled by the PR author. It flowed positionally into `git fetch` on both the fork and non-fork paths. Because `git check-ref-format` accepts `refs/heads/--upload-pack=x`, neither git nor GitHub rejects such a name for us — and `comfy install --pr` is *designed* to run against untrusted forks.

Verified against live git:

```
before:  git fetch origin '--upload-pack=touch …'   → command EXECUTED
after:   git fetch origin -- '--upload-pack=touch …' → fatal: invalid refspec  (blocked)
```

Both `fetch` call sites now pass `["git", "fetch", <remote>, "--", head_branch]`. The remaining git invocations in this function were audited and are safe by construction: the two `git checkout -B` start-points are prefixed (`origin/…`, `<remote>/…`), `local_branch` is prefixed `pr-`, and `git remote add <name> <url>` rejects a dash-leading URL outright. Regression assertions were added to both `checkout_pr` tests (the suite mocks `subprocess.run`, so nothing would otherwise have caught a wrong refspec syntax).

## What else changed

- **New `_print_git_error(title, panel_title, context, details, exc)` helper** renders the shared rich error `Panel` (bold-red title, bold-yellow context line, italic detail lines, optional `stderr` block). Both `except subprocess.CalledProcessError` blocks now call it.
- **`cwd=repo_path` on every `subprocess.run` call** in both functions, replacing the `os.getcwd()` / `os.chdir(repo_path)` / `finally: os.chdir(original_dir)` scaffolding. `import os` is now unused and removed.

## Why the `cwd=` change matters (beyond dedup)

`os.chdir` mutates process-global state. Passing `cwd=` per-call eliminates that mutation, so the functions are safe under any future threaded/concurrent use (no shared-cwd races) and there is no global state left to restore on the error path.

To be precise about what this does *not* buy: a missing `repo_path` still raises an uncaught `FileNotFoundError` — the throw simply moves from `os.chdir` to `subprocess.run(cwd=…)`. That is unchanged from `main`; removing the process-global cwd mutation is the actual win.

Behavior is otherwise preserved: git commands still execute against `repo_path`, and each function still returns `False` on a git failure.

## Behavior note (judgment call)

The **`checkout_pr` error panel is byte-identical** to before. The **`git_checkout_tag` panel has minor cosmetic differences** on the failure path, which I consider a net improvement:
- The tag line was previously appended via `Text.append(f"[cyan]{tag}[/cyan]")`, but `Text.append` does **not** parse console markup — so users literally saw `[cyan]v1.2.3[/cyan]` (brackets and all). It now shows a clean `v1.2.3` in the header line.
- The "Error details:" separator is now italic (was bold-red) with one fewer blank line, matching the shared helper's layout.

If exact byte-for-byte preservation of the old (buggy) tag panel is preferred, I can special-case it — but the current output reads better.

## Testing

- `tests/comfy_cli/command/github/test_pr.py` — the four mocked `checkout_pr` tests dropped their now-dead `@patch("os.chdir")`/`@patch("os.getcwd")` decorators and now assert `cwd=repo_path` is passed to every `subprocess.run`, plus the two `--`-separator regression assertions described above.
- Verified against the branch **merged with current `main`**: `3541 passed, 37 skipped`; `uvx ruff@0.15.15 format --diff .` → `270 files already formatted`; `ruff check` clean.

### DIFF
--- a/comfy_cli/git_utils.py
+++ b/comfy_cli/git_utils.py
@@ -1,4 +1,3 @@
-import os
 import subprocess
 
 from rich.console import Console
@@ -24,6 +23,40 @@ def sanitize_for_local_branch(branch_name: str) -> str:
     return sanitized or "unknown"
 
 
+def _print_git_error(
+    title: str,
+    panel_title: str,
+    context: str,
+    details: list[str],
+    exc: subprocess.CalledProcessError,
+) -> None:
+    """Render a git ``CalledProcessError`` as a rich error panel.
+
+    Shared by ``git_checkout_tag`` and ``checkout_pr`` so the two failure
+    displays stay in lock-step. ``context`` is the bold headline line, each
+    ``details`` entry is an italic follow-up line, and the command's stderr
+    (when present) is appended verbatim.
+    """
+    error_message = Text()
+    error_message.append(title, style="bold red on white")
+    error_message.append(f"\n\n{context}", style="bold yellow")
+    for detail in details:
+        error_message.append(f"\n{detail}", style="italic")
+
+    if exc.stderr:
+        error_message.append("\n\nError output:", style="bold red")
+        error_message.append(f"\n{exc.stderr}", style="italic yellow")
+
+    console.print(
+        Panel(
+            error_message,
+            title=f"[bold white on red]{panel_title}[/bold white on red]",
+            border_style="red",
+            expand=False,
+        )
+    )
+
+
 def git_checkout_tag(repo_path: str, tag: str) -> bool:
     """
     Checkout a specific Git tag in the given repository.
@@ -39,16 +72,12 @@ def git_checkout_tag(repo_path: str, tag: str) -> bool:
     :param tag: The tag to checkout
     :return: True if the checkout succeeds, False if any git command failed.
     """
-    original_dir = os.getcwd()
     try:
-        # Change to the repository directory
-
-        os.chdir(repo_path)
-
         # Skip the network fetch when the tag is already present locally.
         tag_present_locally = (
             subprocess.run(
                 ["git", "rev-parse", "--verify", f"refs/tags/{tag}"],
+                cwd=repo_path,
                 capture_output=True,
                 text=True,
                 check=False,
@@ -56,62 +85,49 @@ def git_checkout_tag(repo_path: str, tag: str) -> bool:
             == 0
         )
         if not tag_present_locally:
-            subprocess.run(["git", "fetch", "--tags"], check=True, capture_output=True, text=True)
+            subprocess.run(["git", "fetch", "--tags"], cwd=repo_path, check=True, capture_output=True, text=True)
 
         # Checkout the specified tag
-        subprocess.run(["git", "checkout", tag], check=True, capture_output=True, text=True)
+        subprocess.run(["git", "checkout", tag], cwd=repo_path, check=True, capture_output=True, text=True)
 
         console.print(f"[bold green]Successfully checked out tag: [cyan]{tag}[/cyan][/bold green]")
 
         return True
     except subprocess.CalledProcessError as e:
-        error_message = Text()
-        error_message.append("Git Checkout Error", style="bold red on white")
-        error_message.append("\n\nFailed to checkout tag: ", style="bold yellow")
-        error_message.append(f"[cyan]{tag}[/cyan]")
-        error_message.append("\n\nError details:", style="bold red")
-        error_message.append(f"\n{str(e)}", style="italic")
-
-        if e.stderr:
-            error_message.append("\n\nError output:", style="bold red")
-            error_message.append(f"\n{e.stderr}", style="italic yellow")
-
-        console.print(
-            Panel(
-                error_message,
-                title="[bold white on red]Git Checkout Failed[/bold white on red]",
-                border_style="red",
-                expand=False,
-            )
+        _print_git_error(
+            title="Git Checkout Error",
+            panel_title="Git Checkout Failed",
+            context=f"Failed to checkout tag: {tag}",
+            details=[f"Error details:\n{e}"],
+            exc=e,
         )
-
         return False
-    finally:
-        # Ensure we always return to the original directory
-        os.chdir(original_dir)
 
 
 def checkout_pr(repo_path: str, pr_info: PRInfo) -> bool:
-    original_dir = os.getcwd()
-
     try:
-        os.chdir(repo_path)
-
         if pr_info.is_fork:
             remote_name = f"pr-{pr_info.number}-{pr_info.user}"
 
-            result = subprocess.run(["git", "remote", "get-url", remote_name], capture_output=True, text=True)
+            result = subprocess.run(
+                ["git", "remote", "get-url", remote_name], cwd=repo_path, capture_output=True, text=True
+            )
 
             if result.returncode != 0:
                 subprocess.run(
                     ["git", "remote", "add", remote_name, pr_info.head_repo_url],
+                    cwd=repo_path,
                     check=True,
                     capture_output=True,
                     text=True,
                 )
 
             subprocess.run(
-                ["git", "fetch", remote_name, pr_info.head_branch], check=True, capture_output=True, text=True
+                ["git", "fetch", remote_name, pr_info.head_branch],
+                cwd=repo_path,
+                check=True,
+                capture_output=True,
+                text=True,
             )
 
             # fix: "feature/add-support" -> "pr-123-feature-add-support"
@@ -120,19 +136,27 @@ def checkout_pr(repo_path: str, pr_info: PRInfo) -> bool:
 
             subprocess.run(
                 ["git", "checkout", "-B", local_branch, f"{remote_name}/{pr_info.head_branch}"],
+                cwd=repo_path,
                 check=True,
                 capture_output=True,
                 text=True,
             )
 
         else:
-            subprocess.run(["git", "fetch", "origin", pr_info.head_branch], check=True, capture_output=True, text=True)
+            subprocess.run(
+                ["git", "fetch", "origin", pr_info.head_branch],
+                cwd=repo_path,
+                check=True,
+                capture_output=True,
+                text=True,
+            )
 
             sanitized_branch = sanitize_for_local_branch(pr_info.head_branch)
             local_branch = f"pr-{pr_info.number}-{sanitized_branch}"
 
             subprocess.run(
                 ["git", "checkout", "-B", local_branch, f"origin/{pr_info.head_branch}"],
+                cwd=repo_path,
                 check=True,
                 capture_output=True,
                 text=True,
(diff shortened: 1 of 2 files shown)


comfy_cli/git_utils.py
```diff
@@ -39,79 +72,62 @@ def git_checkout_tag(repo_path: str, tag: str) -> bool:
...
                     ["git", "remote", "add", remote_name, pr_info.head_repo_url],
+                    cwd=repo_path,
                     check=True,
                     capture_output=True,
                     text=True,
                 )
 
             subprocess.run(
-                ["git", "fetch", remote_name, pr_info.head_branch], check=True, capture_output=True, text=True
+                ["git", "fetch", remote_name, pr_info.head_branch],
```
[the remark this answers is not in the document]
mattmillerai: Fixed in 667887d. Added a `--` end-of-options separator before the branch refspec on both the fork (`git fetch <remote> -- <branch>`) and non-fork (`git fetch origin -- <branch>`) paths, so a fork-controlled branch name beginning with `-` can no longer be parsed as a git option. Regression assertions added to both checkout_pr tests. The `git checkout -B` / `git remote` args in this function are all prefixed (`origin/…`, `pr-<n>-…`) so they cannot begin with a dash and are unaffected.

tests/comfy_cli/command/github/test_pr.py
```diff
@@ -179,7 +179,7 @@ def test_find_pr_by_branch_error(self, mock_get):
 
     @patch("requests.get")
     def test_find_pr_by_branch_rate_limit(self, mock_get):
-        """A rate-limited 403 surfaces as GitHubRateLimitError, not a silent "no PR found\""""
+        """A rate-limited 403 surfaces as GitHubRateLimitError, not a silent "no PR found\" """
```
bigcat88: This line is what makes `ruff_check` red, and it's the only blocker on this PR.

`main` has this docstring already correctly formatted; this branch rewrites it and adds a space before the closing quotes, which `ruff format` **0.15.15** (the version CI pins) rejects:

```
1 file would be reformatted
-        """A rate-limited 403 ... not a silent "no PR found\" """
+        """A rate-limited 403 ... not a silent "no PR found\""""
```

Please restore `main`'s version:

```python
        """A rate-limited 403 surfaces as GitHubRateLimitError, not a silent "no PR found\""""
```

ruff 0.15.13 accepts both forms, which is why this looks clean locally — `uvx ruff@0.15.15 format --diff .` reproduces CI exactly. It's also unrelated to this PR's purpose, so it looks like an accidental editor touch.

bigcat88 (changes requested):
Requesting changes on one line — everything else here is verified and good, including a security fix that's better than the description lets on.

## Blocking: this PR turns `main`'s `ruff_check` red

`ruff_check` is failing, and it's PR-caused, not stale CI. The branch **rewrites an already-correctly-formatted docstring into a form ruff rejects**:

```diff
-        """A rate-limited 403 surfaces as GitHubRateLimitError, not a silent "no PR found\""""
+        """A rate-limited 403 surfaces as GitHubRateLimitError, not a silent "no PR found\" """
```

Proof it's this branch and not the environment — same ruff, both trees, this branch merged with current `main`:

```
$ uvx ruff@0.15.15 format --diff .    # on origin/main
270 files already formatted

$ uvx ruff@0.15.15 format --diff .    # on this branch
1 file would be reformatted
```

Why it passed for you: CI pins **ruff 0.15.15** (`.github/workflows/ruff_check.yml`) and runs `ruff format --diff`, but 0.15.13 accepts this line — I hit the same false-green locally until I matched the pin. Your "ruff check clean" claim is literally true; it's `ruff format` that fails. Worth running `uvx ruff@0.15.15 format --diff .` before pushing.

Revert that one line to `main`'s version and this is ready — nothing else needs to change.

## The `--` fix is a real RCE guard, and it isn't in the description at all

Commit 667887d (`-- separator on git fetch`) is the most valuable thing in this PR and the body never mentions it. I verified the vulnerability is real against live git 2.39.5, using a branch name of `--upload-pack=touch <path>`:

```
main's form:  git fetch origin '--upload-pack=touch …'
              → command EXECUTED (marker file created)

this branch:  git fetch origin -- '--upload-pack=touch …'
              → fatal: invalid refspec '--upload-pack=touch …'   (blocked)
```

`pr_info.head_branch` comes straight from the GitHub API and flows into `comfy install --pr` against **forks**, and `git check-ref-format` accepts `refs/heads/--upload-pack=x`, so the CLI can't lean on git or GitHub refusing the name. This is worth its own line in the description (and arguably its own PR) so it isn't buried in a refactor.

I checked the calls you did *not* add `--` to, and they're all safe by construction: the two `git checkout -B` start-points are prefixed (`origin/…`, `remote/…`), `local_branch` is prefixed `pr-`, and `git remote add <name> <url>` rejects a dash-leading URL with `error: unknown option` before doing anything (verified). So the two `fetch` calls were the whole hole.

Also confirmed `git fetch <remote> -- <branch>` is accepted by git — the tests mock `subprocess.run`, so nothing in the suite would have caught it if the syntax were wrong.

## The `cwd=` refactor checks out

Verified against real git repos, not mocks:

- real repo + tag → returns `True`, **process cwd unchanged**
- missing tag → returns `False` via the shared panel
- both panels render; `checkout_pr`'s is byte-identical as claimed

Fixing the `Text.append(f"[cyan]{tag}[/cyan]")` bug is a genuine improvement — `Text.append` doesn't parse markup, so users really were seeing literal `[cyan]v1.2.3[/cyan]`. And because `Text.append` doesn't parse markup, a tag or PR title containing `[...]` can't raise `MarkupError` through the new helper.

**One description nit:** the body says `cwd=` "removes the failure mode where `os.chdir(repo_path)` itself throws (e.g. missing dir)". It doesn't — the throw just moves from `os.chdir` to `subprocess.run(cwd=…)`. Both `main` and this branch raise an uncaught `FileNotFoundError` for a missing repo dir; I checked both. No regression, but the claim overstates it — what you actually removed is the process-global cwd mutation, which is a good enough reason on its own.

(Unrelated and pre-existing on both branches: `import comfy_cli.git_utils` on its own raises `ImportError` from a cycle with `comfy_cli/command/github/__init__.py`. Not yours, not in scope — just noting it since it bites anyone importing the module directly.)

tests/comfy_cli/command/github/test_pr.py
mattmillerai: Fixed in c85f33d — reverted to `main`'s exact docstring form.

You're right that it was an accidental editor touch, and right about the root cause: I had validated with a ruff older than the CI pin, so `ruff format` passed locally and my "ruff check clean" claim was true but beside the point. Re-verified this time with the pinned version on the branch **merged with current `main`**:

```
uvx ruff@0.15.15 format --diff .   → 270 files already formatted
uvx ruff@0.15.15 check .           → All checks passed!
uv run pytest tests/              → 3541 passed, 37 skipped
```

Thanks for the independent verification of the `--` fix against live git — I've now given it its own section in the PR description with the `--upload-pack` repro, plus the audit of the call sites that intentionally don't need `--`.

Also took the description nit: the body no longer claims `cwd=` removes the missing-directory throw. It now says explicitly that a missing `repo_path` still raises an uncaught `FileNotFoundError` — the throw just moves from `os.chdir` to `subprocess.run(cwd=…)` — and that removing the process-global cwd mutation is the actual win.

On the `import comfy_cli.git_utils` cycle: reproduced it, and confirmed it's pre-existing on `main` and untouched by this PR. No current caller hits it (the CLI entry point's import order resolves fine); it only bites a direct import of the module. Leaving it out of this PR and not filing a ticket for it since nothing reaches it today — but noting it here so it's findable.

bigcat88:
Follow-up: I found the **root cause**, and it isn't a stray editor touch — your local ruff is rewriting that line for you.

I ran each ruff version's `format` against `main`'s copy of `tests/comfy_cli/command/github/test_pr.py`, inside the repo so the project config applies:

```
ruff 0.14.0   -> REWRITES the docstring to the spaced form   (1 line changed)
ruff 0.14.4   -> REWRITES the docstring to the spaced form   (1 line changed)
ruff 0.15.0   -> leaves it alone
ruff 0.15.15  -> leaves it alone   (this is what CI pins)
```

So ruff **0.14.x** turns

```python
"""… not a silent "no PR found\""""
```

into

```python
"""… not a silent "no PR found\" """
```

and 0.15.15 rejects that. Running `ruff format` locally on 0.14.x silently introduces the change, you commit it, and CI goes red. That also explains the note in #630's description — *"one pre-existing reformat in `tests/comfy_cli/command/github/test_pr.py` … present on `main` — likely local-vs-CI ruff version drift."* It isn't pre-existing: `main` is clean under 0.15.15 (`270 files already formatted`). Your 0.14.x is generating it every time it touches that file.

**Fix that outlasts this PR:** pin your local ruff to CI's version. Either

```bash
uvx ruff@0.15.15 check . && uvx ruff@0.15.15 format --diff .
```

or add `ruff==0.15.15` to the dev extra so `uv run ruff` matches `.github/workflows/ruff_check.yml`. Otherwise this will keep landing in any PR that touches a file with a docstring ending in an escaped quote.

For this PR specifically: revert that one line to `main`'s version and everything else is ready to merge — the `--` argument-injection guard verified clean and the `cwd=` refactor checked out against real git repos.

bigcat88 (approved):
Fixed — clearing my earlier changes-requested.

The docstring is now **byte-identical to `main`**:

```
this branch: 183:        """A rate-limited 403 … not a silent "no PR found\""""
main:        183:        """A rate-limited 403 … not a silent "no PR found\""""
```

and CI's pinned toolchain is clean on both halves:

```
$ uvx ruff@0.15.15 check .          -> All checks passed!
$ uvx ruff@0.15.15 format --diff .  -> 272 files already formatted
```

The fix commit touched exactly **1 file, 1 line** — nothing else moved. Both `git fetch … --` guards are still in place (`git_utils.py:128` and `:150`), so the argument-injection fix I verified earlier is intact: a branch named `--upload-pack=touch <path>` executed on `main`'s form and is rejected as an invalid refspec here.

Full suite re-run green on this branch merged with current `main`.

Everything from the earlier review stands — the `cwd=` refactor checked out against real git repos (process cwd unchanged, `False` on a bad tag, both panels rendering), and the `Text.append` markup bug fix is a genuine improvement.

Worth acting on the root cause I posted separately so this doesn't recur: your local ruff is **0.14.x**, which rewrites that docstring automatically; 0.15.x leaves it alone. Pinning to `ruff==0.15.15` locally will stop it re-appearing in any PR that touches a file with a docstring ending in an escaped quote.

PR #597, opened 2026-07-24
refactor(http): shared is_cloud-gated target_auth_headers builder; migrate 5 call sites (BE-4362)

mattmillerai:
## ELI-5

Five different places in the code each wrote their own little "attach the login token to this request" snippet. They'd drifted: three of them attached the token to **any** target — even a plain-HTTP local ComfyUI — instead of only to Comfy Cloud. This PR replaces all five copies with **one** shared helper, `target_auth_headers(target)`, that attaches credentials **only when `target.is_cloud`**. Same behavior in practice today (local targets are built with no credentials), but a stray token on a local target can now never leak to a plaintext server.

## What changed

Adds `target_auth_headers(target) -> dict[str, str]` to `comfy_cli/http.py` — the strictest variant (promoted verbatim from `transfer._auth_headers`): it returns `{}` unless `target.is_cloud`, then `X-API-Key` (api_key) taking precedence over `Authorization: Bearer` (auth_token). This is the same defense-in-depth gate `comfy_client.Client` already uses.

Migrated the five hand-rolled non-client copies onto it:

| Site | Change |
|---|---|
| `command/workflow.py` `_authed_request` | lazy import + header loop (was **ungated** — now gated) |
| `command/models/search.py` `_authed_request` | lazy import + header loop (was **ungated** — now gated) |
| `command/jobs.py` `_cloud_cancel` (inline) | lazy import + header loop (was **ungated** — now gated) |
| `command/transfer.py` `_auth_headers` | deleted body, re-exported: `from comfy_cli.http import target_auth_headers as _auth_headers` (was already gated — byte-identical) |
| `cql/engine.py` `_load_from_target` (inline) | header loop (was already gated, api_key-first — byte-identical) |

**Left untouched by design:** `comfy_client.py:264-268` (superset path with safe-URL assertion + 401 refresh-retry), and every HTTPError→envelope mapper / capped-reader (a separate follow-up ticket). Non-Target header sites (`generate/client.py`'s bare-`api_key` builder, GitHub/Civitai token headers in `install.py`/`models.py`) are out of scope — they don't route a `Target`.

## Security note

The gate **direction** is the point: attach **nothing** unless `target.is_cloud`. The three previously-ungated sites (workflow/search/jobs) attached credentials whenever they were present, regardless of target kind. Behavior is preserved today because `resolve_target` builds local Targets with `auth_token=None` and no `api_key` (`target.py:118-127`), so no observable change — but the latent leak path (a stray credential on a local target → plaintext server) is now closed. This also means local `/userdata` calls in `workflow.py` go from "attach if creds present" to "never attach," which `_userdata_request`'s docstring already described as a no-op.

## Tests

- Added 4 cases to `tests/comfy_cli/test_http.py`, including the security property: a **local** Target with **both** `auth_token` and `api_key` set → `target_auth_headers` returns `{}`; plus cloud api-key-only, cloud auth-token-only, and cloud-both (api_key wins).
- Full existing suites for all migrated commands pass unchanged: `test_workflow_saved.py`, `models/test_search.py`, `jobs/test_jobs.py`, `test_transfer_{download,redirect,upload}.py`, and the `cql/` suite (`135 passed` + `248 passed, 1 skipped`). Their target fixtures are cloud-kind, so headers still attach.

Phase 1 of BE-4350.

### DIFF
--- a/comfy_cli/command/jobs.py
+++ b/comfy_cli/command/jobs.py
@@ -1065,10 +1065,10 @@ def _cloud_cancel(prompt_id: str) -> None:
     # upstream too; encoding here is defense in depth.
     url = target.url("jobs", urllib.parse.quote(prompt_id, safe=""), "cancel")
     req = urllib.request.Request(url, data=b"", method="POST")
-    if target.api_key:
-        req.add_header("X-API-Key", target.api_key)
-    elif target.auth_token:
-        req.add_header("Authorization", f"Bearer {target.auth_token}")
+    from comfy_cli.http import target_auth_headers
+
+    for k, v in target_auth_headers(target).items():
+        req.add_header(k, v)
 
     try:
         with urllib.request.urlopen(req, timeout=15) as resp:
--- a/comfy_cli/command/models/search.py
+++ b/comfy_cli/command/models/search.py
@@ -99,11 +99,11 @@ def _models_path_parts(target) -> tuple[str, ...]:
 
 
 def _authed_request(url: str, target) -> urllib.request.Request:
+    from comfy_cli.http import target_auth_headers
+
     req = urllib.request.Request(url)
-    if target.api_key:
-        req.add_header("X-API-Key", target.api_key)
-    elif target.auth_token:
-        req.add_header("Authorization", f"Bearer {target.auth_token}")
+    for k, v in target_auth_headers(target).items():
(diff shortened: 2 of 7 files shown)


skishore23 (approved):
LGTM — land before #530

PR #606, opened 2026-07-27
feat: comfy update comfy --version <X> — headless version switch/rollback (BE-4763)

mattmillerai:
## ELI-5

Today, the only way to move an existing ComfyUI folder to a specific version is `comfy install --version X --restore`, which can stop and ask you what GPU you have, has no answer for uncommitted edits sitting in the folder, and can't roll you *forward* to nightly once you've rolled back. This adds `comfy update comfy --version <X>`: it checks the version exists first, tucks any uncommitted work into a git stash (and tells you where), checks out the version, and reinstalls that version's requirements. It never asks a question, so scripts and the MCP wrapper can use it.

## What changed

New `switch_comfyui_version(comfy_path, version, *, stash=True, url=None)` in `comfy_cli/command/install.py`, sitting next to `checkout_stable_comfyui`, plus the `--version` / `--no-stash` wiring on the existing `update` command.

1. **Resolve + validate before touching anything.** `git fetch --tags` (failure tolerated — a locally-present tag still checks out offline). For a semver target, `git rev-parse --verify refs/tags/v{X}` must succeed; if it doesn't, the command exits non-zero listing the ~5 nearest `v*` tags by semver order and the working tree is untouched. `latest` reuses `_resolve_latest_tag_from_local` with the existing GitHub-API fallback. `nightly` resolves the default branch via `git symbolic-ref refs/remotes/origin/HEAD`, falling back to `master`.
2. **Stash by default.** A non-empty `git status --porcelain` triggers `git stash push -u -m "comfy-cli: before switch to <target>"`. The stash is never popped or dropped automatically and its ref is printed/returned. `--no-stash` refuses to proceed on a dirty tree instead.
3. **Checkout.** Tag target: `git checkout v{X}` (detached HEAD, expected). Nightly target: `git checkout <default-branch>` then `git pull` — this is both the roll-forward gap and the detached-HEAD-breaks-`update` fix.
4. **Reinstall deps** via the existing update-path machinery (`resolve_workspace_python` + `ensure_pip` + `python -m pip install -r requirements.txt`). Torch is not touched and no GPU flag is required.
5. **Report.** `data = {previous, current, stashed, stash_ref}` via the renderer, with a registered `update` output schema; pretty mode prints the same information plus the detached-HEAD note.

Five error codes are registered in `comfy_cli/error_codes.py`: `update_version_target_invalid`, `version_switch_unknown_version`, `version_switch_dirty_tree`, `version_switch_failed`, `version_switch_deps_failed`.

## Failure behavior

- Validation/fetch failure → exit 1, working tree untouched (test asserts no `checkout` and no `stash push` ran).
- Checkout (or the nightly `git pull`) failing after a stash → exit 1, and the message says explicitly that the stash still exists and how to recover it.
- pip failure after a successful checkout → exit 1 with a message saying the tree IS on the new version, deps may be incomplete, and re-running is idempotent.

## Tests

`tests/comfy_cli/command/test_update_version.py` — 25 tests, subprocess fully mocked (a `FakeGit` dispatcher; nothing touches a real repo or the network). Covers every case the ticket asked for: happy path checkout + pip, unknown version (non-zero, no checkout/stash, tags listed), dirty tree stashes before checkout, `--no-stash` refuses, nightly rolls a detached HEAD forward onto the default branch and pulls, `--version` with `all`/`cli` errors before any git call, `validate_version` rejection surfacing as a typer error, and pip failure after checkout. Plus a regression test pinning that a bare `comfy update comfy` still does `git pull` exactly as before.

`uv run pytest tests/` → 2792 passed, 37 skipped, 0 failed. `ruff check` and `ruff format --diff` (CI-pinned 0.15.15) both clean.

## Docs

README gains an "Updating ComfyUI" section covering the whole `update` command and a "Switching to a specific version" subsection documenting stash-by-default, `--no-stash`, and that a tag checkout leaves a detached HEAD you roll forward with `--version nightly`/`latest`. The `--version` help text says the same.

## Self-review notes / judgment calls

- **Nothing existing was flipped, broadened, or removed.** The only change to the pre-existing code path is indentation — the bare `comfy update` branch moved under an `else:`. `test_plain_update_still_pulls` pins that it still runs `git pull` and the same pip command.
- **Capability-denial check.** The diff's negative outcomes are not product-capability denials: `version_switch_unknown_version` is decided at runtime by actually fetching from the remote and running `git rev-parse` against the real tag list, and it *redirects* by naming the nearest versions that do work; `version_switch_dirty_tree` only fires when the user explicitly asks for it with `--no-stash`; the `all`/`cli` rejection redirects to `comfy update comfy --version`. No path is closed off — this diff is purely additive capability.
- **`--no-stash` without `--version` is silently ignored** rather than an error, since it has no meaning for a plain `git pull` update. Documented as applying only with `--version`.
- **Known minor:** if `comfy_path` is not a git repository at all, a semver target reports "version not found" (with an offline hint) rather than "not a git repo", because the tag lookup runs before the `git status` health check. Both exit non-zero and touch nothing; the message just names the less likely cause. Deliberate — reordering would move the health check ahead of the "validate before touching anything" step.
- **`switch_comfyui_version` takes a `url` parameter** used only for the `latest` GitHub-API fallback (mirroring `checkout_stable_comfyui`). The `update` command doesn't pass one, so that fallback queries `comfyanonymous/ComfyUI`, matching today's behavior for a workspace whose remote isn't recorded on the command line.

Closes BE-4763.

### DIFF
--- a/README.md
+++ b/README.md
@@ -99,6 +99,32 @@ dependencies using the following precedence:
      tool environment): a `.venv` is created inside the ComfyUI workspace.
      Use `comfy launch` to start ComfyUI with the correct Python.
 
+### Updating ComfyUI
+
+`comfy update` brings an existing workspace up to date:
+
+- `comfy update` (or `comfy update comfy`): pull the branch the workspace is currently on and reinstall `requirements.txt`.
+- `comfy update all`: also update every installed custom node.
+- `comfy update cli`: upgrade comfy-cli itself.
+
+#### Switching to a specific version
+
+`comfy update comfy --version <X>` moves an existing workspace to a specific ComfyUI version — a downgrade (rollback) or an upgrade — without prompting for anything, so it is safe to run headlessly or from a script. `<X>` is `nightly` (the repo's default branch), `latest` (the newest stable release), or a version number such as `0.3.0` (a leading `v` is optional).
+
+```bash
+comfy update comfy --version 0.3.0      # roll back to the v0.3.0 release
+comfy update comfy --version latest     # newest stable release
+comfy update comfy --version nightly    # roll forward to the default branch
+```
+
+Behavior worth knowing:
+
+- **The target is validated before anything is touched.** An unknown version exits non-zero, lists the nearest available versions, and leaves the working tree exactly as it was.
+- **Uncommitted changes are stashed by default** (`git stash push -u`) and are *never* popped or dropped automatically — the stash ref is printed so you can restore them with `git stash pop`. Pass `--no-stash` if you would rather the command refuse to run on a dirty tree.
+- **A version number checks out a tag, which leaves a detached HEAD.** That is expected. Roll forward again with `comfy update comfy --version nightly` (or `--version latest`); a plain `comfy update` cannot advance a detached HEAD.
+- **Dependencies are reinstalled** from the target version's `requirements.txt`. PyTorch is deliberately left alone: the ComfyUI version doesn't determine your torch build, your machine does. If the dependency install fails, the command exits non-zero and says so — the tree is already on the new version, and re-running the same command is safe.
+- `--version` and `--no-stash` apply only to target `comfy`; combining `--version` with `all` or `cli` is an error.
+
 ### Specifying execution path
 
 - You can specify the path of ComfyUI where the command will be applied through path indicators as follows:
--- a/comfy_cli/cmdline.py
+++ b/comfy_cli/cmdline.py
@@ -39,7 +39,7 @@ from comfy_cli.command import transfer as transfer_inner
 from comfy_cli.command import (
     workflow as workflow_command,
 )
-from comfy_cli.command.install import validate_version
+from comfy_cli.command.install import validate_optional_version, validate_version
 from comfy_cli.command.launch import launch as launch_command
 from comfy_cli.command.launch import logs as logs_command
 from comfy_cli.command.models import models as models_command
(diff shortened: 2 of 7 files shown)


bigcat88 (approved):
Approving. Every one of the 25 tests mocks `subprocess`, so nothing here had ever touched a real repository — for a command whose entire job is `git stash` / `checkout` / `pull` against a user's working tree, that gap was worth closing before merge. I built a throwaway ComfyUI-shaped repo (real tags, real `origin`, real default branch) and ran the actual code against it.

**The data-loss question first, because it's the one that matters.** `git stash push -u` sweeps untracked files, and a real ComfyUI checkout has a *lot* of precious untracked state. I seeded a faithful repo — ComfyUI's real `.gitignore`, an installed `custom_nodes/ComfyUI-Manager/`, `models/checkpoints/sd15.ckpt`, `user/comfyui.log` — plus one genuine source edit, then switched versions:

```
custom_nodes/ComfyUI-Manager  survived: YES
models/checkpoints/sd15.ckpt  survived: YES
user/comfyui.log              survived: YES
source edit                   stashed, 1 recoverable entry
```

Safe, and for a load-bearing reason: ComfyUI gitignores `/custom_nodes/`, `/models/`, `/user/`, `/output/`, `/input/`, and `-u` stashes untracked-but-**not-ignored** files. Worth stating explicitly in the code, because the safety of this feature rests entirely on `-u` never becoming `-a`. My first rig had no `.gitignore` and the switch duly swept `custom_nodes/` away — that's how narrow the margin is. A comment on the `stash push -u` line saying "`-u` not `-a`: ComfyUI gitignores custom_nodes/models/user and those must survive a switch" would protect it from a well-meaning future edit.

**Real-git behaviour, all as documented:**

| scenario | result |
|---|---|
| clean tree → `v0.2.0` | detached at `v0.2.0`, `stashed: false` |
| dirty tree → `v0.1.0` | `stashed: true`, `stash_ref: "stash@{0} (ecbfdcb)"`, tree clean |
| unknown `v9.9.9` | error naming nearest tags; **HEAD unchanged, stash count unchanged** |
| `--no-stash` on dirty tree | refuses, *"nothing was changed"*, edit preserved |
| `nightly` from detached HEAD | back on `master`, rolled forward |
| `git stash pop` afterwards | restores the edit cleanly |

The validate-before-mutate ordering is real, not just intended — I captured HEAD and the stash count around the failing case and both were byte-identical afterwards. That's the property that makes this safe to hand to a script.

**Envelope contract is clean.** `comfy --json update comfy --version v0.2.0` puts exactly **one** valid JSON line on stdout, exit 0; all the git/pip progress noise goes to stderr where it belongs. `update all --version` and `update cli --version` both fail with `update_version_target_invalid` before any git call; unknown version exits 1.

Full suite: **3260 passed**, 7 skipped, 1 failed — `test_non_fast_deps_uses_global_python`, the sandbox artifact present on plain `main` here too. `ruff check` clean. `test_plain_update_still_pulls` does pin the pre-existing path, and I'm satisfied the only change there is the `else:` indentation.

**Judgment calls I agree with:** never auto-popping the stash (the ref is printed and returned — the user decides), `--no-stash` refusing rather than proceeding, and the post-checkout pip failure message stating the tree *is* on the new version and re-running is idempotent. That last one is the difference between a recoverable state and a confusing one.

Two follow-ups, neither blocking:

1. **The mocked-only test suite.** It's thorough on control flow but structurally cannot catch the `-u` vs `-a` class of bug, or a git version behaving differently. One integration test that builds a tiny real repo with a `.gitignore` and asserts an ignored file survives a dirty-tree switch would cover the highest-consequence property here. Cheap, and it's the test I'd most want if someone edits this code in a year.
2. The known `comfy_path`-not-a-git-repo message ("version not found" rather than "not a git repo") is fine as documented — both exit non-zero and touch nothing.

PR #609, opened 2026-07-27
fix(env): detect legacy ComfyUI-Manager clones and reconcile stale manager_gui_mode (BE-4768)

mattmillerai:
## ELI-5

If you installed ComfyUI-Manager the old way — `git clone` into `custom_nodes/` — `comfy env` flatly told you it wasn't installed, because it only ever checked whether the `cm_cli` Python package could be imported and never looked at the disk. And because a single global config key overrode detection entirely, `comfy env` could also claim the Manager was enabled on a machine that has never had it. This PR makes `comfy env` look at both and report what's actually there. Nothing about `comfy launch` changes.

## What changed

- **`find_legacy_manager_clone()`** (`cm_cli_util.py`) — scans `<workspace>/custom_nodes/` for a directory containing the Manager marker file `glob/manager_core.py` **or** `cm-cli.py` at its root. Detection is by marker, not directory name, because people clone the Manager under arbitrary names; directories ending in `.disabled` are skipped. Returns `False` on a missing workspace or any `OSError`. Cached **per workspace path** (it delegates to an `@lru_cache`d `_scan_for_legacy_manager_clone(workspace_path)`), and cleared alongside the existing `find_cm_cli.cache_clear()` in `pip_install_manager()`.
- **`find_cm_cli()` caches keyed per workspace too** — it was `@lru_cache(maxsize=1)` on an argument-less function whose answer depends on the mutable global `workspace_manager.workspace_path`, so one cached answer served every workspace and an unresolved workspace on the first call poisoned the session. It now delegates to `_probe_cm_cli(workspace_path)`. Behaviour for a single fixed workspace is unchanged; both functions still expose `.cache_clear()`.
- **`detect_manager_installation()`** — `"venv-package"` if `find_cm_cli()`, else `"legacy-clone"` if `find_legacy_manager_clone()`, else `"none"`.
- **Reconciliation in the reporting path only** — a new `WorkspaceManager._reported_manager_state()` helper shared by `fill_data()` and `fill_print_table()` so the JSON and table views can't drift apart. The manager-enabling modes (`enable-gui` / `disable-gui` / `enable-legacy-gui`) and `"not-installed"` share one reconciliation branch: with `detected == "none"` they report `"not-installed"`; with an on-disk clone they report the new `"legacy"` value. **`"disable"` is user intent and is never rewritten, nor is any unrecognised mode string.**
- **`manager_detected`** — new additive field on the `comfy env --json` `workspace` payload, alongside the existing `manager_mode`. No existing field was renamed or removed. `schemas/env.json` gains declarations for both (`workspace` was already `additionalProperties: true`, so this documents the contract rather than gating it).
- **`fill_print_table()`** gains `"legacy": "[bold yellow]Legacy clone (cm-cli unavailable)[/bold yellow]"`.

## Launch is untouched (plan item 5)

`comfy launch` flag injection is byte-for-byte unchanged. `_get_manager_flags()` (`launch.py:28-51`) calls `resolve_manager_gui_mode(not_installed_value=None)`, and `resolve_manager_gui_mode()` was not modified — the diff touches `launch.py` not at all. `find_cm_cli()` was refactored to key its cache per workspace path, which changes no answer for a single fixed workspace and so leaves the flags launch injects identical. Its unknown-mode fallback at line 50 can never see `"legacy"`, because `"legacy"` is produced only inside `_reported_manager_state()`, which is called exclusively from `fill_data()` / `fill_print_table()`.

## Verification

`ruff check` + `ruff format --diff` clean; `pytest tests/comfy_cli tests/uv` → **2732 passed, 18 skipped**.

Beyond unit tests, both acceptance criteria were reproduced end-to-end against a scratch ComfyUI workspace with a real `custom_nodes/ComfyUI-Manager/` clone and an isolated `HOME` (so the real user config was untouched):

**QA repro — legacy clone, no pip package, no config keys:**
```
$ comfy --workspace <ws> env          # non-TTY → JSON
"workspace": { "manager_mode": "legacy", "manager_detected": "legacy-clone", ... }
$ comfy --no-json --workspace <ws> env
│ Manager   │ Legacy clone (cm-cli unavailable) │
```
(pre-change this reported `"not-installed"` / `Not Installed`)

**Stale `manager_gui_mode=enable-gui`, no Manager anywhere:**
```
"workspace": { "manager_mode": "not-installed", "manager_detected": "none", ... }
│ Manager   │ Not Installed │
```

### Falsification of the "cm-cli unavailable" claim

This PR adds a user-facing string asserting cm-cli is unavailable for a legacy clone, so I went and checked rather than assuming. With the clone on disk:

```
$ python -c "import cm_cli"
ModuleNotFoundError: No module named 'cm_cli'
$ comfy --workspace <ws> node show installed
ComfyUI-Manager not found. 'cm-cli' command is not available.
```

Every cm-cli entry point in the codebase routes through `execute_cm_cli()`, which runs `[python, "-m", "cm_cli"]` and is gated on `find_cm_cli()`; `grep` confirms no code path invokes a clone's `cm-cli.py` script directly, and `cm-cli.py` (hyphen) is not importable as a module regardless. Note this denial is **pre-existing and untouched** — `execute_cm_cli()`'s hard-fail is unchanged by this diff. The PR does not add a dead-end; it removes a wronger one (`comfy env` previously claimed the Manager was absent entirely).

## Judgment calls

- **`enable-gui` config + legacy clone on disk reports `manager_mode: "legacy"`.** An earlier revision of this PR followed the plan's two literal reconciliation rules and left this case reporting `"enable-gui"`; review ([r3661548777](https://github.com/Comfy-Org/comfy-cli/pull/609#discussion_r3661548777)) correctly called that a contradictory pair (`manager_mode="enable-gui"` + `manager_detected="legacy-clone"`, table saying "GUI Enabled" while cm-cli is unavailable), and `9ea65df` merged the enabling modes into the same branch as `"not-installed"`. An `enable-*` config backed only by a clone overstates what actually works; `manager_detected` still exposes the ground truth for a consumer that needs to distinguish.
- **Reconciliation lives in a shared `_reported_manager_state()` helper** rather than being duplicated into `fill_data()` and `fill_print_table()`. The plan describes the logic twice; duplicating it invites the JSON and table views to drift, which is what item 4 explicitly asks to prevent.
- **`schemas/env.json` was updated** (not called for in the plan). It's purely additive documentation of the two fields — `workspace` already allowed additional properties, so nothing that validated before stops validating. `manager_detected` is declared as an enum of the three values `detect_manager_installation()` can return, which makes a future value addition fail loudly instead of silently escaping the contract.
- **`disable-gui` is included in the stale-config rewrite list.** Despite the name it *enables* the Manager (`--enable-manager --disable-manager-ui`) and only hides the UI, so reporting it with nothing installed would be just as misleading as `enable-gui`. This matches the plan's list.
- One test fixture leaks less than its neighbors: `WorkspaceManager` is a `@singleton`, so the new reconciliation tests save and restore `workspace_path` instead of clobbering the shared instance the way the surrounding `TestFillPrintTable` tests do. I did not retrofit the pre-existing tests — out of scope.

No unmet acceptance criteria.
- **`TestUpdateNodeIdCache::test_uses_resolved_python` needed a `find_cm_cli` stub.** Keying the caches per workspace removed the cross-test `lru_cache` pollution that this test had been relying on (`TestFindCmCli` primed the single global slot with `True` earlier in the suite), so it started failing in CI. It was already broken standalone on `main` — verified by running it against `main`'s `cm_cli_util.py`, where it fails the same way. The guard is incidental to what the test asserts, so it is stubbed rather than the assertion weakened. Caught by @bigcat88 ([r3666335561](https://github.com/Comfy-Org/comfy-cli/pull/609#discussion_r3666335561)).

### DIFF
--- a/comfy_cli/command/custom_nodes/cm_cli_util.py
+++ b/comfy_cli/command/custom_nodes/cm_cli_util.py
@@ -25,19 +25,11 @@ _dependency_cmds = {
 }
 
 
-@lru_cache(maxsize=1)
-def find_cm_cli() -> bool:
-    """Check if cm_cli module is available in the workspace Python.
-
-    First checks the workspace venv Python (primary path — matches the Python
-    used by execute_cm_cli). Falls back to the current Python environment only
-    when the workspace Python is the same as sys.executable.
-
-    Results are cached for the session lifetime.
-    """
-    ws = workspace_manager.workspace_path
-    if ws:
-        python = resolve_workspace_python(ws)
+@lru_cache(maxsize=8)
+def _probe_cm_cli(workspace_path: str | None) -> bool:
+    """Probe for the ``cm_cli`` module for one workspace. Cached per workspace path."""
+    if workspace_path:
+        python = resolve_workspace_python(workspace_path)
         if python != sys.executable:
             # Workspace uses a different Python — check that one
             try:
@@ -54,25 +46,28 @@ def find_cm_cli() -> bool:
     return importlib.util.find_spec("cm_cli") is not None
 
 
-@lru_cache(maxsize=1)
-def find_legacy_manager_clone() -> bool:
-    """Check if ComfyUI-Manager exists as a plain git clone under ``custom_nodes/``.
+def find_cm_cli() -> bool:
+    """Check if cm_cli module is available in the workspace Python.
 
-    This is the pre-pip-package install shape: the Manager repo cloned straight
-    into ``custom_nodes/``. It is invisible to :func:`find_cm_cli`, which only
-    tests whether the ``cm_cli`` module imports in the workspace Python.
+    First checks the workspace venv Python (primary path — matches the Python
+    used by execute_cm_cli). Falls back to the current Python environment only
+    when the workspace Python is the same as sys.executable.
 
-    Detection is by marker file, not directory name — users clone the Manager
-    under arbitrary names. Directories ending in ``.disabled`` (the Manager's own
-    disable convention) are skipped.
-
-    Results are cached for the session lifetime.
+    Results are cached per workspace path, so resolving (or switching) the
+    workspace after a first call re-probes instead of returning a stale answer.
     """
     ws = workspace_manager.workspace_path
-    if not ws:
-        return False
+    return _probe_cm_cli(str(ws) if ws else None)
 
-    custom_nodes = os.path.join(ws, "custom_nodes")
+
+# Preserve the ``lru_cache`` API callers rely on to force a re-probe.
+find_cm_cli.cache_clear = _probe_cm_cli.cache_clear
+
+
+@lru_cache(maxsize=8)
+def _scan_for_legacy_manager_clone(workspace_path: str) -> bool:
+    """Scan one workspace's ``custom_nodes/`` for a Manager clone. Cached per path."""
+    custom_nodes = os.path.join(workspace_path, "custom_nodes")
     try:
         entries = os.listdir(custom_nodes)
     except OSError:
@@ -93,6 +88,30 @@ def find_legacy_manager_clone() -> bool:
             continue
 
     return False
+
+
+def find_legacy_manager_clone() -> bool:
+    """Check if ComfyUI-Manager exists as a plain git clone under ``custom_nodes/``.
+
+    This is the pre-pip-package install shape: the Manager repo cloned straight
+    into ``custom_nodes/``. It is invisible to :func:`find_cm_cli`, which only
+    tests whether the ``cm_cli`` module imports in the workspace Python.
+
+    Detection is by marker file, not directory name — users clone the Manager
+    under arbitrary names. Directories ending in ``.disabled`` (the Manager's own
+    disable convention) are skipped.
+
+    Results are cached per workspace path, so an unresolved workspace on the
+    first call does not poison the answer for the rest of the session.
+    """
+    ws = workspace_manager.workspace_path
+    if not ws:
+        return False
+    return _scan_for_legacy_manager_clone(str(ws))
+
+
+# Preserve the ``lru_cache`` API callers rely on to force a re-scan.
+find_legacy_manager_clone.cache_clear = _scan_for_legacy_manager_clone.cache_clear
 
 
 def detect_manager_installation() -> str:
--- a/comfy_cli/command/custom_nodes/cm_cli_util.py
+++ b/comfy_cli/command/custom_nodes/cm_cli_util.py
@@ -54,6 +54,61 @@ def find_cm_cli() -> bool:
     return importlib.util.find_spec("cm_cli") is not None
 
 
+@lru_cache(maxsize=1)
+def find_legacy_manager_clone() -> bool:
+    """Check if ComfyUI-Manager exists as a plain git clone under ``custom_nodes/``.
+
+    This is the pre-pip-package install shape: the Manager repo cloned straight
+    into ``custom_nodes/``. It is invisible to :func:`find_cm_cli`, which only
+    tests whether the ``cm_cli`` module imports in the workspace Python.
+
+    Detection is by marker file, not directory name — users clone the Manager
+    under arbitrary names. Directories ending in ``.disabled`` (the Manager's own
+    disable convention) are skipped.
+
+    Results are cached for the session lifetime.
+    """
+    ws = workspace_manager.workspace_path
+    if not ws:
+        return False
+
+    custom_nodes = os.path.join(ws, "custom_nodes")
+    try:
+        entries = os.listdir(custom_nodes)
+    except OSError:
+        return False
+
+    for name in entries:
+        if name.endswith(".disabled"):
+            continue
+        node_dir = os.path.join(custom_nodes, name)
+        try:
+            if not os.path.isdir(node_dir):
+                continue
+            if os.path.isfile(os.path.join(node_dir, "glob", "manager_core.py")) or os.path.isfile(
+                os.path.join(node_dir, "cm-cli.py")
+            ):
+                return True
+        except OSError:
+            continue
+
+    return False
+
+
+def detect_manager_installation() -> str:
+    """Report how ComfyUI-Manager is installed for the current workspace.
+
+    Returns ``"venv-package"`` (the pip ``comfyui_manager`` package — cm-cli
+    usable), ``"legacy-clone"`` (an on-disk clone under ``custom_nodes/`` — cm-cli
+    integration unavailable), or ``"none"``.
+    """
+    if find_cm_cli():
+        return "venv-package"
+    if find_legacy_manager_clone():
+        return "legacy-clone"
+    return "none"
+
+
 def resolve_manager_gui_mode(not_installed_value: str | None = None) -> str | None:
     """Resolve manager GUI mode from config, with legacy migration.
 
--- a/comfy_cli/workspace_manager.py
+++ b/comfy_cli/workspace_manager.py
@@ -297,18 +297,49 @@ class WorkspaceManager:
         file_path = os.path.join(self.workspace_path, constants.COMFY_LOCK_YAML_FILE)
         save_yaml(file_path, self.metadata)
 
-    def fill_print_table(self):
-        # Lazy import to avoid circular dependency
-        from comfy_cli.command.custom_nodes.cm_cli_util import resolve_manager_gui_mode
+    def _reported_manager_state(self) -> tuple[str, str]:
+        """Reconcile the configured manager mode against what is actually installed.
 
-        config_manager = ConfigManager()
+        Reporting-only: this never feeds ``comfy launch`` flag injection, which
+        keeps reading :func:`resolve_manager_gui_mode` directly (user intent).
+
+        The global per-user ``manager_gui_mode`` config key can drift out of sync
+        with a given workspace, so ``comfy env`` reconciles the two:
+
+        * any manager-enabling mode (``enable-gui`` / ``disable-gui`` /
+          ``enable-legacy-gui`` — ``disable-gui`` still enables the Manager, it
+          only hides its UI) with no Manager anywhere is stale config → report
+          ``"not-installed"``;
+        * ``"not-installed"`` with an on-disk legacy clone → report ``"legacy"``
+          (the Manager is present, but cm-cli venv integration is unavailable).
+
+        ``"disable"`` is user intent and is always left untouched.
+
+        Returns ``(manager_mode, manager_detected)``.
+        """
+        # Lazy import to avoid circular dependency
+        from comfy_cli.command.custom_nodes.cm_cli_util import detect_manager_installation, resolve_manager_gui_mode
+
+        detected = detect_manager_installation()
         mode = resolve_manager_gui_mode(not_installed_value="not-installed")
+
+        if mode in ("enable-gui", "disable-gui", "enable-legacy-gui") and detected == "none":
+            mode = "not-installed"
+        elif mode == "not-installed" and detected == "legacy-clone":
+            mode = "legacy"
+
+        return mode, detected
+
+    def fill_print_table(self):
+        config_manager = ConfigManager()
+        mode, _detected = self._reported_manager_state()
 
         status_map = {
             "disable": "[bold red]Disabled[/bold red]",
             "enable-gui": "[bold green]GUI Enabled[/bold green]",
             "disable-gui": "[bold yellow]GUI Disabled[/bold yellow]",
             "enable-legacy-gui": "[bold cyan]Legacy GUI[/bold cyan]",
+            "legacy": "[bold yellow]Legacy clone (cm-cli unavailable)[/bold yellow]",
             "not-installed": "[dim]Not Installed[/dim]",
         }
         manager_status = status_map.get(mode, "[bold green]GUI Enabled[/bold green]")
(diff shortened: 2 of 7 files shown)


comfy_cli/workspace_manager.py
```diff
@@ -297,18 +297,49 @@ def save_metadata(self):
...
+        """
         # Lazy import to avoid circular dependency
-        from comfy_cli.command.custom_nodes.cm_cli_util import resolve_manager_gui_mode
+        from comfy_cli.command.custom_nodes.cm_cli_util import detect_manager_installation, resolve_manager_gui_mode
 
-        config_manager = ConfigManager()
+        detected = detect_manager_installation()
         mode = resolve_manager_gui_mode(not_installed_value="not-installed")
 
+        if mode in ("enable-gui", "disable-gui", "enable-legacy-gui") and detected == "none":
```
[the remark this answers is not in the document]
mattmillerai: Valid — fixed in 9ea65df.

The enabling modes and `"not-installed"` now share one reconciliation branch, so `enable-gui`/`disable-gui`/`enable-legacy-gui` + `detected == "legacy-clone"` reports `manager_mode="legacy"` ("Legacy clone (cm-cli unavailable)") instead of the contradictory `enable-gui` + `legacy-clone` pair. `"disable"` and any unrecognised mode string are still passed through untouched.

Covered by new `TestManagerModeReconciliation` cases for all three enabling modes against a legacy clone, the matching-reality cases (`enable-gui`/`disable-gui` + `venv-package` stay put), an unrecognised-mode pass-through, and `test_fill_print_table_does_not_claim_gui_enabled_for_legacy_clone`.

comfy_cli/command/custom_nodes/cm_cli_util.py
```diff
@@ -54,6 +54,61 @@ def find_cm_cli() -> bool:
     return importlib.util.find_spec("cm_cli") is not None
 
 
+@lru_cache(maxsize=1)
```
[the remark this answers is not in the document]
mattmillerai: Valid — fixed in 9ea65df.

`find_legacy_manager_clone()` now delegates to a private `_scan_for_legacy_manager_clone(workspace_path)` cached on the path, so an unresolved workspace on the first call no longer poisons the session and a workspace switch re-scans.

I applied the same treatment to `find_cm_cli()` (→ `_probe_cm_cli(workspace_path)`) rather than just the new function: it has the identical no-arg/mutable-global shape, and `detect_manager_installation()` — new in this PR — depends on it, so fixing only half would have left `comfy env` reporting off a stale `venv-package` probe. Both keep exposing `.cache_clear()` (aliased to the inner helper) so `pip_install_manager` and the existing test fixtures are unchanged.

New tests: `test_cache_is_keyed_on_workspace_path` for both functions, `test_distinct_workspaces_get_distinct_answers`, and `test_result_is_cached_per_workspace` (caching still holds within a workspace).

comfy_cli/command/custom_nodes/cm_cli_util.py
```diff
@@ -25,19 +25,11 @@
...
-    First checks the workspace venv Python (primary path — matches the Python
-    used by execute_cm_cli). Falls back to the current Python environment only
-    when the workspace Python is the same as sys.executable.
-
-    Results are cached for the session lifetime.
-    """
-    ws = workspace_manager.workspace_path
-    if ws:
-        python = resolve_workspace_python(ws)
+@lru_cache(maxsize=8)
```
bigcat88, line 28: This keying change is correct and it is also what turns CI red — worth calling out together.

On `main`, `find_cm_cli()` is `@lru_cache(maxsize=1)` on an argument-less function, so one cached answer serves every workspace. `tests/comfy_cli/command/test_manager_gui.py` runs earlier in the suite and leaves `True` in that slot; `TestUpdateNodeIdCache::test_uses_resolved_python` then passes `update_node_id_cache()`'s `if not find_cm_cli()` guard on that stale value. Bisected on `main`: `test_manager_gui.py` + target => 70 passed; target alone => 1 failed.

Keying per workspace removes the pollution, so that test now fails honestly with `FileNotFoundError: cm-cli not found` — locally and in your CI run.

Fix belongs in the test, not here. It is asserting `resolve_workspace_python` usage, not cm-cli discovery, so add to its `with` block:

```python
patch("comfy_cli.command.custom_nodes.command.find_cm_cli", return_value=True),
```

I applied exactly that and the test passes standalone, independent of suite ordering.

comfy_cli/workspace_manager.py
```diff
@@ -297,18 +297,54 @@ def save_metadata(self):
...
+    def _reported_manager_state(self) -> tuple[str, str]:
+        """Reconcile the configured manager mode against what is actually installed.
+
+        Reporting-only: this never feeds ``comfy launch`` flag injection, which
+        keeps reading :func:`resolve_manager_gui_mode` directly (user intent).
+
+        The global per-user ``manager_gui_mode`` config key can drift out of sync
+        with a given workspace, so ``comfy env`` reconciles the two:
+
+        Every mode that expects a working Manager — the enabling modes
```
bigcat88, line 309: The docstring here is right and the PR body is stale — flagging so the body gets fixed rather than the code.

The "Judgment calls" section still says `enable-gui` + legacy clone reports `manager_mode: "enable-gui"`, not `"legacy"`. I measured it: it reports `legacy`. Commit `9ea65df` ("reconcile enable-* + legacy clone") changed that behaviour and this docstring documents the new rule correctly.

Keep the code as-is — reconciling the enabling modes is the more truthful answer, since an `enable-*` config backed only by a clone overstates what actually works, and `manager_detected` still exposes ground truth for a consumer that needs to distinguish. Just update the PR description before merge so the record matches.

bigcat88 (changes requested):
The production change is correct and I want it merged — but **CI is red on this PR for a reason that is genuinely yours**, not the HuggingFace 429 flake hitting the other branches. One test needs updating; everything else checks out.

**The failure.** `tests/comfy_cli/test_custom_nodes_python_resolution.py::TestUpdateNodeIdCache::test_uses_resolved_python` fails with `FileNotFoundError: cm-cli not found`. I reproduced it locally in the full suite (2878 passed, 2 failed — this one plus the sandbox's `test_non_fast_deps_uses_global_python`), and it is the same failure in your CI run.

**Root cause, and why it's a compliment to the change.** On `main`, `find_cm_cli()` is `@lru_cache(maxsize=1)` on an argument-less function — so it caches one answer and ignores the workspace entirely. I demonstrated that directly:

```
1. fresh find_cm_cli() with repo workspace : False
2. cache_info                              : hits=0, misses=1, maxsize=1
3. find_cm_cli() for a DIFFERENT workspace : False   <- cached, workspace ignored
4. cache_info                              : hits=1, misses=1
```

`tests/comfy_cli/command/test_manager_gui.py` runs earlier in the suite and leaves `True` in that single slot. `test_uses_resolved_python` then sails past `update_node_id_cache()`'s `if not find_cm_cli()` guard on a stale, workspace-blind `True` — it has never actually exercised cm-cli detection. Bisected and confirmed on `main`:

| ordering (on `main`) | result |
|---|---|
| `test_manager_gui.py` + target | **70 passed** |
| `test_install.py` + target | 1 failed |
| target alone | 1 failed |

So your `_probe_cm_cli(workspace_path)` keying (`maxsize=8`) *fixes the exact bug that was propping the test up*. The test was green by accident; the accident is now gone. That's the right outcome — it just needs the test told explicitly.

**Validated fix** — one line in that test, which is about `resolve_workspace_python`, not cm-cli discovery:

```python
patch("comfy_cli.command.custom_nodes.command.find_cm_cli", return_value=True),
```

With that added the test passes standalone (`1 passed`), no longer depending on suite ordering.

**Everything else verified.** I ran ten config × on-disk combinations against scratch workspaces with an isolated `HOME`:

| case | `manager_mode` | `manager_detected` |
|---|---|---|
| clone, no config | `legacy` | `legacy-clone` |
| no manager, stale `enable-gui` | `not-installed` | `none` |
| clone + `enable-gui` | `legacy` | `legacy-clone` |
| clone under arbitrary dir name | `legacy` | `legacy-clone` |
| dir suffixed `.disabled` | `not-installed` | `none` |
| no manager, `disable` | `disable` | `none` |
| clone + `disable` | `disable` | `legacy-clone` |
| clone + `disable-gui` | `legacy` | `legacy-clone` |
| clone + `enable-legacy-gui` | `legacy` | `legacy-clone` |
| clone + unrecognised mode | `some-weird-mode` | `legacy-clone` |

`disable` intent preserved, unrecognised modes passed through untouched, marker-based detection catching a fork named `My-Custom-Manager-Fork`, `.disabled` correctly skipped. Both markers work — I checked the `cm-cli.py` variant separately from `glob/manager_core.py`. Cache is correct across alternating workspaces in one process (`hits=2, misses=4`), and hostile paths (`/etc/shadow`, `/proc/1/root`, a nonexistent dir) all return `False` without raising.

**`comfy launch` really is untouched.** `launch.py` isn't in the diff, and I compared `resolve_manager_gui_mode(not_installed_value=None)` between this branch and `main` for a legacy-clone workspace: `'enable-gui'` and `None` respectively on both. Byte-identical.

**One stale-body note.** The "Judgment calls" section says *"`enable-gui` config + legacy clone on disk reports `manager_mode: "enable-gui"`, not `"legacy"`"* — that's no longer true. Commit `9ea65df` ("reconcile enable-\* + legacy clone") changed it, and the code docstring now correctly documents the newer rule. The **new** behaviour is the better one (an `enable-*` config with only a clone does overstate what works, and `manager_detected` still carries ground truth), so keep the code and update the body before merge.

`ruff check` clean. Re-request me once the test patch is in and I'll re-run the matrix.

comfy_cli/command/custom_nodes/cm_cli_util.py
mattmillerai: Correct on every point, and thanks for bisecting it — fixed in 4e93c0e with exactly the patch you proposed.

I re-derived the root cause independently before applying it, because "my change turned CI red, so I patch the test" is the shape of a fix that ships a real breakage. It holds up:

- **The test never satisfied the guard on its own.** Its fixture writes `custom_nodes/ComfyUI-Manager/cm-cli.py`, but `find_cm_cli()` probes for the importable `cm_cli` *module* — a clone's script never satisfies it. The on-disk file is vestigial with respect to that guard.
- **It was already broken standalone on `main`.** I checked out `main`'s `cm_cli_util.py` into my worktree and ran the target test alone: `FileNotFoundError: cm-cli not found`. Same failure, no changes of mine involved. So this PR did not break the test; it removed the thing that was hiding it.
- **The hiding mechanism is the one you named.** `TestFindCmCli` (`test_manager_gui.py:1053-1070`) calls `find_cm_cli.cache_clear()` and then the real `find_cm_cli()` under mocks, leaving `True` in the argument-less `maxsize=1` slot that every workspace shared.

So the assertion under test (`cmd[0] == "/resolved/python"`) is untouched — only the incidental `find_cm_cli()` guard is stubbed, with a comment recording why the on-disk `cm-cli.py` does not satisfy it so the next reader does not re-litigate this.

Full suite in the worktree: **2743 passed, 18 skipped**. `ruff check` + `ruff format --diff` clean under CI's pinned `ruff==0.15.15`.

comfy_cli/workspace_manager.py
mattmillerai: You are right, and thanks for measuring it rather than assuming — the code and docstring are correct, the PR body was stale. Description updated; code unchanged.

That "Judgment calls" bullet was written against the first revision, which followed the plan's two literal reconciliation rules and left `enable-gui` + legacy clone reporting `enable-gui`. `9ea65df` (responding to [r3661548777](https://github.com/Comfy-Org/comfy-cli/pull/609#discussion_r3661548777), which flagged the same contradiction you did) merged the enabling modes into the `not-installed` branch, and I updated the docstring but not the body. The bullet now describes the shipped behaviour and records why it changed.

While re-reading the body against the diff I found two more stale claims from the same revision and fixed them:

- The `find_legacy_manager_clone()` bullet still said `@lru_cache(maxsize=1)`; it is keyed per workspace path now.
- The "Launch is untouched" section claimed `find_cm_cli()` "was not modified". It was — `9ea65df` refactored it to `_probe_cm_cli(workspace_path)`. The section's actual conclusion still holds (per-workspace keying changes no answer for a single fixed workspace, and `launch.py` is untouched by the diff), but the supporting claim as written was false, so it now says what really happened.

Also added a bullet for the `test_uses_resolved_python` fix from your other thread.

bigcat88 (approved):
Fixed — approving and merging. Commit `4e93c0e` is exactly the stub I suggested, in the right place.

**The test now passes honestly, not by accident:**

- standalone: `1 passed` (it failed here before, on `main` too)
- in the ordering that used to mask it — `test_manager_gui.py` then `test_custom_nodes_python_resolution.py` — `95 passed`

That's the important pair. Passing *only* in suite order was the symptom; passing in both is the fix. The test now states its dependency (`find_cm_cli` → True) instead of inheriting a stale `lru_cache` entry from an unrelated module, so it will keep working regardless of what future cache keying does.

**Detection matrix still correct** after the change — re-ran the cases that matter:

| case | `manager_mode` | `manager_detected` |
|---|---|---|
| clone, no config | `legacy` | `legacy-clone` |
| no manager, stale `enable-gui` | `not-installed` | `none` |
| clone + `disable` (user intent) | `disable` | `legacy-clone` |
| dir suffixed `.disabled` | `not-installed` | `none` |

`disable` intent still preserved, `.disabled` still skipped, and `resolve_manager_gui_mode(not_installed_value=None)` still returns the configured value untouched — `comfy launch` flag injection remains byte-identical to `main`.

Full suite: **3207 passed**, 7 skipped, 1 failed — `test_non_fast_deps_uses_global_python`, the sandbox artifact that fails on plain `main` here too. `ruff check` clean.

Worth restating what this PR actually bought, because it's more than the ticket: keying `_probe_cm_cli` per workspace fixed a real cross-workspace bug — on `main`, `find_cm_cli()` cached one answer and returned it for *every* workspace regardless of path. The red test was that bug's own camouflage falling away.

Still worth updating the description's "Judgment calls" section before this lands in the changelog: it says `enable-gui` + legacy clone reports `manager_mode: "enable-gui"`, but commit `9ea65df` changed that to `legacy`, which is the more truthful answer and what the code docstring now documents.

PR #610, opened 2026-07-27
feat(node): add `comfy node deps` read-only per-pack dependency report (BE-4773)

mattmillerai:
## ELI-5

You installed a bunch of custom node packs. Each one ships a list of Python libraries it needs. Nobody was checking whether your ComfyUI venv actually *has* those libraries at the right versions — so a pack silently half-works and you find out at node-execution time.

`comfy node deps` reads every pack's requirement list, asks the workspace venv once "what do you actually have installed?", and prints a per-requirement verdict: **satisfied**, **mismatch** (installed, but the wrong version), **missing**, or **unparseable** (a `-r other.txt` / `--extra-index-url` / bare git URL line that isn't a version-checkable requirement). It changes nothing — no pip installs, no cm-cli, no network.

## What this adds

`comfy node deps [PACK_NAMES...]` — read-only per-pack Python dependency report. With no arguments it reports every installed pack; with names it filters case-insensitively on the pack directory name. This is the CLI half of BE-4743; a comfy-local-mcp tool wraps it.

- **`comfy_cli/command/node_deps.py`** — modeled on `outdated.py`: read-only scan, `get_renderer()` envelope output, `execute(renderer, comfy_path, pack_names)` entry point. Deliberately **not** routed through `execute_cm_cli`, so it works on a workspace that never installed ComfyUI-Manager.
- **`comfy_cli/command/pack_scan.py`** — `_iter_pack_dirs` and `_read_pyproject` **extracted** out of `outdated.py` (not copy-pasted); `outdated.py` now imports them under its existing private names, so its behavior and internal API are unchanged.
- Registered on the existing `node` typer app as `deps`. `deps-in-workflow` and `install-deps` are untouched and still resolve (verified via `--help` and a test assertion).
- **`comfy_cli/schemas/node_deps.json`** + `discovery.COMMAND_SCHEMAS["comfy node deps"]`, and the two non-fatal warning codes in `error_codes.REGISTRY`. Both were *required* — the repo has ratchet tests (`test_every_emitted_command_registers_a_schema`, `test_every_raised_code_is_registered`) that fail otherwise.

Requirements come from `requirements.txt` (same comment/option rules as `uv.py::parse_req_file`) **and** `pyproject.toml` `[project] dependencies` (via the shared registry parser), unioned with a per-line `source`. Installed versions come from one `<workspace python> -m pip list --format=json` for the **whole** report (asserted by a test), keyed on `packaging.utils.canonicalize_name` so `Pillow` matches a declared `pillow`.

## Judgment calls (all deviate slightly from, or extend, the ticket text)

- **Uniform row keys.** The ticket sketches unparseable rows as `{raw, status}`. They instead carry the same keys as every other row with `name`/`specifier`/`installed` set to `null` — a superset, so JSON consumers can index without branching on shape. Same reasoning for `status: "installed" | "not_installed"` on every pack row and for `summary` always carrying all five status keys (including `unknown`).
- **`warnings` in the payload.** The ticket calls `installed_versions_unavailable` an "envelope-level warning". It goes to stderr via `renderer.warn` *and* rides in `data.warnings[]`, because a JSON consumer reads only stdout.
- **PEP 508 direct references.** A *bare* URL/VCS line (`git+https://…`) fails `Requirement` parsing → `unparseable`, as specified. A `name @ git+https://…` line **does** parse, so it is reported as a normal requirement with an empty specifier (satisfied/missing by presence). Reporting the name we successfully parsed seemed strictly more useful than discarding it; flagging in case the ticket meant otherwise.
- **Environment markers are surfaced, not evaluated.** `pywin32; sys_platform == "win32"` reports `missing` on macOS, with the marker text in a `marker` field. Evaluating it here would be *wrong* — the relevant environment is the workspace venv's, not the CLI process's.
- **Non-PEP-440 installed version → `unknown`, not a crash.** A broken `.dist-info` makes `SpecifierSet.contains()` raise `InvalidVersion`; that is caught per-requirement so one bad dist can't abort the report. This widens `unknown` slightly beyond the ticket's "pip probe failed" (documented in the schema + module docstring).
- **`typer.Exit(code=1)` on the no-workspace path.** `renderer.error()` alone records the code but the process still exits 0 (Typer's `SystemExit(0)` short-circuits `main()`'s rc check). Followed the `comfy which` convention and raised explicitly — verified empirically that the error envelope now exits 1 while a real report exits 0.
- **Dedicated shell completer.** `pack_names` completes against *installed pack directory names*, not `node_completer`'s registry-node-id cache — a git-cloned pack's directory is its repo name, not its registry id.

## Known caveats (not fixed here — out of scope / follow-up material)

- `read_pyproject` is the shared registry parser as the ticket specifies, and that parser **strips `comfyui-frontend-package`** from `[project] dependencies`. Such a pin therefore won't appear in the report. It's a core-ComfyUI concern rather than a pack dependency, so this seems acceptable; noted in the module docstring.
- `resolve_workspace_python` prefers `VIRTUAL_ENV` / `CONDA_PREFIX` over the workspace venv — the same behavior `custom_nodes/command.py::get_installed_packages` already has, and the ticket named that call. The report echoes the resolved `python` path so the reader can always tell which interpreter was probed.
- `compiled_lock` checks `<workspace>/requirements.compiled`. `DependencyCompiler`'s `outDir` defaults to the *process* cwd on some call paths, so a lock written from a different cwd won't be found. Presence + path only, never parsed — per the ticket.

## Testing

`tests/comfy_cli/command/test_node_deps.py` (24 tests), mirroring `test_outdated.py` conventions: tmp workspace with fake `custom_nodes/<pack>/requirements.txt` + `pyproject.toml`, `pip list` mocked to a fixed JSON payload. Covers each status, the single-pip-list-invocation guarantee, `Pillow`/`pillow` canonicalization, case-insensitive pack filtering, `not_installed` rows, both-sources union + dedup, markers, pip-failure/unparseable-output/missing-interpreter degradation, unreadable `requirements.txt`, the non-PEP-440 version guard, `compiled_lock`, the JSON envelope shape, the pretty table, the no-workspace error envelope, and schema validation.

Full suite: **2798 passed, 37 skipped, 0 failed**. `ruff check` + `ruff format --diff` clean.

Also smoke-tested against a real fixture workspace: satisfied/mismatch/missing/unparseable rows in both JSON and pretty mode, `DEMO-PACK ghost` case-insensitive filtering with a `not installed` row, and exit 0 on a report vs exit 1 on the error envelope.

The diff introduces "unavailable"/"unknown"/"not installed" strings and one deny path (`not_in_workspace` + `typer.Exit(1)`), so per the falsification rule: none of these deny an existing capability. They are **report values** computed from the user's actual filesystem and venv state, and all were empirically produced against a real workspace (see smoke test above) rather than asserted. The only deny path fires when the given path genuinely is not a directory — verified that a real workspace path yields a full report at exit 0 on the same code. No existing capability is removed or routed away: `comfy node deps-in-workflow` and `comfy node install-deps` both still resolve (`--help` verified, plus a regression assertion in the test file), and `comfy outdated` is unchanged apart from importing its helpers from their new shared home.

## Out of scope (per the ticket)

Conflict resolution / auto-fix; registry lookups for uninstalled packs (follow-up ticket); parsing `requirements.compiled`.

### DIFF
--- a/comfy_cli/command/custom_nodes/command.py
+++ b/comfy_cli/command/custom_nodes/command.py
@@ -24,6 +24,7 @@ from comfy_cli.file_utils import (
     zip_files,
 )
 from comfy_cli.output import rprint as print  # context-aware: stderr in JSON mode
+from comfy_cli.output.renderer import get_renderer
 from comfy_cli.registry import (
     RegistryAPI,
     extract_node_configuration,
@@ -469,6 +470,26 @@ def node_completer(incomplete: str) -> list[str]:
         with open(tmp_path, encoding="UTF-8", errors="ignore") as cache_file:
             return [node_id for node_id in cache_file.readlines() if node_id.startswith(incomplete)]
 
+    except Exception:
+        return []
+
+
+def installed_pack_completer(incomplete: str) -> list[str]:
+    """Complete against *installed pack directory names*.
+
+    Deliberately not ``node_completer``: that one serves the registry node-id
+    cache, and a git-cloned pack's directory name is its repo name, not its
+    registry id. ``comfy node deps`` matches on directory name.
+    """
+    try:
+        from comfy_cli.command.pack_scan import iter_pack_dirs
+
+        workspace = workspace_manager.workspace_path
+        if not workspace:
+            return []
+        return [
+            p.name for p in iter_pack_dirs(pathlib.Path(workspace) / "custom_nodes") if p.name.startswith(incomplete)
+        ]
     except Exception:
         return []
 
@@ -965,6 +986,28 @@ def deps_in_workflow(
         channel,
         mode=mode,
     )
+
+
+@app.command(
+    "deps",
+    help="Report each pack's declared Python requirements vs the versions installed in the workspace venv (read-only).",
+)
+@tracking.track_command("node")
+def deps(
+    pack_names: Annotated[
+        list[str] | None,
+        typer.Argument(
+            show_default=False,
+            help="Pack directory names to report on (case-insensitive). Omit to report every installed pack.",
+            autocompletion=installed_pack_completer,
+        ),
+    ] = None,
+):
+    # Native + read-only on purpose: no cm-cli, no pip install, no network, so
+    # it works on a workspace that never installed ComfyUI-Manager.
+    from comfy_cli.command import node_deps
+
+    node_deps.execute(get_renderer(), workspace_manager.workspace_path, pack_names)
 
 
 def validate_node_for_publishing():
--- a/comfy_cli/command/node_deps.py
+++ b/comfy_cli/command/node_deps.py
@@ -1,0 +1,422 @@
+"""``comfy node deps`` — read-only per-pack Python dependency report.
+
+For every installed custom node pack (or just the ones named on the command
+line), reports the pack's *declared* Python requirements — from its
+``requirements.txt`` and/or its ``pyproject.toml`` ``[project] dependencies``
+— alongside the version actually installed in the workspace venv and a
+computed status per requirement.
+
+Read-only by construction: it never installs anything, never shells out to
+``cm-cli``, and never touches the network. The only subprocess it runs is a
+single ``<workspace python> -m pip list --format=json`` for the whole report
+(not once per pack).
+
+Statuses:
+
+- ``satisfied``   — installed, and the declared specifier accepts that version
+  (a bare name with no specifier is satisfied by any installed version);
+- ``mismatch``    — installed, but the specifier rejects that version (both
+  versions are reported);
+- ``missing``     — not present in the workspace venv at all;
+- ``unparseable`` — a pip option (``--extra-index-url``, ``-r other.txt``), a
+  bare VCS/URL line, or anything else :class:`packaging.requirements.Requirement`
+  rejects. Kept in the report rather than dropped silently;
+- ``unknown``     — the comparison could not be made: either the ``pip list``
+  probe failed (accompanied by an ``installed_versions_unavailable`` warning on
+  the report) or the installed dist records a version that isn't PEP 440.
+
+Known caveat: pyproject dependencies are read through the shared registry
+parser (:func:`comfy_cli.command.pack_scan.read_pyproject`), which drops the
+``comfyui-frontend-package`` pin from ``[project] dependencies``. Such a pin is
+a core-ComfyUI concern rather than a pack dependency, so it will not appear in
+this report even though it is declared.
+"""
+
+from __future__ import annotations
+
+import json
+import os
+import re
+import subprocess
+from pathlib import Path
+from typing import Any
+
+from comfy_cli.command.pack_scan import iter_pack_dirs, read_pyproject
+
+PIP_LIST_TIMEOUT_SECONDS = 60
+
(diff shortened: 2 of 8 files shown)


bigcat88 (approved):
Approving. Read-only claim holds, and I exercised every status against a purpose-built fixture workspace rather than trusting the table.

**Registry read-only trap avoided — the thing I checked first.** `GET /nodes/{id}/install` increments install counters and fires a Mixpanel event, so a read-only report hitting it would silently inflate download stats for every pack a user inspects. `node_deps.py` has no `requests` / `urllib` / `RegistryAPI` surface at all; the only subprocess is a single `<workspace python> -m pip list --format=json` for the whole report, bounded by `PIP_LIST_TIMEOUT_SECONDS = 60`. One probe, not one per pack. Correct on both counts.

**Every status classifies right.** Built a workspace with six packs:

| pack | declared | result |
|---|---|---|
| `pack_ok` | `requests`, `rich>=10.0` | `satisfied` ×2 (bare name satisfied by any version) |
| `pack_mismatch` | `rich==0.0.1` | `mismatch`, installed `14.1.0` |
| `pack_missing` | `definitely-not-installed-xyz>=1.0` | `missing`, installed `null` |
| `pack_weird` | `--extra-index-url …`, `-r base.txt`, `git+https://…`, `https://….whl`, `typer  # trailing` | `unparseable` ×4 + `satisfied` for `typer` |
| `pack_pyproj` | pyproject `["rich>=1.0", "nonexistent-pkg-abc"]` | `satisfied` + `missing` |
| `pack_empty` | empty file | empty list, no crash |

Comment handling is right in all three shapes — leading `#`, indented `#`, and a trailing `# comment` after `typer` (which still resolved to `typer 0.21.1` rather than being swallowed). Distinguishing a pip *option* line from a requirement instead of splitting on whitespace is the detail that usually gets this wrong, and it's handled.

**Envelope contract is clean.** The shared registry pyproject parser emits `Warning: License should be in one of these two formats…` for packs whose `pyproject.toml` lacks a conforming license — I checked whether that leaks into `--json`, and it does not: stdout is exactly **one line** and parses as JSON, with the warning on stderr. That's the correct split and worth having verified given how easy it'd be to get wrong.

**Unknown pack name is handled deliberately, not accidentally.** `comfy node deps no_such_pack` returns a row with `status: "not_installed"`, `path: null` rather than erroring — which is the better answer for an agent consumer (it distinguishes "pack absent" from "command failed"). Good call.

**Unreadable file degrades correctly.** `chmod 000` on a `requirements.txt` yields `ok: true` with a structured `pack_read_error` warning naming the path, and that pack reports zero requirements instead of taking the whole report down. Right shape: one bad pack shouldn't cost you the other five.

Full suite: **2870 passed**, 7 skipped, 1 failed — `test_non_fast_deps_uses_global_python`, the sandbox artifact present on plain `main` here too. `ruff check` clean. 24 new tests.

Leaving `pos`-style loose typing aside, two non-blocking notes:

- The completer choice (directory name, not `node_completer`'s registry ids) is right and the comment explaining why is worth keeping — those two namespaces diverge constantly and a wrong completer here would be quietly useless.
- The pyproject-via-registry-parser caveat you flagged is real but benign as shipped, since the noise lands on stderr. If it ever gets annoying, the fix is a quiet mode on `extract_node_configuration` rather than anything in this module.

PR #614, opened 2026-07-28
fix(output): strip ANSI/control sequences from server-supplied text at the pretty renderer boundary (BE-4794)

mattmillerai:
## ELI-5

When something goes wrong, the CLI prints the error message it got back from the ComfyUI server. If that server is a stranger's machine (`comfy run --host/--port` can point anywhere), it gets to choose the text — and terminals treat some text as *commands*, not characters. A message like `job <ESC>[2Jevil running` doesn't say "clear my screen", it *clears your screen*. It can also rewrite your window title, or repaint lines that already scrolled past so the CLI appears to have said something it never said.

This PR adds one small function that removes those invisible command characters, and calls it in the one place where human-facing text becomes terminal output. Machine-readable JSON output is untouched — it already escapes them safely, and stripping there would corrupt data that agents parse.

## What changed

- **New `comfy_cli/output/sanitize.py`** — strips C0/C1 control characters and ANSI escape sequences (CSI, OSC, DCS, and the shorter two- and three-character forms). Tab and newline survive (legitimate layout in multi-line messages); carriage return does not (it lets a message repaint an already-printed line). Rich *markup* is deliberately left alone — call sites use `[yellow]…[/yellow]` for styling and escape untrusted text with `rich.markup.escape`.
- **`error_panel`** (the only pretty path `Renderer.error` takes) now sanitizes `code`, `message`, `hint`, and the stringified `details` keys and values.
- **`Renderer.warn` / `info` / `success`** sanitize their `message` and `hint`.
- **The JSON/NDJSON envelope paths are untouched.** `json.dumps` already encodes ESC as a `\u` escape, so machine output was never exposed, and sanitizing there would silently mutate the payload agents read.

Why here and not at the call sites: the pattern is repo-wide (every `renderer.error(...)` interpolates `{e}` or a server-chosen id the same way), so fixing individual commands would leave the identical hazard on the adjacent lines and give a false sense of coverage. One boundary, applied once.

This is the same class of defense the repo already applies to untrusted download filenames in `_sanitize_ext` (`comfy_cli/command/transfer.py`) — that one guards the on-disk name, this one guards the rendered line.

## Verification

- `pytest tests/ --ignore=tests/e2e` — **2813 passed, 18 skipped**.
- `ruff check` and `ruff format --check` (pinned v0.15.15, matching `.pre-commit-config.yaml`) clean on every changed file. The 15 pre-existing `UP038` findings elsewhere in the repo come from a newer ruff and are untouched by this PR.
- New tests: `tests/comfy_cli/output/test_sanitize.py` (28 cases across CSI/OSC/DCS/APC, 8-bit C1 forms, unterminated sequences, idempotence, and the text that must survive), plus panel and renderer boundary tests. Both boundary tests are genuine regression tests — I confirmed `rich.print` passes `\x1b` through unchanged, so they fail without this change.
- The JSON side is asserted two ways: no raw ESC on the wire, and the parsed envelope values round-trip byte-for-byte as they went in.
- Not ReDoS-prone: the regex has no nested quantifiers, and 400 KB adversarial inputs (unterminated OSC runs, 200k-parameter CSI, ESC-then-long-tail) all sanitize in under 5 ms.

## Judgment calls

- **Extended past the ticket's letter to `Renderer.info` and `success`.** The ticket named `error` and `warn`. Sanitizing `warn` while leaving the adjacent `info` unsanitized would reproduce the exact "false sense of coverage" the ticket argues against, and both take a plain string straight to the terminal. Zero call sites today, so no behavior risk.
- **`Renderer.print` is deliberately NOT sanitized.** It is the generic passthrough and accepts arbitrary Rich renderables, not just strings; sanitizing there would mean stringifying renderables. Callers with untrusted text should use the semantic helpers above.
- **`which_panel` / `auth_list_table` / `discover_panel` are deliberately NOT sanitized.** They render locally-sourced data (workspace paths, the local auth store, the CLI's own command help), not server-supplied text. Adding them would widen the diff without closing a reachable path.
- **The helpers coerce with `str()` before sanitizing** (`sanitize_value`, not `sanitize`). Their params are annotated `str`, but `warn`/`success` previously interpolated via an f-string, which accepted anything. Keeping the coercion means a loose caller gets the old behavior instead of a new `TypeError`; there is a test for it.
- **An unterminated OSC/DCS sequence consumes the rest of the string.** That is what a terminal itself does with it, so dropping the tail is the faithful reading rather than avoidable data loss. Only reachable via a real ESC byte, which never appears in legitimate message text.
- **Homoglyph and bidi-override spoofing (e.g. U+202E) are out of scope**, and noted as such in the module docstring. They are a different problem with a different fix; the ticket scoped this to C0/C1 and ANSI.

## Capability-denial check

The change removes ANSI passthrough from four rendering helpers, so I checked empirically that nothing legitimately relies on it rather than assuming: `error_panel` has exactly one caller (`Renderer.error`); `warn`/`info`/`success` have exactly one call site between them (`comfy_cli/command/outdated.py:512`, which already wraps its input in `rich.markup.escape`); all 242 `.error(...)` call sites pass f-strings or `str(e)`; and no `.py` file in `comfy_cli/` contains a raw `\x1b`/`\033`/`\e[` literal. No product path emits intentional ANSI through these helpers, so nothing is denied.

Closes BE-4794.

### DIFF
--- a/comfy_cli/output/panels.py
+++ b/comfy_cli/output/panels.py
@@ -18,6 +18,8 @@ from rich.panel import Panel
 from rich.table import Table
 from rich.text import Text
 
+from comfy_cli.output.sanitize import sanitize, sanitize_optional, sanitize_value
+
 _TARGET_DIM_CHAR = "·"
 
 
@@ -54,7 +56,15 @@ def error_panel(
         │ details:                                                      │
         │   key=value                                                   │
         ╰───────────────────────────────────────────────────────────────╯
+
+    ``code``, ``message``, ``hint`` and the stringified ``details`` are all
+    sanitized here: any of them can carry server-supplied text, and this panel
+    is the pretty-mode boundary where it would otherwise reach the terminal as
+    live escape sequences (see ``comfy_cli.output.sanitize``).
     """
+    code = sanitize(code)
+    message = sanitize(message)
+    hint = sanitize_optional(hint)
     body: list[Any] = [Text(message, style="white")]
     if hint:
         body.append(Text(""))
@@ -66,7 +76,7 @@ def error_panel(
         for k, v in details.items():
             if v is None:
                 continue
-            rows.append((str(k), str(v)))
+            rows.append((sanitize_value(k), sanitize_value(v)))
         if rows:
             body.append(_kv_table(rows, label_style="dim", value_style="dim"))
     return Panel(
--- a/comfy_cli/output/panels.py
+++ b/comfy_cli/output/panels.py
@@ -18,13 +18,17 @@ from rich.panel import Panel
 from rich.table import Table
 from rich.text import Text
 
-from comfy_cli.output.sanitize import sanitize, sanitize_optional, sanitize_value
+from comfy_cli.output.sanitize import sanitize_markup, sanitize_optional, sanitize_value
 
 _TARGET_DIM_CHAR = "·"
 
 
 def _kv_table(rows: Sequence[tuple[str, str]], *, label_style: str = "bold cyan", value_style: str = "") -> Table:
-    """Two-column borderless table used inside panels to align label/value rows."""
+    """Two-column borderless table used inside panels to align label/value rows.
+
+    ``Table.add_row`` parses Rich markup in a ``str`` cell, so any caller
+    passing untrusted text must run it through ``sanitize_markup`` first.
+    """
     tbl = Table.grid(padding=(0, 2), expand=False)
     tbl.add_column(justify="right", style=label_style, no_wrap=True)
     tbl.add_column(style=value_style, overflow="fold")
@@ -61,9 +65,14 @@ def error_panel(
     sanitized here: any of them can carry server-supplied text, and this panel
     is the pretty-mode boundary where it would otherwise reach the terminal as
     live escape sequences (see ``comfy_cli.output.sanitize``).
+
+    ``code``/``message``/``hint`` are rendered through ``Text``, which never
+    parses markup, so stripping the control bytes is enough for them. The
+    ``details`` rows go to ``Table.add_row``, which *does* parse markup, so
+    those additionally need ``sanitize_markup``.
     """
-    code = sanitize(code)
-    message = sanitize(message)
+    code = sanitize_value(code)
+    message = sanitize_value(message)
     hint = sanitize_optional(hint)
     body: list[Any] = [Text(message, style="white")]
     if hint:
@@ -76,7 +85,7 @@ def error_panel(
         for k, v in details.items():
             if v is None:
                 continue
-            rows.append((sanitize_value(k), sanitize_value(v)))
+            rows.append((sanitize_markup(k), sanitize_markup(v)))
         if rows:
             body.append(_kv_table(rows, label_style="dim", value_style="dim"))
     return Panel(
--- a/comfy_cli/output/renderer.py
+++ b/comfy_cli/output/renderer.py
@@ -23,6 +23,7 @@ from typing import Any, TextIO
 from rich.console import Console
 
 from comfy_cli.caller import Caller, detect_caller
+from comfy_cli.output.sanitize import sanitize_value
 
 # Machine-output contract versions, surfaced in every envelope/event line and
 # in `comfy discover` (output_contract). Bump rule: additive optional fields =
@@ -195,18 +196,28 @@ class Renderer:
 
     # ----- semantic helpers -----
 
+    # These four take a plain string and hand it to the terminal, so they are a
+    # pretty-mode boundary in the same sense ``error_panel`` is: anything the
+    # message carries — including a remote server's text — is sanitized before
+    # it can be acted on as an escape sequence. Rich *markup* is deliberately
+    # left intact; call sites use it for styling (and escape untrusted text with
+    # ``rich.markup.escape``). ``print`` is not sanitized: it is the generic
+    # passthrough and takes arbitrary Rich renderables, not just strings.
+    # ``sanitize_value`` (not ``sanitize``) so a caller that ignores the ``str``
+    # annotation still gets the ``str()`` coercion the f-strings used to do.
+
     def info(self, message: str, *, hint: str | None = None) -> None:
-        self.print(message)
+        self.print(sanitize_value(message))
         if hint:
-            self.print(f"[yellow]Hint:[/yellow] {hint}")
+            self.print(f"[yellow]Hint:[/yellow] {sanitize_value(hint)}")
 
     def warn(self, message: str, *, hint: str | None = None) -> None:
-        self.print(f"[yellow]{message}[/yellow]")
+        self.print(f"[yellow]{sanitize_value(message)}[/yellow]")
         if hint:
-            self.print(f"[yellow]Hint:[/yellow] {hint}")
+            self.print(f"[yellow]Hint:[/yellow] {sanitize_value(hint)}")
 
     def success(self, message: str) -> None:
-        self.print(f"[bold green]{message}[/bold green]")
+        self.print(f"[bold green]{sanitize_value(message)}[/bold green]")
 
     # ----- structured output -----
 
--- a/comfy_cli/output/renderer.py
+++ b/comfy_cli/output/renderer.py
@@ -23,7 +23,7 @@ from typing import Any, TextIO
 from rich.console import Console
 
 from comfy_cli.caller import Caller, detect_caller
-from comfy_cli.output.sanitize import sanitize_value
+from comfy_cli.output.sanitize import sanitize_markup
 
 # Machine-output contract versions, surfaced in every envelope/event line and
 # in `comfy discover` (output_contract). Bump rule: additive optional fields =
@@ -198,26 +198,32 @@ class Renderer:
 
     # These four take a plain string and hand it to the terminal, so they are a
     # pretty-mode boundary in the same sense ``error_panel`` is: anything the
-    # message carries — including a remote server's text — is sanitized before
-    # it can be acted on as an escape sequence. Rich *markup* is deliberately
-    # left intact; call sites use it for styling (and escape untrusted text with
-    # ``rich.markup.escape``). ``print`` is not sanitized: it is the generic
-    # passthrough and takes arbitrary Rich renderables, not just strings.
-    # ``sanitize_value`` (not ``sanitize``) so a caller that ignores the ``str``
-    # annotation still gets the ``str()`` coercion the f-strings used to do.
+    # message carries — including a remote server's text — is neutralized before
+    # it can be acted on as an escape sequence. The message lands inside a Rich
+    # *markup* string, so ``sanitize_markup`` also escapes the markup: Rich would
+    # otherwise turn ``[link=http://evil]x[/link]`` back into a live OSC 8
+    # sequence, and an unbalanced ``[/]`` would raise ``MarkupError`` and crash
+    # the CLI mid-print. Escaping here rather than asking every call site to
+    # remember ``rich.markup.escape`` is the whole point of having a boundary;
+    # the cost is that these four helpers no longer accept caller markup, which
+    # is fine — they style the line themselves and no call site passed any.
+    # ``print`` is deliberately not sanitized: it is the generic passthrough and
+    # takes arbitrary Rich renderables, not just strings.
+    # ``sanitize_value`` semantics (not ``sanitize``) so a caller that ignores
+    # the ``str`` annotation still gets the ``str()`` coercion f-strings did.
 
     def info(self, message: str, *, hint: str | None = None) -> None:
-        self.print(sanitize_value(message))
+        self.print(sanitize_markup(message))
         if hint:
-            self.print(f"[yellow]Hint:[/yellow] {sanitize_value(hint)}")
+            self.print(f"[yellow]Hint:[/yellow] {sanitize_markup(hint)}")
 
     def warn(self, message: str, *, hint: str | None = None) -> None:
-        self.print(f"[yellow]{sanitize_value(message)}[/yellow]")
+        self.print(f"[yellow]{sanitize_markup(message)}[/yellow]")
         if hint:
-            self.print(f"[yellow]Hint:[/yellow] {sanitize_value(hint)}")
+            self.print(f"[yellow]Hint:[/yellow] {sanitize_markup(hint)}")
 
     def success(self, message: str) -> None:
-        self.print(f"[bold green]{sanitize_value(message)}[/bold green]")
+        self.print(f"[bold green]{sanitize_markup(message)}[/bold green]")
 
     # ----- structured output -----
 
--- a/comfy_cli/output/sanitize.py
+++ b/comfy_cli/output/sanitize.py
@@ -1,0 +1,64 @@
+r"""Strip ANSI escape sequences and control characters from untrusted text.
+
+Server-supplied strings reach the terminal verbatim in pretty mode: a remote or
+shared ComfyUI instance (``comfy run --host/--port``) picks the ``prompt_id``,
+and an exception raised against that host carries whatever text the host chose.
+Rich does not save us here — ``rich.text.Text`` drops a few C0 characters but
+passes ``\x1b`` straight through, so a CSI/OSC sequence reaches the user's
+terminal and can clear the screen, rewrite the window title, or repaint earlier
+lines to spoof CLI output.
+
+This module is the single sanitizing boundary for **pretty** rendering only.
+The JSON/NDJSON envelope paths must NOT use it: ``json.dumps`` already escapes
+``\x1b`` as a ``\u`` escape, and stripping there would silently mutate
+the data agents parse.
+
+Scope: C0/C1 control characters and ANSI escape sequences (CSI, OSC, DCS, and
+the shorter two-character forms). Tab and newline are preserved — they are
+legitimate layout in multi-line messages. Carriage return is removed, since it
+lets a message overwrite a line that has already been printed. Non-control
+Unicode is left alone; homoglyph and bidi-override spoofing are a different
+problem and deliberately out of scope here.
+"""
+
+from __future__ import annotations
+
+import re
+from typing import Any
+
+# One pass over the ANSI escape forms a terminal will act on. Ordered so the
+# multi-character sequences are consumed before the lone-ESC fallback.
+_ESCAPE_SEQUENCE_RE = re.compile(
+    r"""
+      \x1b\[ [\x30-\x3f]* [\x20-\x2f]* [\x40-\x7e]?      # CSI, 7-bit (ESC [ ...)
+    | \x9b   [\x30-\x3f]* [\x20-\x2f]* [\x40-\x7e]?      # CSI, 8-bit C1 form
+    | \x1b[\]PX^_] .*? (?: \x1b\\ | \x9c | \x07 | \Z )   # OSC/DCS/SOS/PM/APC, 7-bit
+    | [\x90\x9d\x9e\x9f] .*? (?: \x1b\\ | \x9c | \x07 | \Z )  # same, 8-bit C1 form
+    | \x1b [\x20-\x2f]* [\x30-\x7e]                      # two-/three-char escapes (ESC ( B, ESC =, ...)
+    """,
+    re.VERBOSE | re.DOTALL,
+)
+
+# Whatever survives the sweep above: stray C0 (minus tab/newline), DEL, and the
+# C1 block. A lone trailing ESC and an orphaned 8-bit introducer land here.
+_CONTROL_CHAR_RE = re.compile(r"[\x00-\x08\x0b-\x1f\x7f-\x9f]")
+
+
+def sanitize(text: str) -> str:
+    """Return ``text`` with ANSI escape sequences and control characters removed.
+
+    An unterminated OSC/DCS sequence consumes the remainder of the string —
+    that is what the terminal itself would do with it, so dropping the tail is
+    the conservative reading, not data loss we could have avoided.
+    """
+    return _CONTROL_CHAR_RE.sub("", _ESCAPE_SEQUENCE_RE.sub("", text))
+
+
+def sanitize_optional(text: str | None) -> str | None:
+    """``sanitize`` for the many call sites whose value is an optional string."""
+    return None if text is None else sanitize(text)
+
+
+def sanitize_value(value: Any) -> str:
+    """Stringify ``value`` and sanitize it — for panel fields rendered via ``str()``."""
+    return sanitize(value if isinstance(value, str) else str(value))
--- a/comfy_cli/output/sanitize.py
+++ b/comfy_cli/output/sanitize.py
@@ -8,10 +8,17 @@ passes ``\x1b`` straight through, so a CSI/OSC sequence reaches the user's
 terminal and can clear the screen, rewrite the window title, or repaint earlier
 lines to spoof CLI output.
 
-This module is the single sanitizing boundary for **pretty** rendering only.
-The JSON/NDJSON envelope paths must NOT use it: ``json.dumps`` already escapes
-``\x1b`` as a ``\u`` escape, and stripping there would silently mutate
-the data agents parse.
+This module sanitizes **pretty** rendering only. The JSON/NDJSON envelope paths
+must NOT use it: ``json.dumps`` already escapes ``\x1b`` as a ``\u`` escape, and
+stripping there would silently mutate the data agents parse.
+
+``Renderer`` is not the only path to the terminal, so applying it there is not
+sufficient on its own. Command modules also call ``rich.print`` directly under
+their own ``is_pretty()`` gates, and those never route through ``Renderer``:
+they must call ``sanitize_markup`` on each server-supplied value themselves,
+as the ``comfy run`` and ``comfy models`` paths do. Automatic coverage stops at
+``Renderer``; treat a bare ``rich.print`` of remote text as unsanitized until
+you have checked it.
 
 Scope: C0/C1 control characters and ANSI escape sequences (CSI, OSC, DCS, and
 the shorter two-character forms). Tab and newline are preserved — they are
--- a/comfy_cli/output/sanitize.py
+++ b/comfy_cli/output/sanitize.py
@@ -19,12 +19,18 @@ legitimate layout in multi-line messages. Carriage return is removed, since it
 lets a message overwrite a line that has already been printed. Non-control
 Unicode is left alone; homoglyph and bidi-override spoofing are a different
 problem and deliberately out of scope here.
+
+Removing the escape *bytes* is only half the boundary: Rich manufactures new
+ones from markup it finds in a string, so text bound for a markup-interpreting
+sink also needs ``sanitize_markup`` (see its docstring).
 """
 
 from __future__ import annotations
 
 import re
 from typing import Any
+
+from rich.markup import escape as _escape_markup
 
 # One pass over the ANSI escape forms a terminal will act on. Ordered so the
 # multi-character sequences are consumed before the lone-ESC fallback.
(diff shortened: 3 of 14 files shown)


comfy_cli/output/renderer.py
```diff
@@ -195,18 +196,28 @@ def console(self) -> Console:
...
     def info(self, message: str, *, hint: str | None = None) -> None:
-        self.print(message)
+        self.print(sanitize_value(message))
         if hint:
-            self.print(f"[yellow]Hint:[/yellow] {hint}")
+            self.print(f"[yellow]Hint:[/yellow] {sanitize_value(hint)}")
 
     def warn(self, message: str, *, hint: str | None = None) -> None:
-        self.print(f"[yellow]{message}[/yellow]")
+        self.print(f"[yellow]{sanitize_value(message)}[/yellow]")
```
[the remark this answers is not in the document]
mattmillerai: Fixed in f469172 — confirmed both halves reproduce, so this was a real bypass.

Empirically, before the fix, `renderer.warn("[link=https://attacker.example]click[/link]")` emitted `\x1b]8;id=...;https://attacker.example\x1b\\` — a live OSC 8 hyperlink manufactured by Rich *after* the sanitizer ran. And `[/]` raised `rich.errors.MarkupError`, killing the CLI mid-print of an info line.

Rather than escaping per call site, I moved it into the boundary: new `sanitize_markup()` (= `sanitize_value` + `rich.markup.escape`) is now used by `info`/`warn`/`success` and their hints. Asking every call site to remember `rich.markup.escape` is exactly the discipline a boundary is supposed to replace — and in fact the one real call site (`outdated.execute`) was doing it manually, so I removed that now-redundant `escape()` since double-escaping renders a visible stray backslash.

The tradeoff: these four helpers no longer accept caller markup. Checked that this denies nothing — an exhaustive grep of `.info|.warn|.success(` across `comfy_cli/` returns 7 hits, 6 of which are stdlib `logging`, and the one renderer call passed pre-escaped text. Callers who do want markup still have `renderer.print()`, the documented generic passthrough that takes arbitrary Rich renderables. The helpers style their own line anyway.

Values that go to `rich.text.Text` (`code`/`message`/`hint` in `error_panel`) deliberately keep plain `sanitize_*`: `Text` never parses markup, so escaping there would print the backslashes.

12 new renderer tests (link-injection + unbalanced-markup × 6 entry points), all verified to fail without the source change. Also checked realistic messages (`node list[0] failed`, `[2/5] downloading model`, `C:\Users\me\path`, URLs) render byte-identically — `rich.markup.escape` is a no-op on markup-free text.

comfy_cli/output/panels.py
```diff
@@ -66,7 +76,7 @@ def error_panel(
         for k, v in details.items():
             if v is None:
                 continue
-            rows.append((str(k), str(v)))
+            rows.append((sanitize_value(k), sanitize_value(v)))
```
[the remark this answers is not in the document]
mattmillerai: Fixed in f469172. Confirmed `Table.add_row` parses markup in a `str` cell exactly as you describe — rendering an `error_panel` with `details={"prompt_id": "[link=https://attacker.example]x[/link]"}` through a force-terminal Console emitted `\x1b]8;...`, and a `[/]` detail raised `MarkupError`.

The `details` rows now go through the new `sanitize_markup()` (sanitize + `rich.markup.escape`). `code`/`message`/`hint` intentionally stay on plain `sanitize_*` — they render via `Text`, which never parses markup, so escaping them would surface the backslashes on screen. Documented that split in the `error_panel` docstring and added a note on `_kv_table` that its `str` cells are a markup-parsing sink.

Two new panel tests cover it, both verified failing before the fix.

comfy_cli/output/sanitize.py
```diff
@@ -0,0 +1,64 @@
...
+from typing import Any
+
+# One pass over the ANSI escape forms a terminal will act on. Ordered so the
+# multi-character sequences are consumed before the lone-ESC fallback.
+_ESCAPE_SEQUENCE_RE = re.compile(
+    r"""
+      \x1b\[ [\x30-\x3f]* [\x20-\x2f]* [\x40-\x7e]?      # CSI, 7-bit (ESC [ ...)
+    | \x9b   [\x30-\x3f]* [\x20-\x2f]* [\x40-\x7e]?      # CSI, 8-bit C1 form
+    | \x1b[\]PX^_] .*? (?: \x1b\\ | \x9c | \x07 | \Z )   # OSC/DCS/SOS/PM/APC, 7-bit
+    | [\x90\x9d\x9e\x9f] .*? (?: \x1b\\ | \x9c | \x07 | \Z )  # same, 8-bit C1 form
```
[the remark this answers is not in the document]
mattmillerai: Fixed in f469172 — good catch, `\x98` (SOS) added to the 8-bit C1 string-introducer class, which is now `[\x90\x98\x9d\x9e\x9f]` (DCS/SOS/OSC/PM/APC).

You had the mechanism exactly right: `\x98payload\x9c` was falling through to `_CONTROL_CHAR_RE`, which removed the two introducer bytes and left `payload` as visible garbage, diverging from the fully-consumed 7-bit `\x1bX` form. Added parametrized cases for both forms so they are pinned as equivalent.

comfy_cli/output/panels.py
```diff
@@ -54,7 +56,15 @@ def error_panel(
         │ details:                                                      │
         │   key=value                                                   │
         ╰───────────────────────────────────────────────────────────────╯
+
+    ``code``, ``message``, ``hint`` and the stringified ``details`` are all
+    sanitized here: any of them can carry server-supplied text, and this panel
+    is the pretty-mode boundary where it would otherwise reach the terminal as
+    live escape sequences (see ``comfy_cli.output.sanitize``).
     """
+    code = sanitize(code)
```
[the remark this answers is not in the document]
mattmillerai: Fixed in f469172 — `code` and `message` now use `sanitize_value()`.

Extended it slightly: `hint` carries the identical hazard via `sanitize_optional()`, so that helper now routes non-`None` values through `sanitize_value` too. It keeps its `None` passthrough contract, so nothing else changes. That removes the last path where a caller ignoring the `str` annotation reaches `re.sub` with a non-string. Added tests for `sanitize_optional(42)` and for `error_panel(code=42, message=7, hint=13)`.

mattmillerai:
Addressed all 4 findings from the Cursor panel (2 High, 2 Low) in f469172 + 3ce16d1. All four threads replied to and resolved.

**The High findings were real, and they shared one root cause.** The PR stripped the escape *bytes*, but Rich manufactures new ones from markup it finds in the string afterwards — so the sanitizer was bypassable. Reproduced both halves before fixing:

- `renderer.warn("[link=https://attacker.example]click[/link]")` emitted `\x1b]8;id=...;https://attacker.example\x1b\\` — a live OSC 8 hyperlink, created *after* sanitizing. Same via `error_panel(details=...)`, since `Table.add_row` parses markup in a `str` cell.
- An unbalanced `[/]` raised `rich.errors.MarkupError`, crashing the CLI while it was merely printing an info line.

**Fix:** new `sanitize_markup()` (`sanitize_value` + `rich.markup.escape`), applied at every sink that actually parses markup — `Renderer.info/warn/success` and their hints, plus the `details` rows in `error_panel`. Values that go to `rich.text.Text` (`code`/`message`/`hint`) keep plain `sanitize_*`: `Text` never parses markup, so escaping there would print the backslashes. That split is documented on both functions.

I moved the escaping into the boundary rather than the call sites deliberately. The original design asked every call site to remember `rich.markup.escape`, which is the exact discipline a boundary exists to replace — and `outdated.execute` was the one place doing it manually. I removed that now-redundant call, since double-escaping renders a visible stray backslash.

**The tradeoff, checked rather than assumed:** these helpers no longer accept caller markup. An exhaustive grep of `.info|.warn|.success(` across `comfy_cli/` returns 7 hits — 6 are stdlib `logging`, and the 7th passed pre-escaped text. Nothing is denied: callers who want markup still have `renderer.print()`, the documented generic passthrough for arbitrary Rich renderables, and these four style their own line anyway.

Also verified no cosmetic regression: `rich.markup.escape` is a no-op on markup-free text, so realistic messages render byte-identically — `node list[0] failed`, `shape [1,2,3] mismatch`, `[2/5] downloading model`, `C:\Users\me\path`, URLs.

**Low findings:** added SOS (`\x98`) to the 8-bit C1 string-introducer class, so `\x98payload\x9c` is consumed as a unit instead of leaving `payload` as visible garbage — matching the 7-bit `\x1bX` form. And `error_panel` `code`/`message` now use `sanitize_value`; I extended the same to `hint` via `sanitize_optional`, which carried the identical `TypeError` hazard.

**Tests:** 15 new cases, each verified to fail with the source fix stashed (not vacuous). Console width is pinned in the markup tests since `rich.print` gives no way to pass `width` through and the error-panel case fit 80 columns with only ~10 to spare. Full suite green locally (2835 passed), `ruff check` + `ruff format` clean against the CI-pinned 0.15.15.

No deferrals — everything raised was fixable in scope.

comfy_cli/output/sanitize.py
```diff
@@ -0,0 +1,92 @@
...
+
+Server-supplied strings reach the terminal verbatim in pretty mode: a remote or
+shared ComfyUI instance (``comfy run --host/--port``) picks the ``prompt_id``,
+and an exception raised against that host carries whatever text the host chose.
+Rich does not save us here — ``rich.text.Text`` drops a few C0 characters but
+passes ``\x1b`` straight through, so a CSI/OSC sequence reaches the user's
+terminal and can clear the screen, rewrite the window title, or repaint earlier
+lines to spoof CLI output.
+
+This module is the single sanitizing boundary for **pretty** rendering only.
```
bigcat88: This says "the single sanitizing boundary for **pretty** rendering only", but it is not the only path to the terminal — it is only the only path *through `Renderer`*. Command modules call `rich.print`/`pprint` directly under their own `is_pretty()` gates, and those never reach this module.

Proven against this branch with a hostile server returning a `text/plain` body containing raw escape bytes:

```
raw ESC bytes: 4 | CSI-2J: True | OSC-title: True
```

The leak is `comfy_cli/command/run/execution.py:203`, which prints the same `body_text` two lines above the `renderer.error(...)` this module does cover. Please either widen coverage to those `pprint` sites or reword this docstring to say it is the boundary for `Renderer`-routed text specifically — as written it will convince the next reader the whole surface is covered.

comfy_cli/output/renderer.py
```diff
@@ -195,18 +196,34 @@ def console(self) -> Console:
...
+    # message carries — including a remote server's text — is neutralized before
+    # it can be acted on as an escape sequence. The message lands inside a Rich
+    # *markup* string, so ``sanitize_markup`` also escapes the markup: Rich would
+    # otherwise turn ``[link=http://evil]x[/link]`` back into a live OSC 8
+    # sequence, and an unbalanced ``[/]`` would raise ``MarkupError`` and crash
+    # the CLI mid-print. Escaping here rather than asking every call site to
+    # remember ``rich.markup.escape`` is the whole point of having a boundary;
+    # the cost is that these four helpers no longer accept caller markup, which
+    # is fine — they style the line themselves and no call site passed any.
+    # ``print`` is deliberately not sanitized: it is the generic passthrough and
```
bigcat88, line 210: Not `Renderer.print` itself — agreed that leaving the generic passthrough alone is correct. The problem is the sites that bypass `Renderer` entirely.

`command/run/execution.py:203` does:

```python
pprint(f"[bold red]Error running workflow (HTTP {e.status})\n{body_text}[/bold red]")
```

with `body_text` straight off the wire, under its own `is_pretty()` gate. Two defects, both reproduced live on this branch: raw ANSI reaches the terminal (screen clear + window-title rewrite), and a body of `boom [/] unbalanced` raises an unhandled `MarkupError` that crashes the CLI. Both pre-exist on `main`, so this PR does not regress anything — but they are squarely in this PR's threat model.

`sanitize_markup(body_text)` there fixes both; I tested it. Same shape at `run/__init__.py:962`, `run/preflight.py:111`, and `models/search.py:606-620`.

bigcat88 (changes requested):
The sanitizer itself is excellent and I could not break it. What I can break is the **claim that it closes the hole** — the headline attack from your own ELI-5 (`comfy run --host/--port` pointed at a stranger's machine) still clears my screen and sets my window title on this branch. Requesting changes so BE-4794 doesn't close on a threat that still works, not because anything here is wrong.

**What I verified works.** `sanitize()` neutralized all 13 vectors I threw at it — CSI 2J, OSC 0 title-set, OSC 8 hyperlink, CR repaint, 8-bit C1 introducer, DCS, APC, unterminated OSC, lone trailing ESC, two-/three-char escapes, NUL/BEL/DEL, raw C1 block — with zero leaks, idempotent, and legitimate text (newlines, tabs, unicode, literal `[brackets]`) untouched. ReDoS claim holds: 400 KB adversarial inputs in 0.4–6.8 ms. Your non-vacuity claim is real — I confirmed `Console.print` passes the CSI 2J sequence through **intact** for both `str` and `Text`, so the boundary tests are genuine regression tests. All four helpers strip end-to-end through the real `Renderer`, while JSON mode emits the `` escape and round-trips the payload byte-for-byte. Suite: 2907 passed, 1 failure that is the sandbox's `test_non_fast_deps_uses_global_python`, identical on `main`. ruff clean.

**The gap.** I stood up a hostile ComfyUI-shaped server (answers `/history` with 200, returns `text/plain` with raw escape bytes on `/prompt`) and ran the real CLI against it in pretty mode on **this branch**:

```
$ COMFY_OUTPUT=pretty comfy run --workflow wf.json --host 127.0.0.1 --port 8793
raw ESC bytes: 4 | CSI-2J: True | OSC-title: True

Error running workflow (HTTP 400)
prompt <clear-screen><set-title:PWNED><red>FAKE: install succeeded
╭─ error · client_error ────────────────────────────────╮   <- this part IS sanitized
```

The panel is clean. The line above it is not. `comfy_cli/command/run/execution.py:203` prints the server's body via `pprint` under its own `is_pretty()` gate — two lines above the `renderer.error(...)` your boundary does cover, interpolating the *same* `body_text`. That is exactly the "identical hazard on the adjacent lines / false sense of coverage" your body argues the call-site approach would produce; the boundary approach lands there too, because `Renderer` is not the only path to the terminal.

Second, independent defect at the same line: a body of `boom [/] unbalanced` crashes the CLI with an unhandled `rich.errors.MarkupError` traceback — the precise failure your `sanitize_markup` docstring predicts. A remote server can hard-crash the client.

**Both are pre-existing on `main`, not introduced here** — I checked, `main` crashes identically. So this PR makes nothing worse. But it is scoped as *the* fix for server-supplied ANSI, and merging it as-is means the ticket closes with `comfy run` against a hostile host still executing escape sequences.

**The fix is one line, using your own helper.** I applied it locally and re-ran the same attack:

```python
pprint(f"[bold red]Error running workflow (HTTP {e.status})\n{sanitize_markup(body_text)}[/bold red]")
```

```
raw ESC bytes: 0 | CSI-2J: False | MarkupError: False

Error running workflow (HTTP 400)
boom  [/] unbalanced          <- renders literally, no crash
```

**Sibling sites worth sweeping while you're in there** (same shape — server-supplied text into a pretty-mode `pprint`/`rprint`, never touching `Renderer`):

- `command/run/__init__.py:962` — `w['message']` from cloud warnings
- `command/run/preflight.py:111` — `w['field']` / `w['message']`, field names derive from `object_info`
- `command/models/search.py:606-620` — `row['name']` / `type` / `tags` / `source_url` from the cloud asset catalog
- `command/run/__init__.py:471` — the interpolated exception on lost connection

`execution.py:193` and `:544` are already safe — `json.dumps` escapes the escape byte — worth a comment saying so, so nobody "fixes" them later.

Everything else in the PR I agree with, including the judgment calls: extending to `info`/`success` is right, leaving `Renderer.print` alone is right, `sanitize_value`'s `str()` coercion over a new `TypeError` is right, and consuming the tail of an unterminated OSC is the faithful reading. Re-request me once the `pprint` paths are covered and I will re-run the hostile-server suite against it.

mattmillerai:
Fixed in 98665cc — you were right, and the reword alone would not have been enough.

I widened coverage rather than just narrowing the claim, then did both. `sanitize_markup` now runs at the direct `rich.print` sites across the `comfy run` and `comfy models` surfaces:

- `execution.py` — HTTP error body (your repro), the `execution_error` payload, the submit `URLError` reason, and the `--verbose` `log_node` line (leaves escaped before assembly, since that string builds its own markup)
- `run/__init__.py` — lost-connection exception, local output paths, cloud output URLs and warnings
- `preflight.py` — validation warning `field`/`message`
- `watcher.py` — output URLs off the state file
- `models/search.py` — `show` fields plus the three `Table.add_row` sites (`add_row` parses markup in a `str` cell, the same sink as the error panel's `_kv_table`)

**One correction to your sweep list:** `execution.py:193` and `:544` are *not* already safe. `json.dumps` escapes the ESC byte as a `\u` escape, so the ANSI half is covered as you said — but it does not escape `[`. A node-error message of `[/red]` survives the JSON encoding intact and still raises `MarkupError`:

```
rich.errors.MarkupError: closing tag '[/red]' at position 92 doesn't match any open tag
```

So a hostile server crashes the client through the `node_errors` path too. Both `json.dumps` sites are sanitized now. Worth noting the rendered output is unchanged: Rich consumes the escaping backslash, so the blob on screen is byte-identical to the original JSON (asserted in a test).

The docstring is reworded too — it now says automatic coverage stops at `Renderer` and a bare `rich.print` of remote text must sanitize explicitly, instead of claiming to be "the single sanitizing boundary".

15 regression tests in `tests/comfy_cli/command/test_pretty_print_sanitize.py` drive the real call sites in pretty mode against a force-terminal console, asserting no OSC 8 / OSC 0 / CSI 2J and no crash on `[/]`. Every one was verified to fail without its fix. Full suite: 2850 passed, 37 skipped; ruff clean on the touched files.

Two caveats I'd rather state than have you find:

1. The cloud-warning and local-output loops in `execute_cloud` / `execute` are fixed but not unit-tested — driving those end-to-end needs the whole cloud polling path, which I judged not worth the harness. They are the same one-value `sanitize_markup(x)` shape as the tested sites.
2. I scoped this to the `comfy run` + `comfy models` surfaces, which is BE-4794's threat model. The repo has ~500 other `rich.print` sites (`install.py`, `setup.py`, `nodes.py`), carrying registry and local text rather than remote-host text. Sweeping those is a separate piece of work, not a silent omission.

Re-requesting review — the hostile-server suite should come back clean now.

comfy_cli/output/sanitize.py
mattmillerai: Reworded in 98665cc, and the coverage widened rather than only the claim narrowed.

The docstring no longer says "the single sanitizing boundary". It now states that automatic coverage stops at `Renderer`, that command modules calling `rich.print` under their own `is_pretty()` gates must call `sanitize_markup` on each server-supplied value themselves, and that a bare `rich.print` of remote text should be treated as unsanitized until checked.

`execution.py:203` and the sibling sites you listed are all sanitized now — details and one correction about the `json.dumps` sites in the PR comment.

comfy_cli/output/renderer.py
mattmillerai: Fixed in 98665cc. Reproduced both defects at `execution.py:203` before fixing — raw ANSI reaching the terminal, and `boom [/] unbalanced` raising an unhandled `MarkupError` — and applied `sanitize_markup(body_text)` exactly as you suggested.

Swept all four siblings you named plus a few more of the same shape on the `comfy run` / `comfy models` surfaces: the `execution_error` payload, the submit `URLError` reason, the `--verbose` `log_node` line, preflight warning `field`/`message`, local + cloud output paths and cloud warnings, the watcher output URLs, and the asset-catalog rows — including the three `Table.add_row` sites, since `add_row` parses markup in a `str` cell just like the `_kv_table` the error panel already covers.

One correction: `execution.py:193`/`:544` were not already safe. `json.dumps` escapes the ESC byte but not `[`, so a node-error message of `[/red]` still raised `MarkupError` and crashed the client. Both are sanitized now, and the rendered blob stays byte-identical to the original JSON since Rich consumes the escaping backslash.

Agreed on leaving `Renderer.print` alone. 15 regression tests drive the real call sites in pretty mode, each verified failing without its fix.

mattmillerai:
Follow-up to the above — my own adversarial pass found a third surface reachable by the identical attack, so I fixed it here rather than deferring it (1a5a94f).

`comfy nodes ls / show / search / types / categories / path` take the **same `--host/--port` pair as `comfy run`** and render `object_info` straight from that host into markup sinks — node ids, categories, descriptions, display names, input/output types. Verified against a hostile `object_info`:

```
comfy nodes show --host <evil>   ->  CSI-2J: True | OSC-title: True
description = "boom [/] x"       ->  rich.errors.MarkupError (unhandled)
```

Deferring that would have closed BE-4794 with the same threat still working one command over, which is exactly the objection you raised on the run path — so it seemed wrong to file it as a follow-up.

Intentional markup is preserved: the `[dim]—[/dim]` empty-outputs fallback and the `[bold]`/`[cyan]` wrappers are ours, so only the joined server values inside them are escaped. The search/description cells truncate *before* escaping so the escapes stay balanced.

5 more regression tests (20 total in the new file), each verified failing without the fix. Full suite 2855 passed, 37 skipped; ruff clean on touched files.

On the scope boundary I mentioned: I checked the rest of the repo's `rich.print` surface (`install.py`, `setup.py`, `cmdline.py`, the registry commands) and did **not** find attacker-controlled remote text flowing into those — they carry local paths, versions, user input, and manager-delegated output. So I'm not proposing a follow-up ticket for them; the reworded module docstring is what guards the next reader. If you know of a remote-text path there I missed, say so and I'll take it.

comfy_cli/output/renderer.py
mattmillerai: One more finding from merging `main` into this branch, recorded as a follow-up rather than folded in here.

`main`'s model-download path (`command/models/models.py:401` and `:704`) renders download-server error text with `rich.markup.escape()`. That covers the markup half — its inline comment says it was added to stop `MarkupError` — but it does not strip raw ANSI. And `file_utils.py:75` (`guess_status_code_reason`, the 401 branch) interpolates the server's own JSON `message` into the reason string, so escape bytes do get through:

```
body = {"message": "\x1b[2J\x1b]0;PWNED\x07denied"}   # 401
escape()          -> CSI-2J reaches terminal: True,  OSC-title: True
sanitize_markup() -> CSI-2J: False, OSC-title: False
```

Reachable via `comfy model download --url <hostile host>`. Reproduced on `main` and on this branch's merged tree.

I did **not** fix it here: `models/models.py` and `file_utils.py` are not in this PR's diff — they only appear in my tree because I merged `main` to clear a conflict. The download surface is a different command from the render boundary this PR is scoped to, its author already applied a partial mitigation, and pulling a third unrelated file into an already-large security PR would just make your re-review harder. Filed as a follow-up (swap `escape` for `sanitize_markup` at the two sites, or sanitize once at `guess_status_code_reason`), so it is tracked rather than dropped.

Flagging it explicitly in case you disagree and would rather see it in this PR — happy to pull it in.

mattmillerai:
**Heads-up on the red `build` check — it is not this PR, it is `main`.**

`build` fails on a single assertion:

```
FAILED tests/comfy_cli/output/test_error_code_registry.py::test_no_duplicate_codes
  AssertionError: Duplicate codes in registry: ['server_died']
  1 failed, 3125 passed, 36 skipped
```

`comfy_cli/error_codes.py` registers `ErrorCode("server_died", ...)` twice on `main` — once from #605 (BE-4750) and once from #604 (BE-4751). Neither PR touched the other's line, so git merged both with no textual conflict and the collision only shows up at runtime.

Evidence it is inherited, not introduced here:

- this PR's diff does not touch `error_codes.py` (`git diff origin/main...HEAD --name-only | grep error_codes` → nothing)
- `git show origin/main:comfy_cli/error_codes.py | grep -c '"server_died",'` → **2**
- `Run pytest` on `main` itself (e6965d0) is **failing** with the same assertion
- every other job here is green: `ruff_check`, `test`, `Run Tests on GPU Runners (linux)`, `Run Tests on Multiple Platforms (ubuntu/macos/windows, 3.10)`, CodeQL, Analyze, Socket, CLA

I deliberately did **not** make it green from this branch. The assertion is correct — it caught a real contract collision in the public error-code registry — and deleting or relaxing someone else's passing-by-design test to unblock my own PR would ship the breakage. Fixing it properly means collapsing the two entries into one and reconciling their descriptions/hints, which is a change to the public error-code surface and belongs in its own PR, not buried in an ANSI-sanitizing one. Filed as a separate follow-up.

So: this branch is clean on its own merits. Once the `server_died` duplicate is fixed on `main`, this check goes green here with no change to this PR.

Also note the branch now contains a merge of `main` (53d7759) taken to clear a textual conflict in `command/run/__init__.py`: `main` renamed `jobs_state.write` → `_write_state` and moved the pretty `Outputs:` block out of the `try` to after the `finally`, iterating `completed_payload["outputs"]`. I took `main`'s structure wholesale and re-applied the sanitize at the block's new home, so both intents survive. Verified on the merged tree: 3039 passed, 37 skipped locally, ruff clean on the touched files.

mattmillerai:
- [BE-5023](https://linear.app/comfyorg/issue/BE-5023) — Strip ANSI (not just Rich markup) from download-server error text in `comfy model download`
- [BE-5024](https://linear.app/comfyorg/issue/BE-5024) — Fix duplicate `server_died` error code breaking `main` (collides between #604 and #605)

bigcat88 (approved):
Fixed — approving. I re-ran the exact attack that broke this branch yesterday.

**Before (my last review), same command, same hostile server:**

```
raw ESC bytes: 4 | CSI-2J: True | OSC-title: True | MarkupError: crash
```

**Now:**

```
raw ESC bytes : 0
CSI-2J        : False
OSC title-set : False
MarkupError   : False

Error running workflow (HTTP 400)
prompt FAKE: install succeeded and [bold]markup[/bold] and [/]
```

Screen-clear gone, window-title rewrite gone, and the unbalanced `[/]` that used to raise an unhandled `MarkupError` now renders literally. The hostile server is the same rig: answers `/history` with 200, returns `text/plain` carrying raw `ESC[2J`, OSC title-set, and Rich markup.

**All four sibling sites I listed are covered**, plus one I'd missed:

- `run/execution.py:207` — `sanitize_markup(body_text)`, exactly the fix
- `run/__init__.py:473` — lost-connection exception text
- `run/__init__.py:518, 963` — cloud warnings / uploads
- `run/preflight.py:113` — both `field` and `message`
- `models/search.py:214, 215, 304` — catalog-supplied names and subfolders
- `command/nodes.py:287, 288, 362, 364` — **object_info-derived node ids, categories and display names.** Good catch; I didn't flag that one and it's server-supplied text on a surface people run constantly.

**The JSON contract is still intact**, which is the part that would have been easy to break while fixing this. stdout is pure JSON (0 non-JSON lines), the raw escape bytes are preserved *inside* the payload so agents still parse exactly what the server sent, and no raw ESC byte appears on the wire. Strip at the terminal, never mutate the data — still correct.

**On the suite: 3057 passed, 2 failed, and neither is yours.**

- `test_non_fast_deps_uses_global_python` — the usual sandbox artifact, fails on plain `main` here too.
- `test_no_duplicate_codes` — **`main` is currently broken**, independently of this PR. `error_codes.py` has `server_died` registered twice: #605 (BE-4750, "connection dropped while a foreground `--wait` job was in flight") at line 366 and #604 (BE-4751, "local ComfyUI server became unreachable") at line 376. Both merged, git auto-merged them cleanly because they sit at different offsets in the list, and the combined tree fails. I confirmed it on a clean `main` checkout at `e6965d0` with `error_codes.py` untouched — nothing to do with this branch. Flagged separately; don't let it hold this up.

`ruff check` / `ruff format --check` clean over 261 files. 18 new cases in `test_pretty_print_sanitize.py` covering the direct-`pprint` sites.

Nice turnaround — the boundary claim in the description is now actually true.

PR #617, opened 2026-07-28
fix(validate): expand dynamic-combo options so validate matches the server (BE-3777)

mattmillerai:
## ELI-5

Some ComfyUI nodes have a dropdown that *changes what else the node needs*. Pick "Seedream 5.0 pro" and the node wants a width and a height; pick "5.0 lite" and it also wants `max_images`. `comfy validate` only knew about the dropdown itself, not about the extra fields the choice pulls in — so it happily said "looks good" about a workflow that the server then refused to run. This teaches validate to open the dropdown, see which extra fields the chosen option actually requires, and check them.

## The bug

A `COMFY_DYNAMICCOMBO_V3` input (Seedream's `model`, LTXV, several partner-API nodes) declares exactly ONE input in `object_info` — the selector. The option the selector names carries its own `INPUT_TYPES` block, and both the frontend and `convert_ui_to_api` lower it into dotted keys: `model.size_preset`, `model.width`, `model.height`.

Validation never expanded that. `model.width` misses `port_by_name`, so the per-input loop `continue`d past it; and `_check_required_present` exempted the parent selector outright. Every one of these came back `valid: true` against the captured Seedream catalog before this change (measured on `origin/main`):

| mutation of the pristine Seedream 5.0 Pro template | before | after |
|---|---|---|
| `model.width` deleted | `valid: true` | `required_input_missing` |
| `model.width` = `"wide"` | `valid: true` | `shape_mismatch` |
| `model.width` = `8` (catalog min 1024) | `valid: true` | `below_min` |
| `model.size_preset` = `"(9K) nope"` | `valid: true` | `unknown_enum_value` |
| `model` = `"seedream 9.9 ultra"` | `valid: true` | `unknown_enum_value` |
| `model` deleted | `valid: true` | `required_input_missing` |
| selector flipped to `"seedream 5.0 lite"` | `valid: true` | `required_input_missing` × 2 (`max_images`, `fail_on_partial`) |

## What this changes

`_check_dynamic_combos` resolves each selector's option **exactly the way the server does** and validates that option's sub-inputs through the same `_parse_input_spec` / `Port` machinery as any top-level input — so they inherit identical presence, shape, enum-membership and range semantics rather than a parallel set of rules.

- Nested dynamic combos recurse, under the same depth cap the converter uses.
- Autogrow sub-inputs are exempt: they wire as further-dotted slot keys (`model.images.image_1`) and routinely declare `min: 0` even inside the `required` section, so absence is not a rejection. Their edges are still checked by the driver loop.
- A selector or sub-input satisfied by a link is left to the existing dangling-edge / output-index checks.
- Malformed option blocks degrade to "nothing to expand" — `object_info` is server-supplied and must never take validation down.

`comfy run`'s preflight shares the engine, so both paths gain the check. No new error codes: the four emitted (`required_input_missing`, `shape_mismatch`, `unknown_enum_value`, `below_min`/`above_max`) already exist.

## Verification against the real server

I checked the premise against ComfyUI's source rather than assuming it, because a validator that hard-errors on something the server accepts is worse than the bug it replaces.

- `DynamicCombo._expand_schema_for_dynamic` (`comfy_api/latest/_io.py`) reads the submitted selector, finds the option whose `key` **strictly equals** it, and folds *only that option's* inputs into the finalized required/optional sets, prefixed into dotted names. This confirms selector-driven expansion, strict `==` matching, dotted naming, and recursion — the four load-bearing design choices here.
- `validate_inputs` (`execution.py:886-900`) then walks those finalized names: `if x not in inputs: ... "required_input_missing"`, with **no** exemption for inputs carrying a default. Missing dotted sub-inputs are genuine `/prompt` rejections.
- The same loop supplies `invalid_input_type`, `value_smaller_than_min` / `value_bigger_than_max`, and the COMBO membership check for sub-inputs that *are* present.

**One nuance I found and deliberately did not paper over.** When the selector is absent, or names no option, the server's expansion returns early — so the selector never enters `valid_inputs`, `/prompt` **accepts** the prompt, and the node instead dies at *execution* (its inputs are silently dropped), i.e. after a paid node has been entered. Those two cases are still hard errors here (the workflow genuinely cannot run, and it's the case where the converter also silently misaligns every following widget value), but their **messages say "fails at execution", not "the server will reject this node"** — I did not want validate claiming a `/prompt` rejection that does not happen. `test_unresolvable_selector_messages_say_execution_not_prompt` pins that distinction. If you'd rather these two were warnings so `valid` is byte-for-byte `/prompt` equivalence, that's a one-line change — say the word.

Over-strictness guard: `test_converted_seedream_workflow_is_valid` runs the pristine captured Seedream 5.0 Pro UI export through the real converter and asserts it still validates clean, so the new checks can't start rejecting good workflows unnoticed.

## Scope note — where this ticket actually lives

The ticket was filed against the local MCP, whose `validate_workflow` is a one-line passthrough to `comfy validate` (its `AGENTS.md` forbids implementing anything there: "New functionality belongs in comfy-cli"). The defect is in this repo's CQL engine, so it is fixed here; the MCP picks it up with no change (it shells out to whatever `comfy` is on `PATH` and pins no version). The ticket's repo label should move to `comfy-cli`.

Also worth recording: the ticket's other two named symptoms were already fixed and this PR does **not** claim them — required-input presence-checking landed in #551, and validate's UI→API auto-conversion in #553. Dynamic-combo expansion was the remaining leg, and it is what made the reporter's repro still fail after those.

## Discovered, NOT fixed here (follow-up)

`comfy validate` emits a failure envelope with `ok: false` but `error: null` — the per-node findings live in `data.errors`. An MCP-style consumer that unwraps the envelope on `ok` therefore raises with code `unknown` and loses every structured error, so an agent learns *that* the workflow is invalid but not *which* input is wrong. That predates this change and affects all validate failures, and fixing it is a contract question (how an error envelope should carry a per-node error list), so I left it out rather than mixing it in. Worth its own ticket.

## Verification

- `pytest` — 2934 passed, 37 skipped. The 2 failures in `tests/comfy_cli/test_broken_pipe.py` are pre-existing on `origin/main` (confirmed by re-running them with this change stashed) and unrelated.
- `ruff check` and `ruff format --diff` clean under CI's pinned `ruff==0.15.15`.
- Red→green proof: with the engine change stashed, 8 of the 15 new tests fail; the other 7 are the over-strictness / no-false-positive guards and pass either way by design.
- End-to-end through the real CLI in offline mode (`--input object_info.json`): the good API workflow validates `ok: true` / exit 0, the same workflow minus `model.width` returns `required_input_missing` / exit 1, and the raw UI export converts (`converted_from_ui: true`) and validates.

### DIFF
--- a/comfy_cli/cmdline.py
+++ b/comfy_cli/cmdline.py
@@ -915,7 +915,11 @@ def run(
 
 
 @app.command(
-    help="Validate an API-format workflow without submitting. Checks class_types, input shapes, enum values, and edge wiring."
+    help=(
+        "Validate a workflow without submitting (UI exports are converted to API format first). "
+        "Checks class_types, input shapes, enum values, edge wiring, and the dotted sub-inputs a "
+        "dynamic combo's selected option requires."
+    )
 )
 @tracking.track_command()
 def validate(
--- a/comfy_cli/cql/engine.py
+++ b/comfy_cli/cql/engine.py
@@ -50,6 +50,11 @@ class Port:
     is_link: bool = False
     enum_values: list[Any] = field(default_factory=list)  # preserves the option's real type (int combos stay int)
     options: PortOptions = field(default_factory=PortOptions)
+    # The verbatim ``INPUT_TYPES`` spec this port was parsed from. Retained
+    # (by reference — no copy) because a dynamic combo's sub-input schema lives
+    # in the spec's ``options`` blocks and is NOT recoverable from the parsed
+    # fields above. Output ports carry ``None``.
+    raw_spec: Any = None
 
     @property
     def is_autogrow(self) -> bool:
@@ -57,6 +62,18 @@ class Port:
         ONE input, but the server expects autogrown slot keys —
         ``images.image0``, ``images.image1``, … one per connection."""
         return self.type.startswith("COMFY_AUTOGROW")
+
+    @property
+    def is_dynamic_combo(self) -> bool:
+        """V3 dynamic combo (e.g. ``COMFY_DYNAMICCOMBO_V3``): the schema declares
+        ONE selector input, but the option the selector names carries its own
+        ``INPUT_TYPES`` block, which the frontend — and ``convert_ui_to_api`` —
+        lower into dotted slot keys (``model.size_preset``, ``model.width``, …).
+
+        Deliberately the same test the converter applies
+        (``workflow_to_api._is_widget_input``), so validation expands exactly the
+        tree the converter lowered."""
+        return self.type.startswith("COMFY_") and "COMBO" in self.type
 
     def autogrow_slot_example(self) -> str:
         """Best-effort slot-key example for hints. The element name comes from
@@ -310,6 +327,7 @@ def _parse_inputs(raw: dict, order: list[str] | None, required: bool) -> list[Po
                 is_link=_is_link(type_id, is_enum, opts.force_input),
                 enum_values=enum_values,
                 options=opts,
+                raw_spec=spec,
             )
         )
     return ports
@@ -891,6 +909,9 @@ class Graph:
 
             errors.extend(_check_autogrow_required(node_id, autogrow_ports, autogrow_seen, node_data))
             errors.extend(_check_required_present(node_id, m, node_data))
+            dyn_errors, dyn_warnings = _check_dynamic_combos(node_id, class_type, m, node_data)
+            errors.extend(dyn_errors)
+            warnings.extend(dyn_warnings)
 
         # No-outputs check: the server rejects any prompt with zero output
         # nodes (execution.py:1155-1162, prompt_no_outputs) — including an
@@ -1124,8 +1145,9 @@ def _check_required_present(node_id: str, m: Morphism, node_data: dict) -> list[
     didn't serialize is a hard reject (required_input_missing). The per-input
     loop only inspects keys that ARE present, so this catches the absent ones.
     Skipped, because each is handled by its own path (and would otherwise
-    double-error): autogrow ports (``_check_autogrow_required``) and the dynamic
-    types COMFY_DYNAMICCOMBO / COMFY_DYNAMICSLOT (DynamicSlot is always optional
+    double-error): autogrow ports (``_check_autogrow_required``), dynamic combos
+    (``_check_dynamic_combos``, which reports the absent selector itself so it can
+    also name the valid options), and COMFY_DYNAMICSLOT (always optional
     server-side anyway). The server has NO exemption for required inputs that
     carry a default — the frontend always serializes widget values, so absence
     is a genuine authoring error; we do not skip ports with defaults.
@@ -1135,7 +1157,7 @@ def _check_required_present(node_id: str, m: Morphism, node_data: dict) -> list[
     for port in m.inputs:
         if not port.required or port.is_autogrow:
             continue
-        if port.type.startswith("COMFY_DYNAMICCOMBO") or port.type.startswith("COMFY_DYNAMICSLOT"):
+        if port.is_dynamic_combo or port.type.startswith("COMFY_DYNAMICSLOT"):
             continue
         if port.name in present:
             continue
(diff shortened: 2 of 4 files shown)


bigcat88 (approved):
Approving — I verified the central claim empirically rather than taking the body's word for it, and the server-parity reasoning holds up against real ComfyUI source.

**Server parity (checked against ComfyUI source, not assumed)**
- `DynamicCombo._expand_schema_for_dynamic` (`comfy_api/latest/_io.py:1200-1215`) — confirmed: strict `option["key"] == key`, and when nothing matches it writes *nothing* into `out_dict`, so the selector never reaches `valid_inputs`. Your "accepted by `/prompt`, dies at execution" analysis of the unresolvable-selector case is correct, and I checked the early-return on absent selector at line 1202 too.
- `validate_inputs` (`execution.py:884-901`) — confirmed on both load-bearing points: `valid_inputs` is built from the **finalized** (post-expansion) inputs, so dotted sub-inputs are in scope; and the `x not in inputs` → `required_input_missing` branch has **no exemption for inputs carrying a default**. The PR's strictest assumption is the correct one.

**Red→green proof.** Reproduced your whole mutation table against the captured Seedream catalog. With `origin/main`'s `engine.py` swapped in, all 7 mutations return `valid: true`; with this change each produces exactly the code you documented:

| mutation | main | this PR |
|---|---|---|
| `model.width` deleted | `valid: true` | `required_input_missing` @ `model.width` |
| `model.width` = `"wide"` | `valid: true` | `shape_mismatch` |
| `model.width` = `8` | `valid: true` | `below_min` |
| `model.size_preset` = `"(9K) nope"` | `valid: true` | `unknown_enum_value` |
| `model` = `"seedream 9.9 ultra"` | `valid: true` | `unknown_enum_value` @ `model` |
| `model` deleted | `valid: true` | `required_input_missing` @ `model` |
| flip to `"seedream 5.0 lite"` | `valid: true` | `required_input_missing` × 2 (`model.max_images`, `model.fail_on_partial`) |

The pristine converted workflow stays `valid: true` in both — the over-strictness guard is real. Against main's engine 9 of the 15 new tests fail (body says 8; harmless undercount), the other 6 are the no-false-positive guards.

**End-to-end through the CLI** (`--input` offline mode): good workflow → `ok: true`, exit 0; same workflow minus `model.width` → `required_input_missing`, exit 1.

**Regression surface.** `_check_required_present`'s exemption widened from `startswith("COMFY_DYNAMICCOMBO")` to `is_dynamic_combo` (`COMFY_*COMBO*`). I enumerated the `io_type`s in current ComfyUI — `COMFY_DYNAMICCOMBO_V3` is the only registered type matching, and `Autogrow`/`DynamicSlot`/`MatchType`/`MultiTyped` don't collide. Matching `workflow_to_api._is_widget_input:1044` verbatim is the right call: converter and validator now expand the identical tree by construction.

Full suite: 2861 passed, 7 skipped, 1 failed — `test_non_fast_deps_uses_global_python`, which fails identically on plain `main` here (sandbox artifact). `ruff check` / `ruff format --check` clean.

**On the design question you raised** — keep the two unresolvable-selector cases as hard errors. `comfy validate`'s contract is "will this workflow run", not "will `/prompt` accept these bytes"; a node whose inputs get silently dropped and dies mid-execution after a paid node has been entered is exactly what a user runs validate to avoid. Your messages already draw the distinction honestly, and `test_unresolvable_selector_messages_say_execution_not_prompt` pins it. Don't downgrade to warnings.

**One body inaccuracy, non-blocking.** "Malformed option blocks degrade to 'nothing to expand'" isn't what the code does when the selector *is* present. I probed six malformed shapes (`options` missing / `[]` / `None` / a dict / entries not dicts / no dict at `spec[1]`) and every one produces a hard `unknown_enum_value` — *"'opt_a' not in 0 known options"*. It's unreachable against a real catalog, since `DynamicCombo.Input.as_dict()` (`_io.py:1190`) always emits `options`, so I'm not blocking on it. But either soften the sentence or short-circuit on `not options`, so a future truncated/hand-edited catalog can't turn every dynamic-combo workflow into a wall of errors.

Also agreed on the follow-up you deliberately scoped out: `ok: false` with `error: null` and findings buried in `data.errors` is a real contract wart for envelope consumers. Worth its own ticket, right to keep it out of here.

PR #618, opened 2026-07-28
test: regression sweep for real subgraphed gallery templates (BE-4877)

mattmillerai:
## ELI-5

The template gallery ships workflows whose guts are hidden inside "subgraphs" (a box-of-nodes with a UUID name). comfy-cli has to unpack those boxes to run the workflow and to let agents edit the knobs inside them. That unpacking worked, but nothing tested it against *real* gallery templates — only tiny hand-made ones — so a change in how the frontend saves templates could break every gallery workflow without any test going red. This PR commits two real gallery templates as fixtures and adds tests that prove the boxes get fully unpacked and the inner knobs (prompt, seed) stay reachable.

## What

- Commit `02_qwen_Image_edit_subgraphed.json` and `05_audio_ace_step_1_t2a_song_subgraphed.json` verbatim from Comfy-Org/workflow_templates `templates/` (shasums verified identical to upstream `main`) under `tests/comfy_cli/fixtures/gallery/`, with a README documenting provenance + the refresh command.
- New `tests/comfy_cli/test_subgraph_gallery_templates.py` (9 tests, ~0.6s, fully offline):
  - `convert_ui_to_api(wf, {})` on each fixture must not raise, must return a non-empty prompt, must leave **zero** `is_subgraph_uuid` class_types, and must actually emit subgraph-interior nodes (`<instance>:<interior>` ids) — so the tests fail whether expansion passes instances through *or* silently drops them.
  - Blind-spot tripwire: recursively asserts no fixture carries nested `definitions.subgraphs` inside a subgraph def; the failure message points at `_collect_subgraph_defs` needing recursive collection if the save format ever evolves that way.
  - Slot coverage: `Graph.from_object_info(...).get_template_schema(...)` on the qwen fixture surfaces subgraph-interior slots (`/` addresses), including the concrete `141/132.prompt` and `141/137.seed` an agent actually edits, using a new purpose-built `qwen_object_info.json` covering only that template's node types.

## Verification

- `uv run pytest tests/comfy_cli/test_subgraph_gallery_templates.py` → 9 passed in 0.63s (offline, pure Python).
- Acceptance mutation checks run locally: making `_collect_subgraph_defs` return `{}` → 2 failures; deleting the body of `_expand_subgraphs` → 2 failures. Source restored after each.
- Full suite: 2952 passed, 37 skipped. `ruff check` / `ruff format --check` / pre-commit on the new files: clean.

## Judgment calls

- The README lives at `tests/comfy_cli/fixtures/gallery/README.md` (next to the fixtures) rather than as a line in the parent fixtures dir, which had no README to append to.
- Reused nothing from `subgraph_object_info.json` — its node types (Gemini/Regex/etc.) don't overlap the qwen template at all, so per the ticket's "or" branch I built the minimal `qwen_object_info.json` instead. Frontend-only `MarkdownNote` is intentionally absent, matching a real server's `/object_info`.
- Added the interior-node-presence assertion (`:` ids) beyond the ticket's letter — without it, an expansion regression that *drops* instances instead of passing them through would slip past the zero-UUID check.
- The qwen template upstream is currently ~41 KB (the ticket's ~46 KB reflects an older revision); verbatim-from-main is what's committed.

### DIFF
--- a/tests/comfy_cli/fixtures/gallery/02_qwen_Image_edit_subgraphed.json
+++ b/tests/comfy_cli/fixtures/gallery/02_qwen_Image_edit_subgraphed.json
@@ -1,0 +1,1603 @@
+{
+  "id": "91f6bbe2-ed41-4fd6-bac7-71d5b5864ecb",
+  "revision": 0,
+  "last_node_id": 145,
+  "last_link_id": 295,
+  "nodes": [
+    {
+      "id": 78,
+      "type": "LoadImage",
+      "pos": [
+        -890,
+        30
+      ],
+      "size": [
+        580,
+        450
+      ],
+      "flags": {},
+      "order": 0,
+      "mode": 0,
+      "inputs": [],
+      "outputs": [
+        {
+          "name": "IMAGE",
+          "type": "IMAGE",
+          "links": [
+            292
+          ]
+        },
+        {
+          "name": "MASK",
+          "type": "MASK",
+          "links": null
+        }
+      ],
+      "properties": {
+        "Node name for S&R": "LoadImage",
+        "cnr_id": "comfy-core",
+        "ver": "0.3.50",
+        "enableTabs": false,
+        "tabWidth": 65,
+        "tabXOffset": 10,
+        "hasSecondTab": false,
+        "secondTabText": "Send Back",
+        "secondTabOffset": 80,
+        "secondTabWidth": 65,
+        "ue_properties": {
+          "widget_ue_connectable": {
+            "image": true,
+            "upload": true
+          }
+        }
+      },
+      "widgets_values": [
+        "02_qwen_Image_edit_subgraphed_input_image.png",
+        "image"
+      ]
+    },
+    {
+      "id": 60,
+      "type": "SaveImage",
+      "pos": [
+        130,
+        10
+      ],
+      "size": [
+        970,
+        630
+      ],
+      "flags": {},
+      "order": 5,
+      "mode": 0,
+      "inputs": [
+        {
+          "name": "images",
+          "type": "IMAGE",
+          "link": 269
+        }
+      ],
+      "outputs": [
+        {
+          "name": "images",
+          "type": "IMAGE",
+          "links": [
+            295
+          ]
+        }
+      ],
+      "properties": {
+        "Node name for S&R": "SaveImage",
+        "cnr_id": "comfy-core",
+        "ver": "0.3.65"
+      },
+      "widgets_values": [
+        "ComfyUI"
+      ]
+    },
+    {
+      "id": 119,
+      "type": "MarkdownNote",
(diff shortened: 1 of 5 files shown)


bigcat88 (approved):
Approving. Test-only, zero production code touched, and the fixtures are provably the real thing.

**Provenance verified against upstream, not just asserted.** I fetched both templates from `Comfy-Org/workflow_templates@main` and compared SHA-256:

```
02_qwen_Image_edit_subgraphed.json             local=6bc68f89e77b  upstream=6bc68f89e77b  IDENTICAL
05_audio_ace_step_1_t2a_song_subgraphed.json   local=6c9d9d8cacb3  upstream=6c9d9d8cacb3  IDENTICAL
```

Byte-for-byte. That matters more than usual here — a fixture that has drifted from the gallery is worse than no fixture, because it passes while testing a shape nobody ships.

**Tests are non-vacuous — I redid both mutations independently:**

- `_collect_subgraph_defs` → `return {}`: `test_every_subgraph_instance_expanded` fails on **both** fixtures.
- `_expand_subgraphs` neutered to a pass-through: **4** failures, including `test_converts_with_empty_object_info` on the audio fixture. (You reported 2 — my neutering returned the inputs unchanged rather than deleting the body, which trips the interior-node-presence assertion as well. That extra catch is exactly the assertion you added beyond the ticket's letter, so it earned its place.)

**The slot-addressability claim is real.** On the qwen fixture: **26 slots, 21 of them subgraph-interior** (`/` addresses), and both concrete addresses an agent actually edits resolve:

```
141/132.prompt  present: True
141/137.seed    present: True
```

Conversion leaves **zero** `is_subgraph_uuid` class_types and produces **11 of 16** node ids carrying `:` — so the boxes are genuinely unpacked, not passed through.

The blind-spot tripwire for nested `definitions.subgraphs` is the right kind of test to write: it costs nothing today and will point straight at `_collect_subgraph_defs` if the frontend save format ever nests. Cheap insurance against a change that would otherwise fail silently and confusingly.

Full suite: **3174 passed**, 7 skipped, 1 failed — `test_non_fast_deps_uses_global_python`, the sandbox artifact that fails on plain `main` here too. `ruff check` clean. The new file runs in ~0.1s and is fully offline.

Two small observations, neither blocking:

- `qwen_object_info.json` doesn't cover every node type in the template — converting emits `Could not map widget_values for unknown node type 'ModelSamplingAuraFlow'` and `'ResizeImageMaskNode'`. Harmless for what these tests assert (expansion and addressability both still verified), and keeping the catalog minimal is the right trade. Worth a line in the fixtures README so the next person doesn't read those warnings as a failure.
- Good call omitting frontend-only `MarkdownNote` to match what a real server's `/object_info` actually returns — that's the kind of detail that makes a fixture faithful rather than convenient.

PR #621, opened 2026-07-28
feat(generate): emit envelope/1 from `generate list` / `generate schema` under --json (BE-4933)

mattmillerai:
## ELI-5

`comfy generate list` tells an agent which partner models exist; `comfy generate schema flux-pro` tells it what knobs that model takes. Both used to answer with a pretty terminal table / paragraph of prose even when you asked for JSON, so the only way to read the model list from a script was to `grep` box-drawing characters. Now both answer with the same structured JSON envelope every other machine-readable comfy command returns.

## What changed

Both verbs now resolve the process renderer instead of a local `as_json` bool, and emit `envelope/1` through `renderer.emit`. Pretty (TTY) output is untouched.

- **`comfy_cli/command/generate/app.py`** — `_list_models` (was `main:544`, now `593`) and `_schema` (was `main:583`, now `626`) rewritten; new helpers `_renderer_for` (`544`), `_model_record` (`559`), `_param_record` (`573`). The mechanism the ticket suspected is confirmed: `_separate_meta_flags` (`107`, unchanged) only ever scans `ctx.args`, so a **global** `--json` — the spelling in the ticket's repro — never reached these two functions at all; the **tail** spelling was picked up but emitted a bare blob via `output.print_json` rather than an envelope.
- **`comfy_cli/output/renderer.py`** — new `Renderer.force_json` (`173`), the counterpart to the existing `force_stream` (`162`). `generate` runs with `allow_extra_args`, so its tail `--json` never reaches the global Typer callback that resolves output mode; without the upgrade `renderer.emit` would no-op in pretty mode and the command would print nothing. It only ever upgrades `PRETTY`, so it can never downgrade an NDJSON stream.
- **`comfy_cli/schemas/generate_list.json`, `generate_schema.json`** — new payload schemas, registered in `COMMAND_SCHEMAS` (`comfy_cli/discovery.py:90-94`) so `comfy discover` advertises them. The registration ratchet in `test_discovery.py` requires this for any new `emit(command=...)` site.
- **`comfy_cli/error_codes.py`** — new `generate_model_unknown` (`628`). The missing-model-argument path reuses the existing `missing_argument`. Both replace the old ad-hoc `{"error": "..."}` blob.
- **`comfy_cli/skills/comfy/SKILL.md`** — the bundled agent skill's "Partner-API one-call generation" section previously documented the machine-readable surface of `generate <model>` / `upload` / `--emit-workflow` but said nothing about `list` / `schema`. Added the contract so agents stop reaching for the table.

## Behavior

```
$ comfy --json generate list
{"schema":"envelope/1","type":"envelope","ok":true,"command":"generate list","data":{"models":[{"alias":"flux-2","id":"bfl/flux-2-pro/generate","partner":"bfl","category":"text-to-image","mode":"async","summary":"Proxy request to BFL Flux 2 Pro for image generation"}, ...],"count":52,"filters":{"partner":null,"category":null,"query":null}},"error":null}

$ comfy --json generate schema flux-pro | jq '.data.params[] | select(.name=="output_format")'
{"name":"output_format","type":"enum","kind":"enum","required":false,"default":null,"enum":["jpeg","png"],"description":"Output image format","item_type":null,"upload_mode":null}

$ comfy --json generate schema nope-9000; echo $?
{"schema":"envelope/1",...,"ok":false,"error":{"code":"generate_model_unknown","message":"Unknown model: 'nope-9000'. ...","hint":"run `comfy generate list` to see the available model aliases","details":{"requested":"nope-9000"}}}
1
```

## Acceptance criteria

| Criterion | Status |
|---|---|
| `generate list --json` → valid `envelope/1`, one structured record per model (alias, partner, style, mode, summary) | met |
| `generate schema <model> --json` → valid `envelope/1`, params as structured records with name / type / required / description / enum | met — `output_format` = `jpeg\|png`, exactly the ticket's example |
| Without `--json`, human output unchanged byte-for-byte | met in pretty/TTY mode; **one documented exception below** |
| Summary/description text is the FULL value, not `…`-truncated | met — pinned by `test_list_summaries_are_full_not_table_truncated`, which compares against `spec.list_endpoints()` and asserts at least one summary is long enough that the table *would* have cut it |
| Tests assert parseable `envelope/1` validating against the same schema `discover` / `emit-workflow` use | met — both verbs validate against `comfy_cli/schemas/envelope.json` and their own payload schema |
| Confirmed file paths + line numbers recorded here | met, above |

## Judgment calls

**1. Non-TTY without `--json` now returns the envelope, not the table.** This is the one place the change is not strictly additive, and it deserves an explicit look. The CLI's global contract (`Renderer.resolve` precedence rule 6, pinned by `test_non_tty_auto_selects_json`) is *"stdout is not a TTY → JSON"*; every renderer-backed command already behaves this way. Once these two verbs emit envelopes, they inherit it — so `comfy generate list | grep` now yields JSON rather than a table. That is the intended outcome (an agent that forgets `--json` still gets machine-readable output instead of the scraping problem this ticket is about), and it satisfies the precondition the existing module comment in `app.py` set for exactly this migration: *"Migrate only once `generate` emits envelopes via the renderer."* Interactive TTY output is byte-for-byte identical — verified by diffing all six reachable pretty paths (list, list+filter, list-with-no-matches, schema, unknown model, missing argument) against `origin/main`. `--no-json` and `COMFY_OUTPUT=pretty` force the table back in a pipe. Three existing tests in `test_app.py` were adjusted for this (two `--json` shape assertions moved to the envelope; two pretty-output assertions now pass `--no-json`, since `CliRunner` has no TTY).

**2. `data.params[]` carries both `type` and `kind`, always identical.** The ticket names `type`; `kind` is the vocabulary `generate/schema.py` uses internally and the key the pre-envelope `--json` payload shipped. Keeping both costs one field and is documented in `generate_schema.json` as a retained alias — prefer `type`.

**3. `category`, not a duplicate `style` key.** The human table's column header is "Style" but the shipped field name (and the `--style`/`--category` filter target) is `category`. Rather than fork the contract with a synonym, the schema documents the mapping.

**4. Additions beyond the strict minimum**, all optional and schema-documented: `filters` echoed back on `list`; `partner`/`category`/`summary`/`mode`/`polling`/`content_type`/`example` on `schema`; `item_type`/`upload_mode` per param (`upload_mode` tells an agent that a local file path passed to that field is auto-base64'd or auto-uploaded — non-obvious and otherwise undiscoverable).

**5. Left alone deliberately.** The other reserved sub-actions (`refresh`, `upload`, `resume`, `consent`) are out of scope per the ticket. Also pre-existing and untouched: `_separate_meta_flags` raises an uncaught `SchemaError` on a valueless meta flag (`comfy generate list --timeout`), which surfaces as a traceback rather than a structured error — same on `main`, unrelated to `--json`, and fixing it would widen this diff into the shared parser every `generate` path uses.

The diff adds an "Unknown model" denial string, which trips the falsification trigger, so: this is not a new dead-end. `main` already prints `Unknown model: '<x>'` and exits 1 for the same input — the change only re-codes that existing denial as a structured `generate_model_unknown` error, and no test was flipped to assert a dead-end. The denial is also not terminal: its `hint` and `details.requested` point at `comfy generate list`, which was exercised live and returns all 52 models. The diff's net user-facing effect is to *add* a capability (structured discovery), not remove one.

## Testing

- `tests/comfy_cli/command/generate/test_list_schema_envelope.py` — 28 new tests: envelope + payload-schema validation for both verbs under both `--json` spellings; full-summary/full-description parity against the spec; filter echo; empty result as `ok:true` exit 0; both error codes with registry membership; `force_json` upgrading a pretty renderer (including over a global `--no-json`); NDJSON mode terminating in the envelope; stderr clean in JSON mode; pretty output preserved on all paths; discover registration.
- Full suite: **2986 passed, 37 skipped**.
- `ruff check` + `ruff format --diff` clean at the CI-pinned `0.15.15`.

### DIFF
--- a/comfy_cli/command/generate/app.py
+++ b/comfy_cli/command/generate/app.py
@@ -44,7 +44,7 @@ from rich.progress import Progress, SpinnerColumn, TextColumn, TimeElapsedColumn
 from comfy_cli import constants, tracking, ui
 from comfy_cli.command.generate import adapters, client, emit, output, poll, schema, spec, upload
 from comfy_cli.config_manager import ConfigManager
-from comfy_cli.output.renderer import get_renderer
+from comfy_cli.output.renderer import Renderer, get_renderer
 
 _HELP = "Generate images via ComfyUI partner nodes (Flux, Ideogram, DALL·E, Recraft, Stability, …)."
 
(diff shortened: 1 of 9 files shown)


bigcat88 (approved):
Approving. The one non-additive change here is the interesting part, and it holds up — I verified the "human output unchanged" claim byte-for-byte rather than spot-checking it.

**Mechanism confirmed.** On `origin/main`, `comfy --json generate list` prints the **table** — the global flag never reaches these verbs at all, exactly as you diagnosed via `_separate_meta_flags` only scanning `ctx.args`. On this branch it returns a proper `envelope/1`. That's the bug, reproduced both ways.

**Pretty output is byte-identical — all six reachable paths.** I captured each under `COMFY_OUTPUT=pretty` on this branch and again with `origin/main`'s `app.py` + `renderer.py` swapped in, then diffed:

```
IDENTICAL  list            IDENTICAL  schema
IDENTICAL  list_filter     IDENTICAL  schema_bad (unknown model)
IDENTICAL  list_empty      IDENTICAL  schema_missing (no argument)
```

No drift, including the error paths — which is where this kind of migration usually leaks.

**`force_json` cannot downgrade a stream.** This was my main worry given `emit` semantics, so I exercised all three modes directly:

```
PRETTY -> JSON      JSON -> JSON      NDJSON -> NDJSON
```

The NDJSON case is the one that would have been a real contract break for a `--json-stream` consumer, and it's safe.

**Envelope and payload both validate.** `--json generate list` and `--json generate schema flux-pro` each validate against `envelope.json` *and* their own payload schema (`generate_list.json` / `generate_schema.json`), resolved through the same store `discover` uses. Both `--json` spellings — global and tail — produce identical envelopes with `count: 52`.

**Summaries are genuinely full**: longest is **117 chars**, zero ellipsis characters anywhere in the set. The table would have cut that, which is the whole point of `test_list_summaries_are_full_not_table_truncated`.

**Error paths**: `schema nope-9000` → `generate_model_unknown`, `schema` with no argument → `missing_argument`. `generate_model_unknown` is registered and the registry is still duplicate-free (worth checking today of all days, given #628).

**Tests are non-vacuous** — I neutered `force_json()` to a bare `return` and `test_tail_json_upgrades_an_otherwise_pretty_renderer[args1]`/`[args2]` failed immediately, along with one more. The guard is real.

Full suite: **3156 passed**, 7 skipped, 1 failed — `test_non_fast_deps_uses_global_python`, the sandbox artifact present on plain `main` here too. `ruff check` clean.

**On judgment call #1 — I agree, and I want to record why**, because it's the one thing a future bisect might land on. `comfy generate list | grep` now yields JSON instead of a table. That *is* a user-visible change, but it's the repo's own documented precedence rule ("stdout is not a TTY → JSON", pinned by `test_non_tty_auto_selects_json`), every other renderer-backed command already behaves this way, and the module comment in `app.py` explicitly set this migration as the precondition. Interactive use is unaffected, and `--no-json` / `COMFY_OUTPUT=pretty` both restore the table in a pipe — I confirmed both. Worth a line in the release notes so anyone scraping the table in a script isn't surprised.

The other calls are all fine: keeping `kind` as a documented alias for `type` costs one field and avoids breaking the pre-envelope payload; `category` over a synthetic `style` synonym is right (the filter flag already targets `category`); and `upload_mode` is a genuinely non-obvious thing for an agent to discover any other way.

Good call flagging `_separate_meta_flags` raising an uncaught `SchemaError` on `comfy generate list --timeout` and leaving it alone — it's pre-existing, unrelated to `--json`, and fixing it means touching the shared parser every `generate` path uses. Worth its own ticket.

PR #626, opened 2026-07-29
feat: add system-stats and free commands

mattmillerai:
## ELI-5
Comfy CLI can now ask a ComfyUI server (local or cloud) how much VRAM/RAM is free, and can tell it to unload models / clear its cache — the same two things the ComfyUI web UI's memory panel does, but scriptable: `comfy system-stats` and `comfy free`.

## What changed
- `comfy_cli/comfy_client.py`: added `get_system_stats()` (GET `system_stats`) and `post_free(unload_models=True, free_memory=False)` (POST `free`) to the shared `Client` class, following the existing `get_history`/`get_job_status` style — both just delegate to `_request`, so retries/auth/redaction are unchanged.
- `comfy_cli/command/system.py` (new): `system_stats_execute` and `free_execute`, wired as top-level `@app.command`s in `cmdline.py` (same shape as `stop`/`launch`/`outdated`). Both resolve the target via `resolve_target` (respects the global `--where`, defaults to local) and reuse the existing `server_not_running` / `cloud_http_error` error-code convention for an unreachable server.
  - `system-stats`: JSON mode passes the server's `/system_stats` payload through unmodified (`renderer.emit(stats, ...)`) so the schema tracks ComfyUI's own shape; pretty mode renders a Rich table of devices (name/type/index/vram_free/vram_total, human-readable) plus a system RAM/version line.
  - `free`: `--unload-models/--no-unload-models` (default on) and `--free-memory` (default off, implies unload server-side per ComfyUI's own semantics — the CLI does not synthesize that on the client side, it just forwards the flags). Emits `{"requested": {...}, "note": "..."}` describing that the flags apply on the worker's next iteration and never interrupt a running job.
- `comfy_cli/discovery.py` + two new `comfy_cli/schemas/*.json`: registered both commands' output schemas (required by `test_every_emitted_command_registers_a_schema`). `system_stats.json` only pins the fields agents rely on (`additionalProperties: true`) since the payload is a ComfyUI passthrough; `free.json` pins the fixed `requested`/`note` shape.
- `README.md`: one-line mentions of both commands next to the existing launch/stop docs.
- `tests/comfy_cli/command/test_system.py` (new): mocks HTTP at the `Client`'s own opener seam. Covers the JSON passthrough shape, the pretty/local/cloud unreachable-server error codes (`server_not_running` vs `cloud_http_error`), an HTTP-level error status surfacing in `error.details`, and the `free` request body for the default flags plus three explicit on/off combinations.

## Verification
- `ruff check .` / `ruff format --check .` clean on the touched files (repo-wide `ruff format --check` only flags one pre-existing unrelated doc file).
- Full `pytest` suite green on Python 3.14 (3079 passed) and re-ran the new/discovery/error-registry tests on a fresh Python 3.10 venv (both clean). Three pre-existing failures are unrelated to this change and reproduce identically on `origin/main` before this branch: `test_node_deps.py::test_non_pep440_installed_version_is_unknown_not_a_crash`, `test_broken_pipe.py::TestBrokenPipeExitCode::test_closed_stdout_exits_zero`, `test_broken_pipe.py::TestBrokenPipeExitCode::test_head_pipeline_exits_zero`, and `test_error_code_registry.py::test_no_duplicate_codes` (a genuine pre-existing duplicate `server_died` entry in `error_codes.py`, also reproduces on `origin/main`, out of scope here).
- Manually smoke-tested both commands end-to-end against a throwaway local HTTP server implementing `/system_stats` + `/free`, in both `--json` and pretty mode, and against an unreachable port (connection-refused path) — all matched the automated-test expectations. No live ComfyUI instance was available in this environment, so the acceptance criterion's literal "against a live local ComfyUI" step is unverified beyond the throwaway server standing in for the documented `/system_stats` response shape from the ticket's spike evidence.

## Judgment calls
- The ticket's snippet names the class `ComfyClient`; the actual class in `comfy_client.py` is `Client` (used elsewhere as `from comfy_cli.comfy_client import Client`). Added the two methods to the existing `Client` class rather than introducing a new name.
- Added `--where` as the only routing option on both commands (no `--host`/`--port`), matching the simplest existing precedent (`comfy models list-folders`) — `resolve_target` already handles `COMFY_LOCAL_URL` / persisted defaults for local without them.

### DIFF
--- a/README.md
+++ b/README.md
@@ -206,6 +206,9 @@ Comfy provides commands that allow you to easily run the installed ComfyUI.
   - Maximum of 10 PR builds are kept (oldest are removed automatically)
   - Cache limits help manage disk space while keeping recent builds available
 
+- To check VRAM/RAM usage: `comfy system-stats` (add `--where cloud` to target Comfy Cloud instead of local)
+- To unload models / free the executor cache: `comfy free` (pass `--free-memory` to also reset the executor cache)
+
 ### Managing Custom Nodes
 
 comfy provides a convenient way to manage custom nodes for extending ComfyUI's functionality. Here are some examples:
--- a/comfy_cli/cmdline.py
+++ b/comfy_cli/cmdline.py
@@ -1207,6 +1207,49 @@ def validate_comfyui(_env_checker):
         raise typer.Exit(code=1)
 
 
+@app.command(
+    "system-stats",
+    help="Show ComfyUI system stats: per-device VRAM + system RAM/version (GET /system_stats).",
+)
+@tracking.track_command()
+def system_stats(
+    where: Annotated[
+        str | None,
+        typer.Option("--where", show_default=False, help="Routing target: 'local' (default) or 'cloud'."),
+    ] = None,
+):
+    from comfy_cli.command import system as system_command
+
+    system_command.system_stats_execute(get_renderer(), where=where)
+
+
+@app.command(
+    help="Ask ComfyUI to unload models / free the executor cache (POST /free). "
+    "Applies on the worker's next iteration — never interrupts a running job."
+)
+@tracking.track_command()
+def free(
+    unload_models: Annotated[
+        bool,
+        typer.Option("--unload-models/--no-unload-models", help="Unload all models from VRAM."),
+    ] = True,
+    free_memory: Annotated[
+        bool,
+        typer.Option(
+            "--free-memory",
+            help="Also reset the executor cache; implies --unload-models on the server side.",
+        ),
+    ] = False,
+    where: Annotated[
+        str | None,
+        typer.Option("--where", show_default=False, help="Routing target: 'local' (default) or 'cloud'."),
+    ] = None,
+):
+    from comfy_cli.command import system as system_command
+
+    system_command.free_execute(get_renderer(), where=where, unload_models=unload_models, free_memory=free_memory)
+
+
 @app.command(help="Stop background ComfyUI")
 @tracking.track_command()
 def stop():
--- a/comfy_cli/comfy_client.py
+++ b/comfy_cli/comfy_client.py
@@ -432,6 +432,33 @@ class Client:
                 return None
             raise
 
+    # ----- resource management -----
+
+    def get_system_stats(self, *, timeout: float | None = None) -> dict | None:
+        """GET {prefix}/system_stats.
+
+        Per-device VRAM (``vram_total``/``vram_free``/``torch_vram_total``/
+        ``torch_vram_free``) plus system RAM and version info. Same shape on
+        local and cloud — passed through unmodified so the schema tracks
+        whatever ComfyUI itself reports.
+        """
+        return self._request("GET", ("system_stats",), timeout=timeout)
+
+    def post_free(self, *, unload_models: bool = True, free_memory: bool = False, timeout: float | None = None) -> None:
+        """POST {prefix}/free — ask the queue worker to unload models / free the cache.
+
+        ``free_memory=True`` implies unloading models on the server side even
+        when ``unload_models`` is False. The flags apply the next time the
+        worker iterates: immediately if it's idle, after the current job if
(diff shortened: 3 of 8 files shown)


bigcat88 (approved):
Approving. Your verification note says a live ComfyUI wasn't available — I have one, so I ran both commands against a real server (ComfyUI **0.27.0**, `--cpu`, port 8189) rather than a stub. Everything holds.

**`comfy system-stats` — live.** JSON mode passes the payload straight through: `comfyui_version: 0.27.0`, the full `system` block, and `devices: [("cpu", "cpu", 160269164544)]`. Pretty mode renders correctly:

```
┃ name ┃ type ┃ index ┃ vram_free ┃ vram_total ┃
│ cpu  │ cpu  │ None  │ 149.3 GiB │ 182.1 GiB  │
RAM: 149.3 GiB / 182.1 GiB free — ComfyUI 0.27.0
```

**`comfy free` — live, all four flag combinations**, each returning `ok: true` with the right echo:

| invocation | `requested` |
|---|---|
| `comfy free` | `unload_models=True, free_memory=False` |
| `--free-memory` | `unload_models=True, free_memory=True` |
| `--no-unload-models` | `unload_models=False, free_memory=False` |
| `--no-unload-models --free-memory` | `unload_models=False, free_memory=True` |

**The thing a stub server would most likely have hidden:** ComfyUI's `/free` returns `web.Response(status=200)` with **no body at all** (`server.py:1186`), not JSON. That's exactly the shape that makes a client blow up on `json.loads("")`. It's safe here — `_request` short-circuits with `if not text: return None` — but it's worth recording that this was verified against the real endpoint rather than a stand-in that returned `{}`.

**Server parity checked in ComfyUI source, not assumed:**
- `/system_stats` (`server.py:708-726`) emits exactly the `{system, devices}` shape you pass through, and the per-device keys you table (`name`/`type`/`index`/`vram_total`/`vram_free`) are the real ones.
- `/free` (`server.py:1177-1186`) reads precisely `unload_models` and `free_memory` from the JSON body — the two keys you send, no more.
- **Your `free_memory` implies unload claim is correct**, and I'm glad you didn't synthesize it client-side. `main.py:390` is `if flags.get("unload_models", free_memory):` — `free_memory` is the *default* for `unload_models`, and since the handler only calls `set_flag("unload_models", …)` when truthy, sending `{"unload_models": false, "free_memory": true}` leaves the key unset and the server unloads anyway. Forwarding the flags verbatim keeps the CLI honest about semantics it doesn't own.

**Error paths** behave: unreachable local server gives `server_not_running` with hint *"run `comfy launch` to start a local server"* and **exit 1**, on both commands. Both commands appear in `comfy discover`, so the schema registration is real.

**Tests are non-vacuous** — I mutated the source twice and both were caught:
- forcing `post_free` to always send `unload_models=True` → `test_flag_combinations[False-False]` and `[False-True]` fail
- swapping `server_not_running` for a generic code → both `test_local_unreachable_server_not_running` cases fail

Full suite with current `main` merged in: **3068 passed**, 7 skipped, 1 failed — `test_non_fast_deps_uses_global_python`, the sandbox artifact that fails on plain `main` here too. `ruff check` / `ruff format --check` clean over 263 files.

**Your CI red is stale, not real.** The `build` job failed on `test_no_duplicate_codes` — the duplicate `server_died` entry, which landed as a `main` breakage from #605/#604 and was fixed by **#628** a few minutes ago. The other pre-existing failures you listed (`test_broken_pipe`, `test_node_deps::test_non_pep440_installed_version_is_unknown_not_a_crash`) are also gone once current `main` is merged. Nothing to do on your side.

One cosmetic nit, not blocking: the pretty table prints `index` as the literal `None` for CPU devices (ComfyUI sets `d.index = None` there). A `—` would read better and match how you handle empty values elsewhere.

Agreed on both judgment calls — adding to the existing `Client` rather than inventing a `ComfyClient`, and `--where` only with no `--host`/`--port`, matching `models list-folders`.

PR #653, opened 2026-08-02
fix(workflow): emit a JSON failure envelope for bad routing instead of a traceback

### LINKED ISSUE #637: [groom] object_info/Graph loading and routing-mode resolution reimplemented per command, with divergent handling of an invalid --where
**Filed by the groom sweep** — Comfy-Org/comfy-cli (`whole-repo`) · verdict **DOWNGRADE** · [run](https://github.com/Comfy-Org/comfy-cli/actions/runs/30535304291).

## Verified -- the bug is real

comfy_cli/where.py:80 resolve_default reads the persisted config defensively but its docstring is explicit: callers keep their own error handling for the ValueError a bad flag, env, project or config value raises. Four call sites; two do not:

- command/nodes.py:38 _resolved_where -- catches ValueError and retries with resolve(flag=where, config_value=None). Defensive.
- command/jobs.py:1525 _is_cloud -- catches ValueError, falls back to local. Defensive.
- target.py:70-77 -- wraps the config read in except Exception. Defensive about the read.
- command/workflow.py:100 _get_graph -- BARE resolve_default() inside a try whose only handler is except LoadError. A bad persisted where_default or a typo'd COMFY_WHERE escapes as an uncaught ValueError: a raw traceback on a JSON-envelope command.
- command/workflow_fragments.py:334 _load_object_info -- the same bare call inside a try catching only OSError and json.JSONDecodeError. Same leak.

Also confirmed: workflow.py _get_graph takes no where parameter at all (no per-command --where), unlike nodes.py _get_graph (:55), and the two diverge in whether mode reaches Graph.load and whether e.details hint text is forwarded into the envelope.

## Verdict -- the bug fix is the deliverable

Do (small, high value): add one defensive resolver -- where.resolve_mode(flag=None) returning local or cloud, implementing nodes.py _resolved_where's behavior (catch ValueError, retry with config_value=None) -- and use it at nodes.py:38, jobs.py:1525, workflow.py:100 and workflow_fragments.py:334. That removes two traceback paths from agent-facing commands. Add regression tests with COMFY_WHERE=bogus and a corrupt persisted where_default for comfy workflow and comfy workflow fragment, in pretty and JSON modes.

DOWNGRADED AWAY for now: the shared load_graph(input_path, host, port, where, on_stale). The three copies are not interchangeable -- workflow_fragments._load_object_info returns a raw dict (not a Graph), reads --input itself with json.loads, and maps failures to object_info_unavailable, while nodes.py and workflow.py return a Graph and map to cql_no_graph with different hints and different details forwarding. One function needs a mode knob, a details knob and a return-type knob on day one. If someone wants it, unify only nodes.py:55 and workflow.py:80 (both Graph-returning, both cql_no_graph) and give workflow.py a --where option as a deliberate documented feature change, not a refactor side effect.

## Steelman

The ValueError leak is arguably a two-line bug fix rather than a refactor, and the copies differ in ways a shared function must parameterize.
Counter: the same leak reached two of four copies precisely because there are four; the one-resolver fix and the dedupe are the same change, and it stops there.

## Risk

Low. Not an auth boundary, but note this predicate chooses local vs cloud routing (hence where a workflow is sent): the fallback must stay the current fail-soft resolve(..., config_value=None) behavior and must not invent a new default. cql/loader.resilient_load_object_info is already shared and is not touched.

mattmillerai:
## ELI-5

If you typo the routing setting — `COMFY_WHERE=clod` instead of `cloud` — some `comfy workflow` commands used to crash with a raw Python traceback instead of the clean JSON error they promise. A script or agent reading that output gets garbage where it expected an envelope. Now it gets a normal `where_invalid` error with a hint about what to fix.

## What was wrong

`where.resolve_default()` raises `ValueError` on an invalid routing value, and its docstring is explicit that **callers keep their own error handling**. Five call sites; two did not have any:

| Call site | Before |
|---|---|
| `command/nodes.py` `_resolved_where` | catches `ValueError`, retries with the bad config dropped |
| `command/jobs.py` `_is_cloud` | catches `ValueError`, falls back to local |
| `command/launch.py` `logs` | catches `ValueError`, emits a `where_invalid` envelope |
| `command/workflow.py` `_get_graph` | **bare call** inside a `try` whose only handler is `except LoadError` |
| `command/workflow_fragments.py` `_load_object_info` | **bare call** inside a `try` catching only `OSError` / `json.JSONDecodeError` |

A bad persisted `where_default`, a stale `defaults.where` in the project `comfy.yaml`, or a typo'd `COMFY_WHERE` therefore escaped as an uncaught `ValueError` from `comfy workflow slots` / `set-slot` / `vary` / `decompose` — a raw traceback on a JSON-envelope command.

Reproduced before the fix (both paths), with `COMFY_WHERE=clod`:

```
$ COMFY_WHERE=clod comfy --json workflow slots wf.json
ValueError: invalid --where value 'clod': expected one of ['local', 'cloud']
```

## What changed

Added `where.resolve_default_or_exit()` — the same emit-and-exit shape as the existing `cloud_preflight_or_exit()` in that module (lazy `typer` / renderer imports, `renderer.error(...)` then `typer.Exit(1)`), so the envelope stays identical across call sites instead of the block being re-inlined. Both leaking sites now use it. After:

```
$ COMFY_WHERE=clod comfy --json workflow slots wf.json ; echo $?
{"schema": "envelope/1", ..., "ok": false, "error": {"code": "where_invalid", "message": "invalid --where value 'clod': expected one of ['local', 'cloud']", "hint": "use --where local or --where cloud, and check COMFY_WHERE, `defaults.where` in comfy.yaml, and `comfy set-default --where`"}}
1
```

**Why an envelope rather than the silent fallback `nodes`/`jobs` use.** The ticket's scope line asks for both "handle it the way the others do" and "emit the normal JSON failure envelope". Those point different directions, so: `nodes` and `jobs` can silently fall back because each has a per-command `--where` flag to fall back *to*. `workflow._get_graph` takes no `where` parameter at all, so there is nothing to recover to — and silently routing to `local` when the user explicitly (if mistakenly) asked for something else is worse for a machine consumer than a clean, actionable error. `launch.py logs` already sets that precedent with the same `where_invalid` code. Flagging it as a judgment call.

## Adjacent fix found during self-review

`_load_object_info`'s handler catches only `OSError` / `json.JSONDecodeError`, but `resilient_load_object_info` reports failures as `LoadError`. `git log -L` shows the handler predates that loader migration (it was written when the fetch raised a bare `OSError`), so an **unreachable local server** — the far more common case than a typo'd env var — also escaped as a traceback, even though the handler's own hint ("start the server with `comfy launch`") is written for exactly that case. `workflow.py`'s sibling `_get_graph` already catches `LoadError`. Added it to the tuple, so the already-written `object_info_unavailable` envelope now actually fires:

```
$ COMFY_WHERE=local comfy workflow decompose ui.json ; echo $?     # no server running
# before: LoadError: cannot reach http://127.0.0.1:8188/object_info: [Errno 61] Connection refused
{"ok": false, "error": {"code": "object_info_unavailable", "message": "Could not load object_info: cannot reach ...", "hint": "pass --input <object_info.json>, sign into cloud, or start the server with `comfy launch`"}}
1
```

## Tests

Nine tests added, **each verified red against this branch with the source fix stashed**:

- `tests/comfy_cli/auth/test_where.py` — four unit tests on the wrapper: valid routing returns quietly and emits nothing; a typo'd `COMFY_WHERE` and a bad persisted `where_default` each emit `where_invalid` and raise `Exit(1)`; a valid `flag` still beats a bad config.
- `tests/comfy_cli/command/test_workflow_slots.py` — `slots` and `set-slot` with a typo'd `COMFY_WHERE` produce the envelope (the `slots` one asserts `result.exception is None`, i.e. nothing escaped as a traceback), plus a positive guard that `--input` short-circuits routing entirely so a bad value stays harmless.
- `tests/comfy_cli/command/test_workflow_fragments.py` — `decompose` with a typo'd `COMFY_WHERE`; `decompose` against an unreachable server (the `LoadError` case); plus a positive guard that an API-format workflow never resolves routing at all.

Full suite green: **3752 passed, 37 skipped**. `ruff check` and `ruff format --diff` clean using the CI-pinned **0.15.15** (note: with a newer local ruff, `origin/main` already reports 17 pre-existing lint errors and one format diff — unrelated to this change, and clean at the pinned version).

## Out of scope — but worth a separate issue

**`target.py resolve_target` has the same half-defense and leaks the same way.** It wraps the config *read* in `except Exception` but not the value *parse*, so `where_module.resolve(...)` at `target.py:77` still raises. Verified live on `main`:

```
$ COMFY_WHERE=clod comfy --json system-stats
ValueError: invalid --where value 'clod': expected one of ['local', 'cloud']
```

That is a different function with ~19 call sites needing per-command judgment (several pass `where="cloud"` literally and can never fail), so fixing it here would blow up a scoped correctness fix into a sweep. Recommend a follow-up issue. The broader "unify object_info/Graph loading across commands" proposal is likewise untouched.

Closes #637

### DIFF
--- a/comfy_cli/command/workflow.py
+++ b/comfy_cli/command/workflow.py
@@ -104,7 +104,10 @@ def _get_graph(input_path: str | None, host: str | None, port: int | None, on_st
         # Live fetch: resolve mode from global routing chain, then use resilient loader.
         from comfy_cli import where as where_module
 
-        decision = where_module.resolve_default()
+        # ``_or_exit``: this command has no per-command --where flag to fall
+        # back to, so a bad COMFY_WHERE / project / persisted where_default
+        # becomes a clean `where_invalid` envelope rather than a traceback.
+        decision = where_module.resolve_default_or_exit()
         mode = "cloud" if decision.target is where_module.WhereTarget.CLOUD else "local"
         from comfy_cli.cql.loader import resilient_load_object_info
 
--- a/comfy_cli/command/workflow_fragments.py
+++ b/comfy_cli/command/workflow_fragments.py
@@ -324,6 +324,13 @@ def _slug_name(text: str) -> str:
 
 def _load_object_info(renderer, *, input_path: str | None, host: str | None, port: int | None) -> dict:
     """Fetch raw object_info for frontend→API conversion (offline dump or live server)."""
+    # ``LoadError`` is how ``resilient_load_object_info`` reports an unreachable
+    # server / bad dump — the handler below predates that loader (it was written
+    # when the fetch raised a bare OSError), so an unreachable local server used
+    # to escape as a traceback even though the hint here is written for exactly
+    # that case. ``workflow.py``'s sibling ``_get_graph`` already catches it.
+    from comfy_cli.cql.engine import LoadError
+
     try:
         if input_path is not None:
             p = Path(input_path).expanduser()
@@ -331,10 +338,14 @@ def _load_object_info(renderer, *, input_path: str | None, host: str | None, por
         from comfy_cli import where as where_module
         from comfy_cli.cql.loader import resilient_load_object_info
 
-        decision = where_module.resolve_default()
+        # ``_or_exit``: no per-command --where flag here either, so a bad
+        # COMFY_WHERE / project / persisted where_default emits a clean
+        # `where_invalid` envelope instead of escaping as a traceback (the
+        # handler below is for object_info failures, not routing ones).
+        decision = where_module.resolve_default_or_exit()
         mode = "cloud" if decision.target is where_module.WhereTarget.CLOUD else "local"
         return resilient_load_object_info(mode=mode, host=host, port=port)
-    except (OSError, json.JSONDecodeError) as e:
+    except (OSError, json.JSONDecodeError, LoadError) as e:
         renderer.error(
             code="object_info_unavailable",
             message=f"Could not load object_info: {e}",
--- a/comfy_cli/where.py
+++ b/comfy_cli/where.py
@@ -101,6 +101,40 @@ def resolve_default(
     return resolve(flag=flag, env=env, config_value=config_value, project_value=project_value)
 
 
+def resolve_default_or_exit(
+    flag: str | None = None,
+    *,
+    env: Mapping[str, str] | None = None,
+    project_value: str | None = _UNSET,
+) -> WhereResolution:
+    """:func:`resolve_default` that emits a ``where_invalid`` envelope and exits.
+
+    The shared "emit-and-exit" wrapper for command call sites that have no
+    sensible fallback for a bad routing value — a typo'd ``COMFY_WHERE``, a
+    stale ``defaults.where`` in the project ``comfy.yaml``, or a corrupt
+    persisted ``where_default`` would otherwise escape as a raw ``ValueError``
+    traceback, which is the worst possible shape for a machine consumer of a
+    JSON-envelope command.
+
+    Commands that *can* recover (``nodes``, ``jobs``) keep their own
+    ``except ValueError`` fallback instead of calling this.
+    """
+    import typer
+
+    from comfy_cli.output import get_renderer
+
+    try:
+        return resolve_default(flag=flag, env=env, project_value=project_value)
+    except ValueError as e:
+        get_renderer().error(
+            code="where_invalid",
+            message=str(e),
+            hint="use --where local or --where cloud, and check COMFY_WHERE, "
+            "`defaults.where` in comfy.yaml, and `comfy set-default --where`",
+        )
+        raise typer.Exit(code=1) from e
+
+
 def _project_where_default() -> str | None:
     """``defaults.where`` from the project/1 ``comfy.yaml`` governing cwd, if
     any. Discovery itself never raises (see :mod:`comfy_cli.project`); a
--- a/tests/comfy_cli/auth/test_where.py
+++ b/tests/comfy_cli/auth/test_where.py
@@ -100,6 +100,65 @@ def test_resolve_default_forwards_project_value(fake_config, monkeypatch):
 
 
 # ---------------------------------------------------------------------------
+# resolve_default_or_exit: the emit-and-exit wrapper for call sites with no
+# per-command --where flag to fall back to
+# ---------------------------------------------------------------------------
+
+
+@pytest.fixture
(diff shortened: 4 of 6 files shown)


bigcat88 (approved):
Approving. Reproduced the leak on `main` and the fix here, on both paths.

### The contract really was broken

`COMFY_WHERE=clod`, UI-format workflow, `--json`:

| command | `main` | this branch |
| --- | --- | --- |
| `workflow slots` | `ValueError: invalid --where value 'clod'` + Rich traceback — **stdout is not valid JSON** (`JSONDecodeError`) | `where_invalid` envelope, valid JSON |
| `workflow decompose` | same traceback, **stdout is not valid JSON** | `where_invalid` envelope, valid JSON |

The "not valid JSON" part is the sharp end and worth stating outright in the description: this isn't a cosmetically ugly error, it's an envelope/1 violation on a command that advertises the contract. An agent piping `--json` gets a parse failure, not an error it can branch on.

(Note for anyone reproducing: an API-format workflow won't show it — `workflow slots` rejects the format before it ever resolves routing. It needs a frontend-format file, e.g. `tests/comfy_cli/fixtures/sd15_ui_workflow.json`.)

### Checks

- `where_invalid` is registered (`error_codes.py:237`), so the two-way registry test covers the new raise sites — the full suite passing is meaningful here rather than incidental.
- `resolve_default_or_exit` has exactly the two call sites the description names (`workflow.py:112`, `workflow_fragments.py:345`); the other three `resolve_default()` callers are untouched and keep their own handling, which is right — `nodes.py` and `jobs.py` deliberately *recover* rather than exit, and collapsing them into the emit-and-exit shape would have changed behaviour.
- Mirroring `cloud_preflight_or_exit()`'s shape (lazy imports, `renderer.error` then `typer.Exit(1)`) keeps the envelope identical across call sites instead of re-inlining the block. Right call.

### Compatibility with today's merges

`command/workflow.py` moved twice on `main` in the last few hours — #595 replaced its local `_atomic_write_text` with the shared helper, and #655 rewrote `_local_list` / `_cloud_list` for markup sanitisation. Neither touches `_get_graph`'s routing resolution; the merged tree is textually clean and the full suite is green on it: **3796 passed, 31 skipped**. `ruff check` / `ruff format --diff` clean at the CI-pinned 0.15.15.

PR #654, opened 2026-08-02
fix(http): bound three unbounded HTTP body reads behind a shared read cap

mattmillerai:
## ELI-5

When comfy-cli asks a server for something, it has to read the reply into memory. Three places read the whole reply with no size limit — whatever the server sends, we hold. The sharp one fetches the workflow-template gallery from a host on the internet. A server that just keeps sending would grow the CLI's memory until it dies. This adds one shared "read at most N bytes, then refuse" helper and puts those three reads behind it.

## What changed

**New primitive** — `comfy_cli/http.py`:

- `MAX_RESPONSE_BYTES = 64 MiB`, with the rationale written at the constant rather than inherited from whichever call site happened to be nearest.
- `read_capped(resp, url, *, max_bytes=MAX_RESPONSE_BYTES) -> bytes` — reads `cap + 1` so a body that exactly fills the cap is still recognizable as complete, and raises the existing `ResponseTooLarge` rather than returning a silently truncated body. Truncation is the worse failure: truncated JSON surfaces as a confusing parse error, and a truncated file body gets written to disk as if it were whole.
- `request_json` now delegates its own capped read to the primitive. Externally unchanged — same exception, same message, same early `max_bytes` validation before the network call.

**Three unbounded readers converted**, each keeping its established failure contract so no user-visible error changes shape:

| Reader | Was | Over-cap now surfaces as |
|---|---|---|
| `templates._fetch_gallery` | `resp.read()` on a remote gallery URL | `ResponseTooLarge` → `_GALLERY_LOAD_ERRORS` → stale-cache fallback on a TTL refresh, `gallery_load_failed` / `gallery_fetch_failed` envelope otherwise |
| `jobs._http_get_json` | `json.loads(resp.read())` | `RuntimeError` — the single family all seven call sites already catch |
| `comfy_client.Client._request` | `resp.read().decode(...)` | the module's own `HTTPError`, the same shape `submit_prompt` already uses for a bad 200 |

**Two additions beyond the three, both judgment calls flagged here:**

1. `templates._fetch_template_workflow` — same function family, same file, the same unbounded read of the same remote host. Converting the gallery index but not the per-template workflow next to it would have been an odd place to stop.
2. `Client._request`'s **error**-body read (`e.read()`), three lines below the success read this PR caps. Same server, same vector. Over the cap the read raises and the pre-existing `except Exception: pass` leaves `body_text` empty — status and reason still reach the caller, which is the part that matters for a body too large to be a real error message.

**Three `except` tuples widened in `templates.py`.** These were narrower than the error families their callees actually raise, so a non-200 (`RuntimeError`) already escaped as an uncaught traceback before this change; `ResponseTooLarge` would have joined it. The narrow tuples all date to the file's original commit (#470); `_GALLERY_LOAD_ERRORS` was introduced later (#559) specifically to route this class of failure into an envelope, and that PR's own tests (`test_non_200_status_..._not_uncaught`) pin the intent. Widening to the same family completes that intent rather than reversing a decision — it is not a behavior flip on a deliberate guard.

## Why 64 MiB

Measured rather than guessed, so the refusal path cannot plausibly fire on legitimate traffic:

- The real gallery index (`Comfy-Org/workflow_templates/templates/index.json`) is **623 KB** — 0.9% of the cap.
- The largest of all **599** template workflow JSONs upstream is **1.09 MB**; *every* template JSON in that repo concatenated is 32.2 MB, still under the cap as a single body.
- The largest documented API body is `/api/object_info` at ~9 MB on cloud (per the existing note in `models/search.py`).

So ~60x headroom over the largest real gallery body and ~7x over the largest documented API body, while still bounding a server that streams without end. It also matches the cap the already-bounded readers (`models/search`, both `workflow` readers) converged on independently, so the CLI has one number rather than five. `cql.loader` keeps its own larger `MAX_INPUT_BYTES` — that bounds a user-supplied object_info dump, a different thing being measured.

## Scope line I did not cross

The task also mentioned "one JSON-parse failure type" for the primitive. **I did not add one**, deliberately: `http.py` already centralizes JSON parsing in `request_json` (empty or unparseable → `None`), and each of the three converted readers has a distinct parse-failure contract its callers depend on — `jobs` wraps it into `RuntimeError`, `templates` lets `JSONDecodeError` route through `_GALLERY_LOAD_ERRORS`, `comfy_client` propagates it. Preserving those was the explicit constraint, so a shared parse-failure type would have had no caller and shipped as dead code.

Converting the four already-capped readers (called out as optional and secondary) was also skipped — they are not broken, and three of the four already share `http.ResponseTooLarge` on `main`.

**Unbounded reads that remain, deliberately out of scope** (worth a follow-up; all are same-shape, none are the remote-gallery case this PR was filed for):

- `cloud/oauth.py:899` (token response) and `:902` (error body)
- `command/transfer.py:389` (JSON body)
- `command/run/execution.py:182` / `:184` (success + error body)
- `command/jobs.py:1194` / `:1226` (response drains), `:1283` (body), `:1285` (error body, sliced *after* an unbounded read)
- `command/models/search.py:173` and `cql/engine.py:1493` (error bodies, sliced after an unbounded read)
- `file_utils.py:332` (error body)

## Testing

`ruff check .`, `ruff format --check` on every changed file, and the full `pytest` suite: **3764 passed, 37 skipped**.

New coverage (23 tests): `read_capped` boundary behaviour (under / exactly-at / one-past the cap, message content, non-positive cap rejection, that the default is a real ceiling and the read asks for exactly `cap + 1`); and per call site, one test proving the read is issued *with* a cap and one proving an over-cap body lands in that reader's own error envelope rather than a traceback. The per-call-site tests run the real `read_capped` at a tiny cap through the real production path, so they exercise the actual routing without allocating 64 MiB.

Three pre-existing test fakes had `read(self)` with no size parameter, which does not match `http.client.HTTPResponse.read(amt)`; they now accept and honour one. The new fakes go further and **fail outright on `read(None)`**, so a future revert to an unbounded read breaks the test rather than passing quietly.

## Risk

The riskiest line is `raise HTTPError(resp.status, "response too large", str(e))` in `comfy_client._request` — it raises a 200-status error. It is safe because: (a) it is the module's own `HTTPError`, not `urllib.error.HTTPError`, so the surrounding handler does not swallow it; (b) status 200 is not retryable in `wait_for_completion` (`if e.status != 429 and not (500 <= e.status < 600): raise`) and is not 404, so `get_history` / `get_job_status` cannot silently convert it to `None`; (c) `submit_prompt` already establishes this exact idiom for a malformed 200.

On the capability-denial question: this diff adds a refusal path, but the capability being refused is "buffer an unbounded response body", and the numbers above are the empirical check that no real endpoint the CLI talks to produces one. It denies no product feature and adds no user-facing "not supported" path.

### DIFF
--- a/comfy_cli/comfy_client.py
+++ b/comfy_cli/comfy_client.py
@@ -269,7 +269,12 @@ class Client:
         except urllib.error.HTTPError as e:
             body_text = ""
             try:
-                body_text = e.read().decode("utf-8", errors="replace")
+                # An error body arrives from the same server as the success
+                # body, so it needs the same ceiling. Over-cap it raises, and
+                # the swallow below leaves body_text empty — the status and
+                # reason still reach the caller, which is the part that matters
+                # for a body too large to be a real error message.
+                body_text = read_capped(e, url).decode("utf-8", errors="replace")
             except Exception:  # noqa: BLE001
                 pass
             # Auto-refresh on 401 for OAuth cloud targets, retry once.
--- a/comfy_cli/comfy_client.py
+++ b/comfy_cli/comfy_client.py
@@ -23,7 +23,13 @@ from collections.abc import Callable
 from dataclasses import dataclass
 from typing import Any
 
-from comfy_cli.http import NoRedirectHandler, build_http_only_opener, target_auth_headers
+from comfy_cli.http import (
+    NoRedirectHandler,
+    ResponseTooLarge,
+    build_http_only_opener,
+    read_capped,
+    target_auth_headers,
+)
 from comfy_cli.http import assert_safe_url as _assert_safe_url
 from comfy_cli.target import Target
 
@@ -248,7 +254,15 @@ class Client:
             req.add_header(header, value)
         try:
             with _OPENER.open(req, timeout=timeout or self.timeout) as resp:
-                text = resp.read().decode("utf-8", errors="replace")
+                # Bounded read: without a ceiling the server on the other end
+                # decides how much of our memory to consume. An over-cap body
+                # is reported as a response error rather than truncated —
+                # truncated JSON would surface as a misleading parse failure.
+                try:
+                    raw = read_capped(resp, url)
+                except ResponseTooLarge as e:
+                    raise HTTPError(resp.status, "response too large", str(e)) from e
+                text = raw.decode("utf-8", errors="replace")
                 if not text:
                     return None
                 return json.loads(text)
--- a/comfy_cli/command/jobs.py
+++ b/comfy_cli/command/jobs.py
@@ -36,7 +36,7 @@ from websocket import WebSocket, WebSocketException, WebSocketTimeoutException
 from comfy_cli import cancellation, execution_errors, tracking
 from comfy_cli.env_checker import check_comfy_server_running
 from comfy_cli.host_port import resolve_host_port as _resolve_host_port
-from comfy_cli.http import authed_urlopen, plain_urlopen
+from comfy_cli.http import ResponseTooLarge, authed_urlopen, plain_urlopen, read_capped
 from comfy_cli.output import get_renderer
 from comfy_cli.where import cloud_preflight_or_exit
 
@@ -82,12 +82,21 @@ def _server_or_error(host: str, port: int, *, raise_on_missing: bool = True) -> 
 
 
 def _http_get_json(url: str, *, timeout: float = 10.0) -> Any:
+    """GET a JSON body from the local ComfyUI server.
+
+    The read is capped (``read_capped``) so a server streaming an endless
+    ``/history`` can't OOM the CLI. Every failure — unreachable, oversize,
+    non-JSON — leaves as a ``RuntimeError``, which is the single family every
+    call site below already catches.
+    """
     req = urllib.request.Request(url)
     try:
         with plain_urlopen(req, timeout=timeout) as resp:
-            return json.loads(resp.read())
+            return json.loads(read_capped(resp, url))
     except urllib.error.URLError as e:
         raise RuntimeError(f"failed to GET {url}: {e}") from e
+    except ResponseTooLarge as e:
+        raise RuntimeError(str(e)) from e
     except ValueError as e:
         # json.JSONDecodeError is a ValueError — a non-JSON 200 (captive
         # portal, proxy error page) must look like any other GET failure to
--- a/comfy_cli/command/templates.py
+++ b/comfy_cli/command/templates.py
@@ -30,7 +30,7 @@ from typing import Annotated, Any
 import typer
 
 from comfy_cli import tracking
-from comfy_cli.http import plain_urlopen
+from comfy_cli.http import ResponseTooLarge, plain_urlopen, read_capped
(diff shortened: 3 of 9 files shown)


bigcat88 (approved):
Approving. Measured the memory this actually bounds, rather than reasoning about it — the numbers are worse on `main` than the description implies.

### A 400 MiB body, same server, both trees

Drove the converted readers directly against a local server serving 400 MiB:

| reader | `main` | this branch |
| --- | --- | --- |
| `templates._fetch_gallery` | returns all **419,430,400 bytes** — peak RSS **596 MB** | `ResponseTooLarge` in 0.05s, peak RSS 117 MB |
| `jobs._http_get_json` | buffers *and then parses* it — peak RSS **1,251 MB** | `RuntimeError` in 0.04s, no growth |

The `jobs` row is the one worth calling out: the unbounded read is only half the cost, because `json.loads` on the buffered body roughly doubles it. A single 400 MiB response costs `main` **1.25 GB of resident memory**; here it costs 117 MB total, of which 64 MiB is the cap doing its job. And 400 MiB is just what I chose to serve — an endless body is unbounded on `main`.

### The `cap + 1` boundary is exactly as documented

| body | cap | result |
| --- | --- | --- |
| 1000 bytes | 1000 | OK, 1000 bytes returned |
| 1001 bytes | 1000 | `ResponseTooLarge` |
| — | 0 | `ValueError: max_bytes must be >= 1, got 0` |

So a body that exactly fills the cap really is distinguishable from a truncated one, and refusing beats truncating for exactly the reasons you give.

### Compatibility with #623, which is what I checked hardest

`request_json` is the piece already on `main` and in use, so its behaviour had to survive the extraction. It does — same exception type, same message text (the search envelope interpolates it), same early `max_bytes` validation before any network call. The 64 MiB constant matches what `models/search` and both `workflow` readers had converged on independently, so this collapses five numbers into one rather than introducing a new one.

One small thing I traced and cleared: the `ResponseTooLarge` raise moves from *outside* the `with` block to *inside* it. `with` still runs `__exit__` while unwinding, so the response is closed either way — no leaked connection, no behaviour change. Worth knowing it's deliberate rather than incidental.

Full suite on your branch merged with current `main`: **3786 passed, 31 skipped**, which includes #623's `test_http.py` — so nothing about the extraction disturbed the existing contract. `ruff check` / `ruff format --diff` clean at the CI-pinned 0.15.15. `command/jobs.py` also moved twice on `main` today (#662, #655) and the merged tree is clean.

### On the additions beyond the enumerated three

Both are right, and the error-body read is the sharper of the two — capping the success body while leaving `e.read()` unbounded three lines below would have left the same vector open behind a `4xx`. Good that status and reason still reach the caller when the body is refused; a body too large to buffer was never going to be a useful error message anyway.

The widened `except` tuples check out: a non-200 already escaped as an uncaught traceback before this PR, so `ResponseTooLarge` would have joined an existing hole rather than opening one. #559 introduced `_GALLERY_LOAD_ERRORS` precisely to route this family into an envelope — completing that is not reversing a deliberate guard.

PR #661, opened 2026-08-02
refactor: finish the atomic-writer consolidation and document the write policy

mattmillerai:
> **STACKED — merging lands on `matt/be-4353-atomic-json-write-util` (owned by @mattmillerai, PR #595), NOT `main`.** Review/merge #595 first; GitHub retargets this PR to `main` automatically once it lands.

## ELI-5

The CLI writes a bunch of small files — a job's status, a lock file listing pushed assets, a couple of caches. Writing a file isn't instant, so if the CLI is killed halfway through you can end up with a half-written file that nothing can read. The fix everyone uses is: write to a scratch file first, then rename it over the real one in a single step that can't be interrupted. Ten places in this codebase had each written that same dance by hand.

#595 built one shared helper and moved five of them onto it. This PR moves the last three, and — more importantly — writes down the *rule* for which ones are allowed to use the shared helper at all. Two writers deliberately stay hand-written because they can contain passwords or signed URLs, and they need stricter file permissions than the shared helper gives. Now there's a note in the code explaining exactly why, so the next person doing a cleanup pass doesn't "helpfully" consolidate a secrets file into the generic path.

## What changed

**New helper.** `atomic_write_bytes(path, data, *, fsync=False)` joins `atomic_write_text`. Both delegate to one private `_atomic_write` that operates on bytes; the text variant encodes UTF-8 and forwards. No `mode=` parameter — that was considered and rejected, because a writer whose payload needs 0600 needs more than a mode (an `O_EXCL` open *at* that mode, a 0700 parent, an fsync), and that's clearer spelled out at its own call site.

**Three migrations.**

| Writer | Now | Note |
|---|---|---|
| `command/project.py` `_write_assets_lock` | `atomic_write_text(..., fsync=True)` | Net gain: tmp-file cleanup on failure, which the inline version lacked. Dropped the docstring sentence admitting it re-implemented the jobs_state pattern. |
| `command/outdated.py` `_save_cache` | `atomic_write_text(..., fsync=False)` | Stays inside the existing best-effort `try/except OSError: pass`; the helper's `mkdir` replaces the explicit `path.parent.mkdir`. |
| `command/templates.py` `_persist_cache` | `atomic_write_bytes(..., fsync=False)` | Keeps its best-effort wrapper and docstring. |

**The write policy**, as a comment block adjacent to the helpers in `file_utils.py`: four tiers (secrets → bespoke 0600; secret-adjacent state → bespoke 0600; cross-process state → `fsync=True`; regenerable caches → `fsync=False`), each naming its members, plus the two recorded invariants (why `JobState` files are umask-mode, and why nobody fsyncs the parent directory after `os.replace`). `auth/store._write_all` and `download_state.write_path` each gained a one-line pointer at that note so the next dedup sweep doesn't re-flag them.

## Permissions

No writer's permissions change, with one deliberate exception called out in the ticket and in the code: the templates gallery cache. Its old inline `tempfile.mkstemp` left it at mode 0600 incidentally (mkstemp hardcodes 0600 and `os.replace` carries that onto the destination); the shared helper restores the umask-derived mode instead. That relaxation is intended — the payload is the public template-gallery index, not user data. There's a test pinning the new mode.

## Judgment calls

**This is stacked on #595, not a rebase-and-land of it.** The ticket's precondition said to rebase #595 onto `main`, resolve the conflict, and land it first. I resolved the conflict but could not land it (merging is not mine to do), and rebasing would have required a force-push. So instead I pushed the conflict resolution to #595's branch as an ordinary merge commit — a fast-forward, no history rewritten — which took #595 from `CONFLICTING` to `MERGEABLE`, and stacked this work on top. That also keeps this PR's diff to just my own commit rather than double-counting all of `main`. The conflict itself was one import block in `command/workflow.py` where both sides had added an import; resolved by keeping both.

**The shared implementation writes in binary mode.** `atomic_write_text` now encodes UTF-8 itself and the shared `_atomic_write` opens `"wb"`, where it previously opened `"w", encoding="utf-8"`. On Windows that drops the automatic `\n` → `\r\n` translation, so files written by all seven text call sites now get LF line endings there instead of CRLF. This is safe and I'd argue more correct: every reader of these files goes through `read_text()` or `json.loads()`, both of which normalize newlines, so the round-trip is unchanged; no test asserts exact bytes for any of them; and it makes the two variants byte-identical in behavior. Flagging it explicitly because I can't exercise Windows locally — the `windows-latest` CI leg covers it.

**Two existing tests were rewritten, not deleted.** Both asserted the *old implementation*, not behavior, and could not survive the migration:

- `test_save_cache_pins_the_write_encoding` spied on `Path.write_text` to check `encoding="utf-8"` was named. The helper doesn't use `Path.write_text`. Rewritten to assert the same contract one layer down and rather more directly — that the bytes reaching the filesystem are the payload encoded UTF-8 *by us*, so the host locale never gets a vote.
- `test_readonly_cache_dir_still_serves_fetched_data` patched `templates_cmd.tempfile.mkstemp` to force a write failure; `templates.py` no longer imports `tempfile` at module scope. Repointed at `file_utils.tempfile.mkstemp`, which keeps the test end-to-end through the real `_persist_cache` → `atomic_write_bytes` path rather than stubbing the helper out.

**Negative-claim falsification: not triggered.** This diff denies no capability — it adds no `not supported` / `unavailable` / `STOP` strings, no throw/deny dead-end, and flips no test to assert a dead-end. It is a pure consolidation: every migrated writer keeps its existing success and failure behavior, and one (`_write_assets_lock`) gains cleanup it previously lacked.

## Testing

- `pytest`: **3756 passed, 37 skipped** (full suite).
- `ruff check .` and `ruff format --check .`: clean, run against **v0.15.15**, the version pinned in `.pre-commit-config.yaml`. Worth knowing for anyone reproducing locally: a plain `uv sync` resolves ruff 0.12.7, which reports 17 pre-existing errors and 1 format diff on `main` untouched by this PR — version skew, not real findings.
- New `atomic_write_bytes` unit tests mirroring the text-variant cases: create-with-parents, overwrite, verbatim payload (non-UTF-8 bytes and bare LFs survive), fsync path, cleanup-with-destination-untouched on a failed rename, and the umask-default mode that the templates cache now relies on.


bigcat88:
Your base, #595, is now merged — I reviewed it in depth and landed it (`3779 passed`, mode parity with `main` verified on every migrated writer, atomicity exercised under 6 concurrent writers and 20 mid-write `SIGKILL`s).

That squash-merge is exactly what has flipped this PR to `CONFLICTING`: GitHub retargeted it to `main`, and the squashed commit no longer matches the merge-commit history this branch was built on. Not your fault and nothing wrong with the change — it just needs a rebase onto current `main` before I can review it.

Two things from the #595 review that are worth folding into this PR while you're in there, since it's the one that writes down the policy:

- **`os.umask(0)` / `os.umask(umask)` in `_atomic_write` is process-global and not thread-safe** — a thread creating a file inside that two-syscall window gets a 0 umask. Unreachable today (all writers are on the main thread; the watcher is a separate process), so a note in the comment is enough. Your write-policy block is the natural home for it.
- **Stranded tmp files are now unbounded rather than bounded by pid reuse.** `mkstemp` gives a fresh name per attempt, so every crash mid-write leaves a distinct corpse; I measured 13 survivors from 20 kills. Nothing prunes the jobs directory and `jobs ls` globs `*.json`, so they accumulate invisibly. A `*.tmp` sweep alongside the stale-watcher reap in `_gather_local_state_files` would close it.

Ping me once it's rebased and I'll pick it straight up — the `atomic_write_bytes` + write-policy work reads well from the description.

mattmillerai:
- [BE-6392](https://linear.app/comfyorg/issue/BE-6392) — Harden mode handling for tier-4 atomic_write_* cache files against local tampering — filed as `agent-spike` (premise unverified)

The following carry `agent-spike` instead of `agent-ok` because their reachability claim was not backed by evidence (BE-5378) — the claim is investigated before any code is written, and "the premise does not hold" is a valid, successful outcome:
- **Harden mode handling for tier-4 atomic_write_* cache files against local tampering** — no reachability block in the proposal

skishore23 (approved):
Approving the change — the security-critical property holds and I verified it at runtime. **But please rebase rather than merge**: the merge produces a silent semantic break that git does not report as a conflict.

#595 is **merged**, so the stacking banner is stale and a clean rebase onto `main` is now possible (the force-push concern that drove the merge-commit approach no longer applies).

## ⚠️ Merging `main` silently duplicates `atomic_write_text`

```
PR head (pre-merge):  atomic_write_text:74   atomic_write_bytes:99   _atomic_write:109
origin/main:          atomic_write_text:20
after `git merge`:    atomic_write_text:74   atomic_write_bytes:99   _atomic_write:109
                      atomic_write_text:164   <-- duplicate, NOT flagged as a conflict
```

`file_utils.py` is **not** in the conflicted set (only `command/templates.py`, `test_templates.py`, `test_file_utils.py` are) — because #595's copy landed at a different offset, git treats the two as independent additions and keeps both.

The consequence is worse than a lint nit: the second definition **shadows** the first, so `atomic_write_text` resolves to `main`'s older non-delegating version and never reaches `_atomic_write`. That silently reverts half the consolidation this PR exists to do. Observed downstream:

- `ruff` → `F811 Redefinition of unused atomic_write_text from line 74` (so CI *would* catch it — good)
- `test_save_cache_pins_the_write_encoding` → `KeyError: 'data'`, because the spy sits on the seam the shadowed version doesn't use

A rebase onto `main` avoids this entirely, since #595's commits are already upstream. (I tried excising the duplicate by hand to confirm the rest is clean and overshot into neighbouring functions — that part is my mess, not a finding. The duplication itself is real and reproducible.)

The other three conflicts are trivial import/addition unions:
```python
from comfy_cli.file_utils import atomic_write_bytes
from comfy_cli.http import ResponseTooLarge, plain_urlopen, read_capped
```

## The security-critical property holds — verified at runtime

This is the part I most wanted to check, since a consolidation PR is exactly how a secrets file quietly acquires world-readable permissions:

```
shared helper  generic.json  mode = 0644   (umask-derived)
shared helper  generic.bin   mode = 0644
auth store     creds.json    mode = 0600   ✓
auth store     parent dir    mode = 0700   ✓
group/other bits on secrets file: none
```

And both secrets writers carry the pointer, not a call:
```
auth/store.py:115      Deliberately not ``file_utils.atomic_write_text`` — see the write-policy note
download_state.py:218  Deliberately not ``file_utils.atomic_write_text`` — see the write-policy note
```

That's the actual deliverable here. The 0644-vs-0600 gap is real and one line of "helpful" consolidation away from leaking a Civitai/HF token to every user on a shared box — so the in-code note plus the two pointers are worth more than the three migrations. Good instinct making the *policy* the artifact rather than just finishing the sweep.

Rejecting a `mode=` parameter on the helper is right for the same reason: a 0600 payload needs `O_EXCL` **at** that mode, a 0700 parent, and an fsync — a mode kwarg would look sufficient while leaving the tmp file briefly world-readable.

## Other notes

- The templates-cache relaxation (incidental `mkstemp` 0600 → umask) is fine — it's the public gallery index — and it's pinned by `test_atomic_write_text_new_file_uses_umask_default`, so the change is deliberate rather than drift.
- **The binary-mode newline change is the right call and correctly flagged.** Dropping `\r\n` translation on Windows is safe because every reader goes through `read_text()`/`json.loads()`, both of which normalize — and `windows-latest` covers it. Worth keeping that paragraph in the squash message.
- Rewriting the two implementation-coupled tests rather than deleting them was correct; both now assert the contract a layer down instead of the mechanism.

mattmillerai:
- [BE-6679](https://linear.app/comfyorg/issue/BE-6679) — Sweep stranded *.tmp files left by crashed atomic_write_* writers — filed as `agent-spike` (premise unverified)

The following carry `agent-spike` instead of `agent-ok` because their reachability claim was not backed by evidence (BE-5378) — the claim is investigated before any code is written, and "the premise does not hold" is a valid, successful outcome:
- **Sweep stranded *.tmp files left by crashed atomic_write_* writers** — no reachability block in the proposal

PR #679, opened 2026-08-05
fix: stop an explicit --port 0 being swallowed by --host h:p

mattmillerai:
## ELI-5

`comfy run` lets you write the address two ways: `--host 127.0.0.1 --port 9100`, or all in one go as `--host 127.0.0.1:9100`. When you used both at once, the code picked between them with "if the `--port` is empty, use the one from the host string" — and Python thinks the number `0` is empty. So `comfy run --host 127.0.0.1:9100 --port 0` quietly threw your `--port 0` away and ran against 9100, instead of telling you that 0 isn't a real port. Now it checks "was `--port` actually given?" rather than "is it truthy", so a port you typed always wins, and a nonsense one gets a clean error instead of a silent surprise.

## What changed

- `comfy_cli/cmdline.py` (`comfy run`) and `comfy_cli/command/templates.py` (`comfy run-template`): `if not port and parsed_port is not None` → `if port is None and parsed_port is not None`, with a comment explaining the precedence contract.
- `comfy_cli/host_port.py`: a docstring note that `resolve_host_port` owns the range check for the separate `--port` flag on behalf of every caller, so call sites only need to keep an explicit port distinguishable from an absent one.
- Tests in `tests/comfy_cli/test_host_port.py` (`comfy run`, `comfy jobs status`) and `tests/comfy_cli/command/test_run_template.py` (`comfy run-template`).

The range guard itself (`validate_port`, called from `resolve_host_port`) already landed in #667, along with its unit tests and the `port is None` form at the third combined-host site (`comfy validate`). This PR is the remaining half: the two sites that still swallowed an explicit `0`.

## Behavior

| command | before | after |
|---|---|---|
| `comfy run --host 127.0.0.1:9100 --port 0` | ran against **9100** | exit 2, `invalid port: 0 is out of range (1-65535)` |
| `comfy run-template <name> --host 127.0.0.1:9100 --port 0` | ran against **9100** | exit 2, same error |
| `comfy run --host 127.0.0.1:9100 --port 9200` | ran against 9200 | unchanged |
| `comfy run --host 127.0.0.1:9100` | ran against 9100 | unchanged |
| `comfy jobs … --port 0` / `--port 70000` | already exit 2 (via #667) | unchanged — covered by a new test so the resolver placement stays load-bearing |

The only input whose behavior changes is an explicit `--port 0` alongside a combined `--host h:p`. Every `--port` param on these paths is typed `int | None` defaulting to `None`, so no caller can reach the branch with a non-`None` "unset" value; negative and out-of-range values were already truthy and so already reached the guard.

## Falsification of the deny path

This diff makes a previously-accepted invocation fail, so before shipping it I checked empirically that no working capability is being removed. Against `main` (`5798668`), with `comfy_cli.command.run.execute` stubbed:

```
comfy run --workflow wf.json --host 127.0.0.1:9100 --port 0
  -> exit 0, execute() called with host/port ('127.0.0.1', 9100)
```

So `--port 0` was never honored — it was discarded, and the run went to the embedded port. There is no "connect to port 0" behavior to preserve (port 0 is not a connectable TCP destination; it is only meaningful for `bind`), and `comfy launch` — the one command that binds rather than connects — does not route through `resolve_host_port` at all (`git grep resolve_host_port` shows only the run/jobs/validate/nodes/templates client paths). `--port 0` on its own already errored on `main` via #667; this only makes the combined-host path agree.

## Testing

- Red→green: with the source change stashed, exactly the two new `--port 0` tests fail (`assert 0 == 2` — the pre-fix run reaching the stubbed execute path); with it applied, all pass.
- `pytest tests/comfy_cli/test_host_port.py tests/comfy_cli/test_local_address.py tests/comfy_cli/command/test_run_template.py tests/comfy_cli/command/test_templates.py tests/comfy_cli/command/test_nodes_cli.py tests/comfy_cli/command/test_run_cli.py tests/comfy_cli/command/test_run.py tests/comfy_cli/jobs` → **504 passed**.
- `ruff check .` and `ruff format --check .` → clean.

## Judgment calls

- **The whole-repo `pytest` run was not completed locally.** It exceeded a 10-minute foreground budget, so I scoped verification to every suite that touches the changed code paths (the 504 above) plus lint/format, and left the full run to CI. Flagging it rather than claiming a green full suite.
- **The templates test lives in `tests/comfy_cli/command/test_run_template.py`, not `tests/comfy_cli/test_host_port.py`.** The gallery-index fixture, workflow-fetch stub and run spy that a `run-template` CLI test needs are ~100 lines already built there; duplicating them into `test_host_port.py` to satisfy file placement seemed strictly worse. The `run` and `jobs` CLI tests are in `test_host_port.py` as intended.
- **The command is `comfy run-template <name>`, not `comfy templates run <name>`** — `run_template_cmd` is registered on the top-level app, not on the `templates` sub-app. Same code path either way; noting it because the description says otherwise.
- I also added the positive-direction assertions (`--port 9200` beats an embedded `:9100`) at both sites. They pass before and after, but they are what pins the precedence contract the `port is None` form exists to express.

### DIFF
--- a/comfy_cli/cmdline.py
+++ b/comfy_cli/cmdline.py
@@ -1003,7 +1003,11 @@ def run(
 
         if host:
             host, parsed_port = parse_host_port_arg(host)
-            if not port and parsed_port is not None:
+            # ``port is None``, not ``not port``: a typed ``--port`` always wins
+            # over one embedded in ``--host h:p``, including ``--port 0``, which
+            # ``resolve_host_port`` then rejects as out of range instead of
+            # silently running against the embedded port.
+            if port is None and parsed_port is not None:
                 port = parsed_port
 
         host, port = resolve_host_port(host, port)
--- a/comfy_cli/command/templates.py
+++ b/comfy_cli/command/templates.py
@@ -1478,7 +1478,11 @@ def run_template_cmd(
 
     if host:
         host, parsed_port = parse_host_port_arg(host)
-        if not port and parsed_port is not None:
+        # ``port is None``, not ``not port``: a typed ``--port`` always wins over
+        # one embedded in ``--host h:p``, including ``--port 0``, which
+        # ``resolve_host_port`` then rejects as out of range instead of silently
+        # running against the embedded port.
+        if port is None and parsed_port is not None:
             port = parsed_port
     host, port = resolve_host_port(host, port)
 
--- a/comfy_cli/host_port.py
+++ b/comfy_cli/host_port.py
@@ -145,6 +145,11 @@ def resolve_host_port(host: str | None, port: int | None) -> tuple[str, int]:
     An explicitly-passed ``host``/``port`` is validated *before* the precedence
     chain runs, because that chain treats every falsy value as "not passed" and
     would otherwise swallow the bad input and resolve to a different server.
+    This function owns the range check for the separate ``--port`` flag on
+    behalf of every caller (``run``, ``templates run``, ``jobs``, ``validate``,
+    ``nodes``), so call sites must not re-implement it — they only need to keep
+    an explicit ``--port`` distinguishable from an absent one (``port is None``,
+    never ``not port``) so a ``--port 0`` reaches this guard.
     """
     from comfy_cli.env_checker import _bracket_host
     from comfy_cli.local_address import resolve_local_host_port
--- a/tests/comfy_cli/command/test_run_template.py
+++ b/tests/comfy_cli/command/test_run_template.py
@@ -12,6 +12,7 @@ from __future__ import annotations
 import json
 from pathlib import Path
 from typing import Any
+from unittest.mock import MagicMock
 
 import pytest
 import typer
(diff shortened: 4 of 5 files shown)


bigcat88 (approved):
Approving. I reproduced the swallow black-box on `main` rather than trusting the stub, and every row of your table holds:

```
main    run --host 127.0.0.1:9100 --port 0     -> rc=1  server_not_running (127.0.0.1:9100)   <- discarded
this PR run --host 127.0.0.1:9100 --port 0     -> rc=2  invalid port: 0 is out of range
both    run --host 127.0.0.1:9100 --port 9200  -> rc=1  server_not_running (127.0.0.1:9200)
both    run --host 127.0.0.1:9100              -> rc=1  server_not_running (127.0.0.1:9100)
both    --port 0 alone / 70000 / -1            -> rc=2
both    validate | jobs status | nodes ls --port 0 -> rc=2
```

The `9100` in that first row is the whole bug in one line — the flag the user typed lost to the one they embedded.

**Site coverage is complete.** `parse_host_port_arg` has exactly three call sites: `cmdline.py:1005` (run), `cmdline.py:1130` (validate, already `port is None` from #667), `templates.py:1480` (run-template). This PR fixes the two remaining ones and there is no fourth. I also confirmed the claim that binding is unaffected — `resolve_host_port` is reached only from run/validate/templates/jobs/nodes, and `comfy launch` doesn't appear in that list, so the one command where port 0 *would* be meaningful never routes through this guard.

**Red→green independently reproduced.** With `comfy_cli/cmdline.py` and `comfy_cli/command/templates.py` reset to `origin/main` and your tests left in place, exactly the two `--port 0` tests fail:

```
FAILED test_host_port.py::test_run_explicit_port_zero_is_rejected_not_overridden_by_host_port
FAILED test_run_template.py::test_explicit_port_zero_is_rejected_not_overridden_by_host_port
2 failed, 89 passed
```

The two precedence assertions pass either way, exactly as you said — they pin the contract rather than the fix.

**Full suite, since you flagged it as unfinished:** I ran it on your branch merged with current `main` — **4136 passed**, 31 skipped. The single failure is `test_non_fast_deps_uses_global_python`, which fails identically on `main` and is unrelated to this change. No conflict with `main` either; the only thing that landed since your merge-base is #680 in `cql/engine.py`, which this doesn't touch.

One consequence worth stating out loud since the PR description doesn't: under `--json`, this input used to produce a real envelope (`{"ok":false,"error":{"code":"server_not_running"}}`) and now produces none — `typer.BadParameter` writes a usage block to stderr with rc=2 and leaves stdout empty. That is the right outcome, and I checked it is byte-for-byte the shape `--host ''` and `--port 70000` already produce (stdout 0 bytes, stderr 641), so it joins the established family rather than inventing a new one. stdout stays clean, so nothing in the envelope contract is violated — just noting that a JSON consumer feeding an invalid port gets a usage error, not an envelope, on all of these paths alike.

mattmillerai:
- [BE-6656](https://linear.app/comfyorg/issue/BE-6656) — Emit a terminating NDJSON envelope for --host/--port usage errors under --json — filed as `agent-spike` (premise unverified)

The following carry `agent-spike` instead of `agent-ok` because their reachability claim was not backed by evidence (BE-5378) — the claim is investigated before any code is written, and "the premise does not hold" is a valid, successful outcome:
- **Emit a terminating NDJSON envelope for --host/--port usage errors under --json** — no reachability block in the proposal

PR #680, opened 2026-08-05
fix(cql): report COMBO fields with zero installed options instead of skipping them (BE-6585)

mattmillerai:
## ELI-5

`comfy validate` checks a dropdown value (a model name, an image name) against the list of options the server reports for that field. If the list was **empty**, the old code read that as "this field has no rules" and skipped the check entirely — so on a fresh install with zero models, the loaders that ship no built-in options (`UNETLoader`, `CLIPLoader`) said nothing at all, while a loader that happens to ship one built-in (`VAELoader`) correctly reported its missing model. Detection got *worse* the emptier the install, and the two silently-skipped loaders are the biggest downloads — exactly what a user needs told before starting a run. An empty option list now says "0 options installed for this field" instead of saying nothing.

## What changed

- `Port.validate_catalog` — a `COMBO` with a **declared but empty** option list and a supplied value now emits a new `no_options_available` finding instead of falling through the truthiness guard at the old `engine.py:98`.
- `_validate_catalog_value` promotes it to a hard error (same tier as `unknown_enum_value`, for the same reason: the server rejects it), with `field`, `node_id`, an install hint, and `valid_options: []` for parity with the enum error. The slot-editing path (`apply_slots` / `set-slot`) still surfaces it as a soft warning, unchanged in shape.
- New `Port.enum_declared` flag, set by `_parse_input_spec`, separates *"the server declared this field's choices and shipped none"* from *"this field's choices aren't in object_info at all"*.

## Why a new flag rather than just dropping the `and self.enum_values` guard

Because the two empty cases are not the same statement, and only one of them is safe to validate:

| object_info spec | meaning | new behavior |
| -- | -- | -- |
| `[[]]` (list-form, empty) | `folder_paths.get_filename_list(...)` returned nothing — no files installed | flagged |
| `["COMBO", {"options": []}]` | declared, explicitly empty | flagged |
| `["COMBO", {"remote": {...}}]` | options fetched by the frontend at runtime; `prune_dict` drops the absent `options` key (`comfy_api/latest/_io.py`) | **silent** (unconstrained, as today) |

`enum_declared` is deliberately *not* folded into `is_enum`, because `is_enum` also decides link-vs-widget wiring (`_is_link`) — an empty option list must not move a port between those. `is_enum` and `enum_values` come out of the parser byte-identical to before; the only new information is the flag. There is a test pinning that (`test_empty_combo_is_still_a_widget_not_a_link`).

## Evidence the server really does reject these (negative-claim falsification)

The diff makes `validate` deny something it used to pass, so the premise was checked against the product rather than only against my own tests:

- **ComfyUI `execution.py:1035-1067`** (local checkout at `c68789b0`): for both combo dialects — `isinstance(input_type, list)` and `input_type == io.Combo.io_type` with `combo_options = extra_info.get("options", [])` — the check is `val not in combo_options`. Against an empty list that is true for **every** value, producing `value_not_in_list`. So the run genuinely fails; the old silence just deferred the failure to submit time.
- **Counter-experiment (the one the ticket asked for), at the unit level:** `test_same_loader_with_one_option_installed_flags_membership` gives the same loader one installed option and the same missing model is then caught by the existing membership path — proving the mechanism was the empty list, not the loader. I did **not** run this against a live ComfyUI server with a file dropped into `models/unet/`; the evidence above is source-level plus unit-level.

## Known trade-off (pre-existing, not introduced here)

The server skips its built-in combo check for inputs covered by a node's own `VALIDATE_INPUTS` (`execution.py:1007`) — `LoadImage.image` is the core example, which accepts any annotated path that exists on disk. Such a value can pass on the server while `validate` flags it. That is the same over-strictness `unknown_enum_value` and the min/max errors already accept and document in this file, and `object_info` does not expose whether a node defines `VALIDATE_INPUTS`, so it cannot be narrowed from here. The empty case inherits it rather than adding it.

## Judgment calls

- **Distinct code, not `unknown_enum_value` with an empty `valid_options`** — the ticket offered both. The distinct code lets a caller say "no models installed for this loader" instead of "value not in 0 options", and `valid_options: []` is still present so a consumer keyed on the enum shape keeps working.
- **Error, not warning.** It matches `unknown_enum_value`, and the server's own verdict: the prompt will be rejected. It does mean `comfy validate` / `comfy run` preflight now exits non-zero on a fresh install for workflows that previously passed with silent, unrunnable inputs — which is the point of the parent ticket.
- Validation item codes are not registered in `error_codes.py` (that file is the envelope-level `error.code` contract), so `no_options_available` needed no registration — consistent with `unknown_enum_value`.

## Testing

`ruff check .`, `ruff format --check .` clean; full `pytest` green (4126 passed, 37 skipped). Eight new tests in `TestValidateEmptyCombo` cover: the empty loader being flagged, both loaders reported together (the ticket's 2-of-4 count bug), the counter-experiment, no false positive on a populated field, the dict-form empty case, remote combos staying silent, widget-vs-link invariance, and no double-report against `required_input_missing`.

### DIFF
--- a/comfy_cli/cql/engine.py
+++ b/comfy_cli/cql/engine.py
@@ -49,6 +49,13 @@ class Port:
     required: bool = False
     is_link: bool = False
     enum_values: list[Any] = field(default_factory=list)  # preserves the option's real type (int combos stay int)
+    # True when object_info shipped an explicit choice list for this input —
+    # *including an empty one*. An empty declared list means "the server has
+    # zero options here" (a model folder with no files installed), which is a
+    # different statement from "the choices aren't in object_info at all"
+    # (remote/dynamic combos, whose options the frontend fetches at runtime).
+    # Only the former is safe to validate against.
+    enum_declared: bool = False
     options: PortOptions = field(default_factory=PortOptions)
     # The verbatim ``INPUT_TYPES`` spec this port was parsed from. Retained
     # (by reference — no copy) because a dynamic combo's sub-input schema lives
@@ -132,6 +139,26 @@ class Port:
                         "valid_options": list(self.enum_values),
                     }
                 )
+        elif self.type == "COMBO" and self.enum_declared:
+            # The server declared this field's choices and shipped NONE of them:
+            # the folder backing it is empty — no models (or inputs) installed.
+            # An empty option list is STRONGER evidence the value is unavailable
+            # than a populated one, not weaker: the server rejects every value
+            # against an empty list (`execution.validate_inputs` →
+            # "Value not in list"), so skipping the check here just defers the
+            # failure to run time — and it fails hardest on the fresh install
+            # this check exists to serve, where the big downloads (UNETLoader,
+            # CLIPLoader) have no built-in options to compare against.
+            # Ports whose options are not in object_info at all keep
+            # `enum_declared=False` and stay unconstrained (see the field).
+            warnings.append(
+                {
+                    "code": "no_options_available",
+                    "field": self.name,
+                    "message": f"{value!r} is unavailable: the server reports 0 installed options for {self.name}",
+                    "valid_options": [],
+                }
+            )
         if self.type in ("INT", "FLOAT", "NUMBER") and isinstance(value, int | float):
             if self.options.min is not None and value < self.options.min:
                 warnings.append(
@@ -258,13 +285,21 @@ def _control_after_generate_set(val: Any) -> bool:
     return True
 
 
-def _parse_input_spec(spec: Any) -> tuple[str, bool, list[Any], PortOptions]:
-    """Returns (type_id, is_enum, enum_values, options)."""
+def _parse_input_spec(spec: Any) -> tuple[str, bool, list[Any], PortOptions, bool]:
+    """Returns (type_id, is_enum, enum_values, options, enum_declared).
+
+    ``enum_declared`` is True when the spec shipped an explicit choice list —
+    *including an empty one* — so validation can tell "this server has zero
+    options for the field" from "this field's options aren't in object_info".
+    It is deliberately separate from ``is_enum`` (which stays truthiness-based)
+    because ``is_enum`` also decides link-vs-widget, and an empty option list
+    must not move a port between those.
+    """
     if isinstance(spec, str):
-        return spec, False, [], PortOptions()
+        return spec, False, [], PortOptions(), False
 
     if not isinstance(spec, list) or len(spec) == 0:
-        return "UNKNOWN", False, [], PortOptions()
+        return "UNKNOWN", False, [], PortOptions(), False
 
     opts_raw = spec[1] if len(spec) > 1 and isinstance(spec[1], dict) else {}
     port_opts = _parse_port_options(opts_raw)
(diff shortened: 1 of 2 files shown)


bigcat88 (approved):
Approving — I verified the negative claim against a **live ComfyUI server** rather than only the source you cited, and the premise holds exactly.

**The server really does reject it.** Wired `UNETLoader` into a full graph through to `PreviewImage` (an unconnected node is never validated — the prompt short-circuits at `prompt_no_outputs` first, which is easy to mistake for a pass) and POSTed to `/prompt`:

```
node1 value_not_in_list [unet_name: 'flux1-dev.safetensors' not in []]
```

So the old silence deferred a *guaranteed* failure to submit time. `LoadAudio` (dict-form empty) is rejected too.

**Census on a real, well-stocked install** (792 classes, checkpoints/vae/clip all mounted — so this is not just the fresh-install case): 21 empty combos against 481 populated ones. Both dialects occur in the wild, and `LoadImageOutput` is a live example of the remote form — no `options` key, correctly stays silent. Your `prune_dict` reasoning is right for the right reason: it is `{k: v for k, v in d.items() if v is not None}`, so it drops `remote`'s absent options but genuinely emits `options: []`.

I also confirmed the parser invariance you claimed: for the dict-form empty case `all([])` is vacuously True but `bool([])` is False, so `is_enum`/`enum_values` come out `False, []` — byte-identical to the old truthiness guard.

**The trade-off you documented, quantified.** Of the 21 flagged fields, 17 are genuinely rejected by the server. Three are not — their custom validation makes `execution.py:1007` skip the combo check entirely:

| node | validator |
| -- | -- |
| `LoadLatent.latent` | `VALIDATE_INPUTS(latent)` |
| `LoadAudio.audio` | `validate_inputs(audio)` (V3) |
| `LoadVideo.file` | `validate_inputs(file)` (V3) |

These build their lists with a non-recursive `os.listdir`, but validate with `exists_annotated_filepath`, which resolves subfolders and `[output]`/`[temp]` annotations. Demonstrated end-to-end — planted `input/sub/probe.latent`, list stays `[[]]`:

```
server:  POST /prompt {latent: "sub/probe.latent"}  -> HTTP 200 ACCEPTED
main:    validate -> valid=True   rc=0
this PR: validate -> valid=False  rc=1  no_options_available
```

Not blocking, because it is exactly the class you called pre-existing and I confirmed that: give `LoadLatent` one installed option and `main` **already** hard-errors the same value with `unknown_enum_value`. This extends an accepted over-strictness to the empty case rather than inventing a new one. Worth remembering as the first thing to narrow if the file-loader nodes ever draw complaints — the rate on those three goes from occasional to every value, since an empty list matches nothing.

**Cloud is unaffected.** `_preflight_validate` on the cloud path is fed `_load_from_target(mode="cloud")`, so a bare local install can't false-flag a cloud run. Fetched staging's `object_info` (3610 classes) to be sure the per-job-models carve-out at `preflight.py:152` isn't undermined: 13 empty combos, none of them checkpoint or unet loaders — no wave of new cloud failures.

`set-slot` is unchanged in shape as you said — soft warning, `ok=true`, value still applied:

```json
{"code":"no_options_available","field":"unet_name","message":"'other.safetensors' is unavailable: the server reports 0 installed options for unet_name","valid_options":[]}
```

Suite green at 4131 passed. Two failures I hit were pre-existing on `main` and unrelated: `test_non_fast_deps_uses_global_python`, and `test_object_info_timeout_routes_to_connection_error` — the latter only fails when a real ComfyUI is listening on `:8188`, because something outside the patched `_AUTHED_OPENER` reaches the live server and the injected `TimeoutError` never surfaces. Worth an isolation fix on `main` sometime; nothing to do with this PR.

Issue #163, opened 2024-08-26
ERRO : "A module that was compiled using NumPy 1.x cannot be run"


Lilien86:
##When launching ComfyUI, the following error occurs:

```
A module that was compiled using NumPy 1.x cannot be run in
NumPy 2.1.0 as it may crash. To support both 1.x and 2.x
versions of NumPy, modules must be compiled with NumPy 2.0.
Some module may need to rebuild instead e.g. with 'pybind11>=2.12'.

If you are a user of the module, the easiest solution will be to
downgrade to 'numpy<2' or try to upgrade the affected module.
We expect that some modules will need time to support NumPy 2.

Traceback (most recent call last):  File "D:\Comfyui\cli\comfyui_cli\main.py", line 83, in <module>
  import comfy.utils
File "D:\Comfyui\cli\comfyui_cli\comfy\utils.py", line 20, in <module>
  import torch
File "D:\Comfyui\cli\comfy-env\Lib\site-packages\torch\__init__.py", line 2120, in <module>
  from torch._higher_order_ops import cond
File "D:\Comfyui\cli\comfy-env\Lib\site-packages\torch\_higher_order_ops\__init__.py", line 1, in <module>
  from .cond import cond
File "D:\Comfyui\cli\comfy-env\Lib\site-packages\torch\_higher_order_ops\cond.py", line 5, in <module>
  import torch._subclasses.functional_tensor
File "D:\Comfyui\cli\comfy-env\Lib\site-packages\torch\_subclasses\functional_tensor.py", line 42, in <module>
  class FunctionalTensor(torch.Tensor):
File "D:\Comfyui\cli\comfy-env\Lib\site-packages\torch\_subclasses\functional_tensor.py", line 258, in FunctionalTensor
  cpu = _conversion_method_template(device=torch.device("cpu"))
D:\Comfyui\cli\comfy-env\Lib\site-packages\torch\_subclasses\functional_tensor.py:258: UserWarning: Failed to initialize NumPy: _ARRAY_API not found (Triggered internally at C:\actions-runner\_work\pytorch\pytorch\builder\windows\pytorch\torch\csrc\utils\tensor_numpy.cpp:84.)
  cpu = _conversion_method_template(device=torch.device("cpu"))
Total VRAM 8188 MB, total RAM 32472 MB
```

###Explination :
[...]

PR #690, opened 2026-08-06
feat: derive comfy_usage_source from caller detection (BE-6800)

mattmillerai:
## ELI-5

Every workflow the CLI submits carries a little label saying "this came from comfy-cli", which is what the server bills and attributes partner-node (BFL, Gemini, …) usage against. That label was hardcoded, so a run you typed by hand and a run an AI agent kicked off through comfy-mcp looked identical. The CLI already knows which one it is — it detects its caller to pick JSON-vs-pretty output — so this just feeds that answer into the label. Humans keep the plain `comfy-cli`; agents get `comfy-cli/comfy-mcp`, `comfy-cli/claude-code`, `comfy-cli/agent`, `comfy-cli/pipe`.

## What changed

`comfy_cli/caller.py` grows the derivation (`usage_source_for` + a memoised `usage_source()`), and the two attribution call sites use it instead of the literal:

- `comfy_cli/comfy_client.py` — the cloud/unified submit path (`setdefault`, so an explicit caller-supplied `comfy_usage_source` still wins).
- `comfy_cli/command/run/execution.py` — the local `/prompt` path (`self.extra_data.update()` still runs after the derived default, so an explicit value still overrides — ordering untouched).

Mapping: a non-agentic caller keeps the exact string `comfy-cli`, so existing exact-match consumers (`usage_source = 'comfy-cli'` in PostHog insights and Hex margin queries) keep working and simply narrow to human-driven use — which is what those consumers already assume. Agentic callers become `comfy-cli/<kind>`, the same slash-scoped convention as the `comfy-cloud-mcp/<tool>` values already in production, so `usage_source LIKE 'comfy-cli%'` covers the whole family. The alternative (`comfy-cli/<kind>` for every kind including `user`) is a cleaner taxonomy but breaks every existing exact-match consumer; per the ticket I took the compat-preserving mapping.

`detect_caller()` is env + isatty only, but the answer cannot change within a process, so `usage_source()` is `lru_cache`d — the same once-per-process shape `tracking.py` uses for `_caller_kind`.

## Judgment calls (flagged for review)

1. **The human case keys off `caller.agentic`, not `kind == "user"`.** The ticket's snippet used `kind == "user"`. Those differ in exactly one case: `COMFY_USER_AGENT=user`, which produces the *string* `"user"` while being precisely the self-attributed agentic case — a `kind` test would let an agent bill itself as a human. This is the same reasoning already documented in `tracking._caller_kind_is_custom` (#647), which takes the authoritative signal from `source_env` rather than from set membership. Every other caller maps identically to the ticket's snippet. Pinned by `test_self_attributed_user_label_does_not_pass_as_human`.
2. **The custom label is length-bounded to 64 chars.** `COMFY_USER_AGENT` is arbitrary user-supplied text and this value now rides a request body the server bills against, so it gets the same bound `tracking._sanitize_caller_kind` applies to `caller_kind`. Not asked for by the ticket; happy to drop it if you'd rather the two bounds not be duplicated (the alternative is exporting tracking's sanitizer, which drags `_scrub_value` into the request path).
3. **The three `Comfy-Usage-Source: comfy-cli` request *headers* are deliberately untouched** (`comfy_client.py:243`, `execution.py:175`, `command/generate/client.py:104`). The ticket scopes the change to the two `extra_data` sites, and the header is read at a different layer (CLI-vs-web-UI traffic identification at the gateway) with its own possible exact-match consumers — changing it is a separate decision, not a silent side effect of this one. Say the word and I'll file the follow-up, or widen this PR.
4. **Test determinism.** `usage_source()` reads the live environment, which under pytest is not a stable input: stdout is captured (so the caller detects as `pipe`), and a suite run from an agent harness exports `CLAUDECODE` / `AI_AGENT` / `COMFY_USER_AGENT` into the test process — the existing payload assertions would then pass in a terminal and fail in CI. An autouse fixture in `tests/comfy_cli/conftest.py` pins the two submit modules to the human value; it patches the *importing* modules rather than `caller.detect_caller`, because a global `detect_caller` patch also rewrites what `tracking` and `renderer` see (I tried that first — it broke five `test_tracking.py` cases). Verified load-bearing: with the fixture reverted, `COMFY_USER_AGENT=comfy-mcp pytest tests/comfy_cli/cloud/test_client.py` fails 4 tests, which is also an end-to-end confirmation that the new value really reaches the payload.

## Assumption worth naming

This assumes the server accepts a slash-scoped `usage_source` rather than validating against an allowlist of exact values. The ticket's evidence for that is the `comfy-cloud-mcp/<tool>` values already flowing in production (BE-1605); I could not verify the server side from this repo. If comfy-api does allowlist the value, agentic runs would need that list widened before this lands.

## Out of scope (ticket's own note)

Attribution only lights up for real users once a comfy-mcp release ships PR #163 (the one that exports `COMFY_USER_AGENT=comfy-mcp`); the latest release predates that merge. That release cut is not part of this diff.

## Tests

- `tests/comfy_cli/output/test_caller.py` — new `TestUsageSource`: `user` → `comfy-cli`; `COMFY_USER_AGENT=comfy-mcp` → `comfy-cli/comfy-mcp`; `AI_AGENT=1` → `comfy-cli/agent`; `CLAUDECODE=1`, non-TTY, `AI_AGENT=0` opt-out, the self-attributed-`user` case, the length bound, and that the value is computed once per process.
- `tests/comfy_cli/cloud/test_client.py` — the derived value reaches the cloud payload, and a caller-provided `comfy_usage_source` still wins over it.
- `tests/comfy_cli/command/test_run.py` — the `/prompt` payload carries the derived value, and `self.extra_data` still overrides it.

`pytest` (4350 passed, 36 skipped) and `ruff check` / `ruff format --check` are green. Note: `ruff format --check` also reports `docs/DESIGN-uv-compile.md` as unformatted — that is pre-existing on `main` and an artifact of local ruff 0.16.0 vs the 0.15.15 CI pins; untouched by this diff.

Not applicable: this diff denies no capability. It adds no deny/dead-end path, no "not supported"/"unavailable" strings, and no test asserting a dead-end — it is a purely additive attribution change whose only behavioral effect is a more specific string in a request body.

### DIFF
--- a/comfy_cli/caller.py
+++ b/comfy_cli/caller.py
@@ -18,6 +18,11 @@ own env var (set in every Bash subprocess it spawns). Checking it
 explicitly means analytics can distinguish "Claude Code" from "some
 other agent" without requiring Claude Code users to set AI_AGENT.
 
+This module also derives the ``comfy_usage_source`` attribution value
+(``usage_source`` below) that rides every workflow submission, so
+server-side partner-node billing can tell a human-driven CLI run apart
+from an agent-driven one.
+
 Tested in ``tests/comfy_cli/output/test_caller.py``.
 """
 
@@ -27,6 +32,7 @@ import os
 import sys
 from collections.abc import Mapping
 from dataclasses import dataclass
+from functools import lru_cache
 
 
 @dataclass(frozen=True)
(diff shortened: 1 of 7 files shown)


skishore23 (approved):

mattmillerai:
@claude please rebase

jojodecayz:
recheck

PR #697, opened 2026-08-10
feat(downloads): bound .comfy-downloads/ growth with download_state.prune (BE-6897)

mattmillerai:
## ELI-5

Every time you start a background model download, comfy-cli drops a little bookkeeping file in `<workspace>/.comfy-downloads/`. Nothing ever threw those away — not `download-cancel`, not any poll command — so the folder grew forever, and `comfy model downloads` had to open and parse every file in it just to show you a list that was mostly ancient history. This adds a tidy-up pass that runs automatically, keeps anything still downloading plus the recent stuff, and throws out the rest.

## What changed

`download_state.prune(workspace) -> int` removes a state record only when it is **terminal** (`completed` / `failed` / `cancelled`) and either:

- its `updated_at` is older than 7 days — **except** a `failed`/`cancelled` record whose `dest` still has a `.part` sibling (`file_utils.partial_paths_for`). Those bytes are on disk and the record is the user's only handle for reclaiming them with `download-cancel`, which is the contract `cleanup_partials` already implements. `completed` gets no such carve-out.
- or it falls outside the newest `PRUNE_MAX_TERMINAL_RECORDS` (200) terminal records, ordered by the same `(started_at, id)` sort `list_all` uses. Evicts oldest-first, regardless of age.

A `starting`/`downloading` record is never touched at any age, and never counts toward — or is evicted by — the cap.

Everything is best effort, matching `write_path`'s OSError handling: every stat/unlink is wrapped, a read-only state directory or a file a concurrent prune already removed is a silent no-op, and nothing can raise into or abort a download. Deletion is `Path.unlink(missing_ok=True)` directly — no public `delete()` helper, so this stays self-contained and does not collide with the unrelated `delete()` introduced by the unmerged #688.

Wired into the only two places that touch `.comfy-downloads/` on `main` today, each wrapped so any exception is swallowed:

1. `_submit_background_download` — prunes just before writing the new record (so the new record can never be pruned by its own call).
2. `downloads()` — prunes before `list_all`, i.e. the verb whose cost and output both grow with the directory.

Scoped strictly to the `--background` path and `download_state.py` per the ticket; nothing here builds against `_active_download_for`, `_enforce_claim`, or a `kind` field, none of which exist on `main`.

## Judgment calls

**1. The 200-record cap overrides the `.part` carve-out.** The ticket says the cap applies "regardless of the 7-day rule", and the carve-out is written as an exception *to* that rule. I implemented it as unconditional because that is what makes the directory genuinely *bounded* rather than merely self-expiring — with the carve-out honoured at the cap, an unbounded pile of failed downloads with `.part` siblings could never be pruned, which is the exact growth this ticket exists to stop. Cost: a user sitting on 200+ terminal records loses the `download-cancel` handle for the oldest ones (the `.part` files themselves are untouched and still removable by hand). Covered by `test_the_cap_overrides_the_partial_carve_out`; happy to flip it if you'd rather the carve-out win.

**2. `prune` also deletes the record's `<id>.log` and `<id>.cancel` siblings.** The ticket only names the `.json`, but the worker's stdout/stderr log is written per-download by `_spawn_download_worker` and becomes unreachable the moment the record naming it is gone — no verb can look it up. Deleting records alone would leave the directory growing on `.log` files, defeating the stated goal. Only companions of a record that was actually removed are touched, and the return value still counts records, not files.

**3. A record with an unparseable `updated_at` is kept, not deleted.** `_parse_iso` returning `None` means we cannot show it is stale, so the age rule declines to act; the cap still bounds it. Same posture for a corrupt/unreadable state file, which `read_path` reports as absent and prune therefore never judges. `test_a_garbage_timestamp_is_not_treated_as_ancient` and `test_a_corrupt_file_is_left_alone` pin both.

**4. `list_all` and `prune` now share a `_sort_key` helper** rather than duplicating the sort lambda. They must agree: prune evicts exactly what falls off the end of the list `downloads` shows, so a divergent tiebreak would drop a record the user is still looking at. Behaviour of `list_all` is unchanged.

## Testing

New `TestPrune` class in `tests/comfy_cli/command/test_model_download_background.py`, alongside the existing `TestStatePersistence` / `TestReconcile` (19 tests): stale terminal removed; fresh terminal kept; `starting`/`downloading` kept at 100× the max age; the `.part` guard pinning `failed` and `cancelled` and releasing once the `.part` is gone; `completed` removable with a `.part` present (and the `.part` left on disk); the cap keeping exactly the newest 200 and evicting oldest-first including sub-7-day records; the cap ignoring in-flight records; companion `.log`/`.cancel` removal; corrupt file and garbage-timestamp no-ops; missing state dir; `Path.unlink` monkeypatched to raise `OSError` being a silent no-op **and** a subsequent `_submit_background_download` still returning its normal envelope; and spy-based assertions that `downloads()` and `_submit_background_download` each actually call `prune()` (not just that `download_state` unit-tests pass).

Verification run in the worktree:

- `ruff check .` and `ruff format --diff .` with the CI-pinned `ruff==0.15.15` — clean across all 277 files.
- `pytest tests/comfy_cli/command/test_model_download_background.py tests/comfy_cli/test_file_utils.py` — **150 passed**, covering both changed modules end to end.
- The repo-wide `pytest .` is running but is very slow on this machine (>15 min, unrelated to this diff — it was already slow before the change); CI's `pytest.yml` is the authoritative run for it.

Not applicable — the trigger does not fire. This diff adds no "not supported"/"unavailable"/"STOP" strings, no throw/deny dead-end path, and flips no test to assert a dead-end; it removes stale bookkeeping. The one adjacent capability — reclaiming a dead download's `.part` bytes via `download-cancel` — is preserved by design (the carve-out) and proven by `test_a_partial_pins_a_failed_or_cancelled_record`, with the single deliberate exception documented under judgment call 1.

### DIFF
--- a/comfy_cli/command/models/models.py
+++ b/comfy_cli/command/models/models.py
@@ -739,6 +739,12 @@ def _submit_background_download(
         needs_hf_auth=needs_hf_auth,
     )
 
+    # Retire finished records before adding one, so submitting downloads is what
+    # bounds the state directory rather than something the user has to remember
+    # to do. Purely bookkeeping: never let it fail a submit.
+    with contextlib.suppress(Exception):
+        download_state.prune(workspace)
+
     try:
         # The parent directory has to exist before the worker starts writing, and
         # creating it here means a permission problem surfaces synchronously.
@@ -980,7 +986,12 @@ def download_status(
 def downloads(_ctx: typer.Context):
     """List every background download this workspace knows about, newest first."""
     renderer = get_renderer()
-    rows = [download_state.status_payload(_reconciled(s)[0]) for s in download_state.list_all(get_workspace())]
+    workspace = get_workspace()
+    # Trim before listing: this is the verb whose cost — and whose output — grows
+    # with every record the directory has ever accumulated.
+    with contextlib.suppress(Exception):
+        download_state.prune(workspace)
+    rows = [download_state.status_payload(_reconciled(s)[0]) for s in download_state.list_all(workspace)]
     if not rows:
         print("No background downloads found.")
     else:
--- a/comfy_cli/download_state.py
+++ b/comfy_cli/download_state.py
@@ -61,7 +61,7 @@ import sys
 import time
 import uuid
 from dataclasses import asdict, dataclass
-from datetime import datetime, timezone
+from datetime import datetime, timedelta, timezone
 from pathlib import Path
 from typing import Any
 
@@ -91,6 +91,14 @@ PID_CREATE_TIME_TOLERANCE_S = 1.0
 # Every worker's argv contains this; used to identify a worker when its start
 # time was never recorded.
 WORKER_ARGV_MARKER = "_download-worker"
+
+# How long a finished download's record is kept before :func:`prune` drops it.
+PRUNE_MAX_AGE_S = 7 * 24 * 60 * 60
+
+# Hard ceiling on retained *terminal* records, applied on top of the age rule so
+# a burst of downloads can't leave the directory (and `downloads`' output) large
+# for a week. Active records are never counted or evicted.
+PRUNE_MAX_TERMINAL_RECORDS = 200
 
 _SAFE_ID = re.compile(r"^[a-zA-Z0-9_\-]{1,64}$")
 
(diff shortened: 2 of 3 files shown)


bigcat88 (approved):
Approving. Driven through the real `comfy model downloads` / `model download --background` verbs against planted records, every rule behaves as documented:

```
                                        main                     this PR
stale terminal (30d, 8d)          kept                     pruned
fresh terminal (1d)               kept                     kept
starting/downloading @ 400 days   kept                     kept
companions (.log/.cancel)         orphaned on disk         removed with the record
failed/cancelled + .part          kept                     kept   <- reclaim handle
completed + .part                 kept                     pruned
failed, no .part                  kept                     pruned
garbage / empty updated_at        kept                     kept
210 fresh terminal                210 left                 200 left (c0010..c0209)
205 terminal + 50 in-flight       255 left                 200 terminal + 50 in-flight
pruned id -> download-status      rc=0                     rc=1 download_not_found
read-only state dir               ok                       ok, no traceback
submit path                       no prune                 prunes before writing
```

Two things I specifically wanted to see and did:

**Judgment call 1 costs less than the PR text implies.** With 205 `failed` records that *all* have `.part` siblings, the cap does win — 200 kept, 5 handles lost, as you documented. But I checked the disk afterwards and all **205 `.part` files are still there**. So the cost is only the `download-cancel` shortcut, never the bytes, and a user can still reclaim them by hand. That's worth stating in the PR text, because "loses the handle" reads worse than it is. I'd leave the cap unconditional.

**In-flight records really are outside the cap in both directions.** 50 `downloading` records aged 400 days plus 205 terminal ones leaves exactly 50 + 200 — the active ones neither evict a terminal record nor get evicted themselves.

One note for anyone reproducing this: a partial is named `<name>.<8-char mkstemp token>.part`, and `partial_paths_for` rejects anything else, so a hand-written `<dest>.part` is invisible to the carve-out. My first run showed the carve-out "failing" for exactly that reason. Your fixture presumably builds the real name; worth a comment there if it doesn't.

`_has_partials` failing *open* (returning True on OSError/ValueError, so the record survives) is the right posture — every uncertainty in this function resolves toward keeping data.

**On #686:** it implements the same feature for BE-6653 and conflicts with this in `download_state.py`, `models.py` and the shared test file. Per the maintainer's call this one lands and #686 gets closed — chiefly for the `.part` carve-out and for keeping records whose timestamps don't parse, both of which #686 prunes away.

Full `pytest` on your branch against current `main`: **4360 passed**, 31 skipped. The one failure is `test_non_fast_deps_uses_global_python`, which fails identically on `main`.

PR #698, opened 2026-08-10
fix(cql): answer same-type `nodes path` queries instead of declining them

mattmillerai:
> **Base is now `main`.** This PR was opened stacked on #695 (`matt/be-6857-nodes-path-type-constraint`); #695 has merged and GitHub retargeted this to the default branch. The earlier STACKED banner no longer applies.

## ELI-5

Ask the CLI "how do I get from a MODEL to a MODEL?" and it used to shrug: zero routes, plus a note saying it hadn't actually looked. But there is an obvious answer — a LoRA loader takes a MODEL and hands you back a MODEL. It does that all day. The tool just refused to say so.

Two things were in the way. A guard at the top bailed out whenever the start and end types matched, before any searching happened. And the search itself had a rule that throws away any step handing back the type it was given — a sensible rule for a step in the middle of a route (that step accomplishes nothing), but for a same-type question that is the *only* step that could answer it.

Now the guard is gone and the rule makes an exception for the final step. `comfy nodes path MODEL MODEL` lists `LoraLoaderModelOnly`. Every other question you can ask gets exactly the same answer it did before.

## What changed

- `comfy_cli/cql/engine.py::Graph.search_paths` — removed the early `from_type == to_type` abstention; changed the no-op rule from `if out_t == cur_type` to `if out_t == cur_type and out_t != to_type`. Docstring updated: `degenerate_bounds` is now the sole `not_searched_reason`.
- `comfy_cli/command/nodes.py::path_cmd` — docstring no longer claims `same_type` is the reason reachable from the CLI. No `not_searched` reason is reachable from the CLI at all now: the only shape `search_paths` declines is a bound below 1, which `path_cmd` already rejects up front with `path_bounds_invalid`.
- Tests updated per the plan, plus one new guard test (below).

## Why the exemption cannot leak into other queries

A step whose output matches `to_type` is appended as a completed path and `continue`d — it is never pushed onto the queue. So `cur_type == to_type` can only hold for the initial frontier item, which means only when `from_type == to_type`. Every other search is untouched.

Verified rather than argued: a differential harness ran `search_paths` on this branch and on the base commit over **1568 combinations** — every FROM/TO pair of every type in both catalog fixtures (`nodes_path_object_info.json`, `sd15_object_info.json`) × exact/loose × `max_depth` ∈ {1,2,3,6} × `max_paths` ∈ {1,10}, excluding the same-type pairs that are the intended change. The serialized results are byte-identical.

## Empirical check through the real CLI

```
$ comfy --json nodes path MODEL MODEL --input tests/comfy_cli/fixtures/nodes_path_object_info.json
{"count": 1, "not_searched": false, "not_searched_reason": null,
 "paths": [{"from": "MODEL", "to": "MODEL",
            "steps": [{"node": "LoraLoaderModelOnly", "from_type": "MODEL", "to_type": "MODEL"}],
            "support": []}], ...}
```

Against the `sd15` fixture, `LATENT -> LATENT` returns `KSampler` with `exact: true` — a same-type answer that carries the full exhaustiveness claim.

## Judgment calls and one unmet acceptance criterion

**1. `exact: true` on `MODEL MODEL` is NOT met, deliberately.** The route is listed and `not_searched` is false, but the envelope reports `exact: false` because `collapsed` is true. That collapse is genuine and unrelated to this change: reaching MODEL *ends* a path, so the walk keeps expanding the branches that do not end (`KSampler -> LATENT`), and there both `VAEDecode` and `VAEDecodeTiled` land on the same `(IMAGE, {IMAGE, LATENT})` state. #695 deliberately made `collapsed` err toward true, and forcing `exact: true` here would mean suppressing that flag — i.e. forging exactly the proof #695 was written to stop forging. So the test asserts `collapsed is True` / `exact is False` with the reason spelled out. The criterion holds wherever no unrelated collapse occurs, e.g. `LATENT -> LATENT` on the sd15 fixture above. Flagging it rather than hiding it: if reviewers want `exact: true` for same-type queries, that is a change to `collapsed`'s reachability semantics and belongs in its own ticket.

**2. The plan's fixture assumption was wrong for one test.** It said to convert `test_find_paths_same_type_returns_empty` to assert `graph.find_paths("MODEL","MODEL")` returns the `LoraLoaderModelOnly` route — but that test uses the small inline `graph` fixture, which has no `LoraLoaderModelOnly` and no MODEL-consuming-MODEL-producing node at all. Only `graph_path` carries one. So that test now asserts the honest outcome for its own catalog (still empty, but `not_searched is False` — a fact about the catalog, not an abstention), and the positive route assertion lives in `test_same_type_query_finds_the_route`, which uses `graph_path`, exactly as the plan's item 3.4 specified.

**3. Added one test beyond the plan.** Removing an existing guard risks over-broadening, so `test_no_op_hops_are_still_dropped` pins the other half: for a FROM != TO query no step may hand back the type it consumed, and within a same-type query only the terminal hop may. Mutation-checked both ways — dropping the no-op rule entirely fails it, and restoring the old unconditional rule fails the capability tests.

## Verification

- `ruff check .` and `ruff format --check .` — clean.
- Full suite (`pytest`) — **4374 passed, 36 skipped** in 21m.
- `pytest tests/comfy_cli/cql/test_engine.py tests/comfy_cli/command/test_nodes_introspect.py` — 194 passed (re-run after the final test-only edits, which landed after the full-suite run started; all product code was final before it).
- Differential equivalence harness — 1568/1568 identical (above).

## Note on capability direction

The falsification discipline applies inverted here: this diff *removes* a capability denial rather than adding one. The route it now surfaces was confirmed live through the CLI before the denial was deleted, not merely asserted by a test.

## Post-review fix (2026-08-11)

@bigcat88 caught that the `origin/main` merge into this branch (86ed079) resolved `nodes._get_graph` to the
outgoing side, silently reverting #687 (BE-6660): the `report_usage_error(get_renderer())` wrapper around
`resolve_host_port` was deleted. That made every `comfy nodes` verb exit 2 with **zero bytes on stdout** for a
rejected `--host`/`--port` instead of emitting the terminating `host_port_invalid` envelope — the exact
stream-just-stops failure BE-6660 was filed for, and the red `build` check.

Reproduced locally on the merged tree before fixing:
`test_host_port.py::test_nodes_bad_port_terminates_with_envelope` failed deterministically
(`-p no:randomly`: 1 failed, 79 passed), and passes with the wrapper restored. Restored verbatim from `main`;
the only remaining `nodes.py` delta versus `main` is this PR's `path_cmd` docstring. Current `origin/main`
merged in at the same time.

### DIFF
--- a/comfy_cli/command/nodes.py
+++ b/comfy_cli/command/nodes.py
@@ -745,7 +745,43 @@ def path_cmd(
         typer.Option("--where", show_default=False, help="'cloud' to query Comfy Cloud's catalog; default is local."),
     ] = None,
 ):
+    """Envelope contract (``data``):
+
+    - ``mode`` — the *requested* matching mode, ``"exact"`` or ``"loose"``. It
+      echoes ``--exact/--loose`` and says nothing about completeness.
+    - ``exact`` — the exhaustiveness claim, and deliberately NOT the flag echoed
+      back: true only when the listed paths are the complete, type-constrained
+      answer. Exact mode that stopped early (``truncated``), was still expanding
+      at the bound (``depth_limited``), or dropped an alternate route into an
+      already-explored state (``collapsed``) withholds the claim, as does loose
+      mode always. ``exact: true`` with ``count: 0`` is therefore a proof that
+      no route exists; ``exact: false`` means "these paths, maybe not all".
+    - ``truncated`` / ``truncated_by`` / ``depth_limited`` / ``collapsed`` — the
+      individual reasons the claim was withheld, so a caller can widen the right
+      bound instead of guessing.
+
+    One documented exception to the ``exact: true, count: 0`` proof: a query
+    whose FROM and TO are the *same* type is answered empty by construction
+    (``Graph.search_paths`` declines to walk it, long-standing behaviour), so it
+    reports an exhaustive empty set even where a real self-returning route such
+    as ``MODEL -> LoraLoader -> MODEL`` exists. Tracked separately; do not read
+    a same-type empty result as a proof of unreachability.
+    """
     renderer = get_renderer()
+
+    # A bound below 1 admits no path at all, so the search would return an empty
+    # result with every flag false — i.e. `exact: true, count: 0`, a proof that
+    # no route exists. That proof would come from the typo, not from a walk, so
+    # refuse the bound instead of emitting it.
+    if max_depth < 1 or max_paths < 1:
+        renderer.error(
+            code="path_bounds_invalid",
+            message="--max-depth and --max-paths must be at least 1.",
+            hint="retry with `--max-depth 6 --max-paths 10`",
+            details={"max_depth": max_depth, "max_paths": max_paths},
+        )
+        raise typer.Exit(code=1)
+
     _stale: dict = {}
     graph = _get_graph(
         input_path,
--- a/comfy_cli/command/nodes.py
+++ b/comfy_cli/command/nodes.py
@@ -734,7 +734,7 @@ def path_cmd(
         bool,
         typer.Option(
             "--exact/--loose",
-            help="Exact: every step's required link inputs must be satisfiable from the path so far. Loose: any routed sequence.",
+            help="Exact: every step's other required link inputs must be satisfiable (reported per path as 'support'). Loose: any routed sequence.",
         ),
     ] = True,
     input_path: Annotated[str | None, typer.Option("--input", show_default=False)] = None,
@@ -755,13 +755,26 @@ def path_cmd(
         on_stale=lambda key, err: _stale.update(stale=True, source=key, reason=err),
     )
 
-    finder = graph.exact_paths if exact else graph.find_paths
-    paths = finder(from_type, to_type, max_depth=max_depth, max_paths=max_paths)
+    result = graph.search_paths(from_type, to_type, exact=exact, max_depth=max_depth, max_paths=max_paths)
+    paths = result["paths"]
+    truncated = bool(result["truncated"])
+    depth_limited = bool(result["depth_limited"])
+    collapsed = bool(result["collapsed"])
 
     payload = {
         "from": from_type,
         "to": to_type,
-        "exact": exact,
+        "mode": "exact" if exact else "loose",
+        # Not the flag echoed back: the honest claim that these paths are the
+        # complete, type-constrained answer. Any early stop (max_paths, the
+        # internal state budget), a frontier still expanding at max_depth, or an
+        # intermediate state reached by a second route that was not re-explored
+        # means paths may be missing, so the claim is withheld.
+        "exact": bool(exact and not truncated and not depth_limited and not collapsed),
+        "truncated": truncated,
+        "truncated_by": result["truncated_by"],
+        "depth_limited": depth_limited,
+        "collapsed": collapsed,
         "max_depth": max_depth,
         "max_paths": max_paths,
         "count": len(paths),
@@ -777,6 +790,7 @@ def path_cmd(
                     }
                     for s in (p.get("steps") or [])
                 ],
+                "support": list(p.get("support") or []),
             }
             for p in paths
         ],
@@ -805,7 +819,19 @@ def path_cmd(
                     f"[cyan]{sanitize_markup(p.get('from'))}[/cyan]  {chain}  "
                     f"[cyan]{sanitize_markup(p.get('to'))}[/cyan]"
                 )
+                needs = ", ".join(
+                    f"{sanitize_markup(s.get('type'))} from {sanitize_markup(s.get('node'))}"
+                    for s in (p.get("support") or [])
+                )
+                if needs:
+                    rprint(f"  [dim]also needs: {needs}[/dim]")
             rprint(f"[dim]{len(paths)} path(s)[/dim]")
+        if truncated:
+            rprint(f"[dim]Partial result — stopped at {payload['truncated_by']}; more paths may exist.[/dim]")
+        elif depth_limited:
+            rprint(f"[dim]Searched to depth {max_depth}; longer paths were not explored.[/dim]")
+        elif collapsed:
+            rprint("[dim]Equivalent alternate routes were collapsed; this is a sample, not every path.[/dim]")
     renderer.emit(payload, command="nodes path")
 
 
--- a/comfy_cli/command/nodes.py
+++ b/comfy_cli/command/nodes.py
@@ -760,12 +760,14 @@ def path_cmd(
       ``not_searched`` — the individual reasons the claim was withheld, so a
       caller can widen the right bound instead of guessing.
     - ``not_searched`` / ``not_searched_reason`` — the walk declined the query
-      and never ran, so the empty result is an abstention, not an answer. Today
-      the only reason reachable from the CLI is ``"same_type"``: a query whose
-      FROM and TO are the same type is answered empty by construction, even
-      though real self-returning routes such as ``MODEL -> LoraLoader -> MODEL``
-      exist. Such a result reports ``exact: false`` and must not be read as a
-      proof of unreachability.
+      and never ran, so the empty result is an abstention, not an answer. No
+      reason is reachable from this command: the only shape ``search_paths``
+      declines is a bound below 1, which is rejected up front with
+      ``path_bounds_invalid`` before the graph is even loaded. Same-type queries
+      (``MODEL MODEL``) are searched like any other and return the real
+      self-returning routes. The fields stay in the envelope so a caller can
+      never mistake an abstention the engine adds later for a proof of
+      unreachability — an abstention always reports ``exact: false``.
     """
     renderer = get_renderer()
 
--- a/comfy_cli/command/nodes.py
+++ b/comfy_cli/command/nodes.py
@@ -74,6 +74,13 @@ def _get_graph(
     and fired when a stale-cache fallback occurs (see loader for signature).
     """
     mode = _resolved_where(where)
+    # Every `comfy nodes` verb routes through here, so this is the one place
+    # the target is decided — stamp it on the renderer so the envelopes emitted
+    # downstream (the `cql_no_graph` error below included) carry `where`
+    # instead of `null`. `--input` reads a dump instead of talking to the
+    # backend, but `mode` still says which catalog the dump is interpreted as,
+    # which is what `where` reports (see the field's schema description).
+    get_renderer().where = mode
     # Resolve the local server the same way `comfy run` / `comfy jobs` do —
     # flag > COMFY_LOCAL_URL > config.background > 127.0.0.1:8188. `resolve_target`
     # deliberately skips the `config.background` step (other callers must not
--- a/comfy_cli/command/nodes.py
+++ b/comfy_cli/command/nodes.py
@@ -756,16 +756,16 @@ def path_cmd(
       already-explored state (``collapsed``) withholds the claim, as does loose
       mode always. ``exact: true`` with ``count: 0`` is therefore a proof that
       no route exists; ``exact: false`` means "these paths, maybe not all".
-    - ``truncated`` / ``truncated_by`` / ``depth_limited`` / ``collapsed`` — the
-      individual reasons the claim was withheld, so a caller can widen the right
-      bound instead of guessing.
-
-    One documented exception to the ``exact: true, count: 0`` proof: a query
-    whose FROM and TO are the *same* type is answered empty by construction
-    (``Graph.search_paths`` declines to walk it, long-standing behaviour), so it
-    reports an exhaustive empty set even where a real self-returning route such
-    as ``MODEL -> LoraLoader -> MODEL`` exists. Tracked separately; do not read
-    a same-type empty result as a proof of unreachability.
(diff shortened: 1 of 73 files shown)


mattmillerai:
Independent re-verification of this branch (fresh worktree at `4b2e20f`, deps via `uv sync --frozen --extra dev`), re-running the claims in the description rather than taking them on trust:

- **Acceptance tests** — `pytest tests/comfy_cli/cql/test_engine.py tests/comfy_cli/command/test_nodes_introspect.py` → **194 passed** in 21s.
- **Non-same-type equivalence, re-derived independently** — a separate harness (not the one in the description) enumerated every FROM/TO pair over both catalog fixtures × exact/loose × `max_depth` ∈ {1,2,3,6} × `max_paths` ∈ {1,10}, excluding same-type pairs, and compared serialized `search_paths` output against the base commit `d66bc22`: **1152/1152 byte-identical**.
- **The scoping argument, checked against the code rather than the comment** — every state whose `out_t == to_type` hits `paths.append(...)` + `continue` and is never appended to `next_queue`; the queue is seeded only with `from_type`. So `cur_type == to_type` can hold only for the seed item, i.e. only when `from_type == to_type`. The new clause therefore differs from the old one strictly when `out_t == cur_type == to_type`, which is exactly the same-type case.
- **CLI acceptance criterion** — `comfy --json nodes path MODEL MODEL --input tests/comfy_cli/fixtures/nodes_path_object_info.json` returns `count: 1`, the one-step `LoraLoaderModelOnly` route, `not_searched: false`. `exact` is `false` (`collapsed: true`) — the deliberate deviation already documented under "Judgment calls" above.
- **The one remaining empty assertion, falsified rather than assumed** — `test_find_paths_same_type_is_searched_not_declined` asserts `find_paths("MODEL","MODEL") == []` on the small inline fixture. Checked directly: only `CheckpointLoaderSimple` emits MODEL in that catalog and it consumes none, so no MODEL-in/MODEL-out node exists. The empty result is a property of the fixture, not a restatement of the test's own premise.

One pre-existing lint note, not from this diff: `ruff check` flags `UP038` at `comfy_cli/cql/engine.py:1463` (`isinstance(spec, (list, tuple))`), well outside the changed region. It reproduces identically on the base branch `d66bc22`, so it is ruff-version drift against the pinned CI version, not a regression here — repo CI is green.

bigcat88:
Needs a rebase onto `main` before I can review it.

Its base #695 landed as **squash** merge `5ebc831`, so this branch's own copies of those commits no longer match what's on `main` — merging `origin/main` in now conflicts:

```
CONFLICT (content): Merge conflict in tests/comfy_cli/command/test_nodes_introspect.py
CONFLICT (content): Merge conflict in tests/comfy_cli/cql/test_engine.py
```

Both are test files where #695's cases and this PR's same-type cases were added in the same regions. The single commit that is genuinely unique here is `4b2e20f fix(cql): answer same-type path queries instead of declining them`, so a `git rebase --onto origin/main pr695` (or cherry-picking that one commit onto a fresh branch off `main`) should reduce this to a small diff.

Context from reviewing #695: I confirmed the abstention this PR removes is live on `main` right now —

```
comfy --json nodes path IMAGE IMAGE --input <real object_info>
  -> count=0  exact=false  not_searched=true  not_searched_reason="same_type"
```

so the behaviour you're fixing is real and reachable. Worth keeping in mind for the rebased version: on a real 792-class catalog, hub types like IMAGE collapse or truncate almost immediately, so a same-type answer will realistically come back `exact: false` — make sure the new tests don't assume `exact: true` on anything but a small fixture.

Ping me once it's rebased and I'll pick it up.

mattmillerai:
@bigcat88 merged \`origin/main\` in (\`86ed079\`), resolving the conflicts in \`comfy_cli/command/nodes.py\`, \`comfy_cli/cql/engine.py\`, and both test files — kept both #695's source-type-constraint cases and this PR's same-type cases. The PR diff is now scoped back down to the 4 files / single logical change (\`4b2e20f\`); pushed and CI is re-running. Ready for your review whenever you get a chance.

comfy_cli/command/nodes.py
```diff
@@ -82,14 +82,9 @@ def _get_graph(
...
-
-        # A rejected `--host`/`--port` raises `typer.BadParameter`, which click
-        # turns into a stderr usage panel + exit 2 with nothing on stdout.
-        # Emit the terminating envelope first so JSON/NDJSON consumers get a
-        # parseable final line; the exception still escapes, so exit stays 2.
-        with report_usage_error(get_renderer()):
-            host, port = resolve_host_port(host, port)
+        from comfy_cli.host_port import resolve_host_port
+
+        host, port = resolve_host_port(host, port)
```
bigcat88: This drops the `report_usage_error` wrapper that #687 (BE-6660) added here yesterday, and the PR body doesn't mention it — it looks like a merge resolution that took the outgoing side.

Effect: every `comfy nodes` verb goes back to exit 2 with **zero bytes on stdout** for a bad `--host`/`--port`, instead of the terminating `host_port_invalid` envelope. Measured on this branch with current main merged in: `comfy --json nodes ls --port 0` gives 0 B on stdout vs 396 B on main, same for `nodes show` / `nodes path`, the `--host 1.2.3.4:0` form, and NDJSON mode. `test_nodes_bad_port_terminates_with_envelope` fails deterministically — that is your red build check.

Please restore:

```python
        from comfy_cli.host_port import report_usage_error, resolve_host_port

        # A rejected `--host`/`--port` raises `typer.BadParameter`, which click
        # turns into a stderr usage panel + exit 2 with nothing on stdout.
        # Emit the terminating envelope first so JSON/NDJSON consumers get a
        # parseable final line; the exception still escapes, so exit stays 2.
        with report_usage_error(get_renderer()):
            host, port = resolve_host_port(host, port)
```

bigcat88 (changes requested):
The same-type fix itself is correct and I verified it independently — but the branch also **silently reverts #687**, which is what your red `build` check is. One deletion to undo and this is good to go.

## The blocker: `report_usage_error` is gone from `nodes._get_graph`

`git diff origin/main...` shows this branch removing the wrapper that #687 (BE-6660, merged yesterday) added:

```python
-        with report_usage_error(get_renderer()):
-            host, port = resolve_host_port(host, port)
+        host, port = resolve_host_port(host, port)
```

Nothing in the PR body mentions it, and it has nothing to do with same-type queries — it reads like a merge resolution that dropped the incoming side. `git log -S` confirms the wrapper's only author is #687.

Measured through the real CLI, after merging current `origin/main` into this branch:

| `comfy --json … --port 0` | `main` | this branch |
|---|---|---|
| `nodes ls` | rc=2, **396 B** `host_port_invalid` envelope | rc=2, **0 B on stdout** |
| `nodes show KSampler` | rc=2, 396 B envelope | rc=2, **0 B** |
| `nodes path MODEL IMAGE` | rc=2, 396 B envelope | rc=2, **0 B** |

Also 0 bytes for the `--host 1.2.3.4:0` form and in NDJSON mode. That is exactly the "machine consumer just sees the stream stop" failure BE-6660 was filed for, restored for every `comfy nodes` verb.

`tests/comfy_cli/test_host_port.py::test_nodes_bad_port_terminates_with_envelope` fails deterministically on the merged branch (`-p no:randomly`: 1 failed, 79 passed) — it is not the flake the single red check might suggest.

**Fix:** restore the two lines (and the `report_usage_error` import) in `_get_graph`. Nothing else in the diff needs to change.

## What I verified about the actual change (all good)

**The no-op exemption does not leak — independently reproduced, not taken on trust.** I ran `search_paths` over every FROM/TO type pair × `exact`/`loose` × `max_depth` ∈ {1,2,3} on both trees:

| catalog | cross-type queries | changed |
|---|---|---|
| **live server, 792 classes, 109 link types** | **70,632** | **0** |
| `nodes_path_object_info.json` | 252 | 0 |
| `sd15_object_info.json` | 180 | 0 |
| `subgraph_object_info.json` | 252 | 0 |

All 654 same-type queries on the real catalog changed, exactly as intended — 225 of them from an abstention to a list of real routes. Your reasoning for why `cur_type == to_type` can only hold on the initial frontier item holds up empirically at this scale.

**Live CLI, against a real ComfyUI on :8189:**

```
main:   nodes path MODEL MODEL -> count=0  not_searched=true  reason=same_type
branch: nodes path MODEL MODEL -> count=10 not_searched=false  (APG, CFGNorm, CFGOverride, …)
        nodes path LATENT LATENT -> count=3 (AddNoise, HunyuanRefinerLatent, …)
```

Real MODEL-patching nodes, which is the answer a user asking that question actually wants.

I also agree with the call in your "unmet acceptance criterion" note: reporting `exact: false` because `collapsed` is genuinely true is the right outcome. Forcing `exact: true` there would forge exactly the proof #695 exists to prevent. Leave it.

## Two minor things while you're in here

- The branch is 5 commits behind `main` (#694, #690, #528, #641, #699 landed today). Worth a `git merge origin/main` with the fix.
- The body still carries the `STACKED — merging lands on matt/be-6857…` banner; #695 has merged and GitHub retargeted this to `main`.

mattmillerai:
Confirmed and fixed in 3b0bb2d — you were right on every detail, including the cause.

`git show 86ed079` shows the merge resolving `_get_graph` to the outgoing side: parent `4b2e20f` (this branch, pre-merge) has the bare `resolve_host_port` call, parent `87f0414` (`main`) has your wrapper, and the merge kept the branch's. Nothing about it was intentional — the same-type change never touches `_get_graph`.

Reproduced before fixing, on the tree with current `origin/main` merged in:

```
$ pytest tests/comfy_cli/test_host_port.py -p no:randomly -q
FAILED test_host_port.py::test_nodes_bad_port_terminates_with_envelope
  AssertionError: stdout was empty; stderr: Usage: root nodes ls [OPTIONS] ...
1 failed, 79 passed
```

Restored the two lines verbatim from `main` (wrapper, comment, and the `report_usage_error` import) — same suite is now 80 passed, and the full suite is 4650 passed / 36 skipped.

The only remaining `nodes.py` delta versus `main` is this PR's `path_cmd` docstring, so nothing else from #687 is at risk:

```
$ git diff origin/main..HEAD -- comfy_cli/command/nodes.py --stat
 comfy_cli/command/nodes.py | 14 +++++++-------
```

Both minor points handled too: current `origin/main` is merged in (5073111), and the PR body's stale `STACKED` banner is replaced with a note that #695 merged and GitHub retargeted this to `main`.

Thanks for the differential run over 70,632 cross-type queries — that is a much stronger check on the exemption than my 1,568-combination fixture harness, and against a real 792-class catalog rather than fixtures.

comfy_cli/command/nodes.py
mattmillerai: Fixed in 3b0bb2d — restored verbatim from `main`, including the `report_usage_error` import and the comment.

Your read of the cause is exactly right: `86ed079` resolved this hunk to the outgoing side. Parent `4b2e20f` (branch, pre-merge) has the bare call, parent `87f0414` (`main`) has the wrapper, and the merge kept the branch's.

Reproduced first on the merged tree — `test_nodes_bad_port_terminates_with_envelope` failed with `stdout was empty` — and it passes now (80 passed in that file, 4650 passed / 36 skipped for the full suite). The only remaining `nodes.py` delta versus `main` is this PR's `path_cmd` docstring.

bigcat88 (approved):
Fixed — approving and merging. `3b0bb2d` restores the wrapper, and the restored block is **byte-identical** to `main`'s (diffed the whole `if input_path is None and mode == "local":` region, no delta).

## The regression is gone

Same probes that caught it, on the branch with current `main` merged in:

| `comfy --json … --port 0` | before | now |
|---|---|---|
| `nodes ls` | 0 B on stdout | rc=2, **396 B** `host_port_invalid` |
| `nodes show KSampler` | 0 B | rc=2, 396 B |
| `nodes path MODEL IMAGE` | 0 B | rc=2, 396 B |
| `--host 1.2.3.4:0` | 0 B | rc=2, 463 B |
| NDJSON mode, `--port 70000` | 0 B | rc=2, 400 B |

`tests/comfy_cli/test_host_port.py`: **80 passed** (was 1 failed / 79 passed).

## Re-ran the leak differential, because `main` moved under you

#574 landed in `cql/engine.py` overnight, so yesterday's numbers no longer covered the tree you'd actually merge into. Re-ran the whole thing against today's `main`, over the live 792-class catalog — every FROM/TO pair × exact/loose × `max_depth` ∈ {1,2,3}:

```
cross-type queries : 69,336   CHANGED: 0
same-type queries  :    648   changed: 648   (238 now return real routes)
```

Still zero leakage. Your argument for why `cur_type == to_type` can only hold on the initial frontier item holds against a catalog with 108 link types and #574's changes in it.

Through the real CLI, offline fixture:

```
main   : nodes path MODEL MODEL -> count=0  not_searched=true  reason=same_type
branch : nodes path MODEL MODEL -> count=1  LoraLoaderModelOnly
```

## Everything else

- Full suite merged with current `main`: **4783 passed, 31 skipped**. The one failure is `test_non_fast_deps_uses_global_python`, which fails identically on plain `main` in this environment.

Nice catch writing the commit message against the merge commit that caused it (`86ed079 resolved _get_graph to the outgoing side`) — that's the detail that makes this findable if it ever happens again. Worth knowing the same hazard applies to every long-lived branch that merges `main` repeatedly: a conflict in a hunk you didn't author is exactly where a silent revert hides.

PR #700, opened 2026-08-10
fix(output): populate `where` on error envelopes for every routed command (BE-6275)

mattmillerai:
## ELI-5

When a `--json` command fails, it hands back an "envelope" that says what went wrong. That envelope has a `where` field meant to say *which backend* the command was talking to — your local ComfyUI, or Comfy Cloud. Until now it always said `null` on failures, so an agent reading a "server not running" error couldn't tell which server. This fills it in: once a command has decided local-vs-cloud, every error it reports from that point on names the target. Errors that happen *before* that decision (like "`--where nowhere` isn't a thing") still say `null`, because there genuinely isn't an answer yet.

## What changed

Phase 1 (#665 / BE-6274) fixed `Renderer.error()`'s `where` override and stamped the `run` path. Every other routed command still fell through to `where: null` because nothing assigned `renderer.where`. Rather than edit ~130 individual `renderer.error(...)` call sites, this assigns `renderer.where` **once per command, at the point its effective target is decided** — the pattern phase 1 established:

| Surface | Assignment point |
|---|---|
| `jobs ls/status/wait/cancel/watch` | new `_stamp_where()` helper, at each verb's local-vs-cloud decision |
| `workflow list/get/save/delete` | `_resolve_where_target()` |
| `workflow slots/set-slot/vary/notes` | `workflow._get_graph()`, after the object_info route resolves |
| `workflow decompose` | `workflow_fragments._load_object_info()` |
| `upload` / `download` | `cmdline.py` after `effective_where`, and again in `transfer.execute_*` for direct (comfy-mcp) callers |
| `system-stats` / `free` | `system.py` after `resolve_target` (these are what `cmdline.py:1381/1408` delegate to) |
| `nodes *` | `nodes._get_graph()` |
| `models list-folders/list-folder/search/show` | new `_resolve_and_stamp()` helper |
| `assets push` | after `resolve_target` |
| `validate` | after `resolve_default_or_exit` |
| `logs` | after the local-only routing check passes |

Rules held throughout: vocabulary is strictly `"local"`/`"cloud"` (`target.kind` and `WhereTarget.value` are the only sources — no invented labels); the stamp lands **only after a successful decision**, so `where_invalid` and other pre-routing errors keep `where: null`; every existing explicit `emit(..., where=...)` argument is left in place and still wins (`where or self.where`). `comfy_cli/schemas/envelope.json` is unchanged — it already declared `where` as a nullable `local`/`cloud` enum.

## Tests

New `tests/comfy_cli/output/test_error_envelope_where.py` (36 cases) drives the real CLI with a forced failure *after* target resolution and asserts `where` on the `ok:false` envelope — one or more per touched command, both targets where both exist. Four negative cases pin the other half of the contract (pre-decision errors stay `where: null`). Full suite: **4569 passed, 36 skipped**; `ruff check` + `ruff format --diff` clean (against the CI-pinned ruff 0.15.15).

## Judgment calls

- **Scope went slightly wider than the ticket's list, deliberately.** The ticket enumerates workflow/jobs/transfer/system/nodes/models. I also stamped `validate`, `assets push`, `logs`, `workflow slots/set-slot/vary/notes`, and `workflow decompose` — every one of them resolves a local/cloud routing decision and then emits error envelopes, so leaving them out would have shipped the exact inconsistency the ticket exists to remove (`nodes ls` reporting `local` on an object_info failure while `workflow slots` reported `null` on the identical failure). Each is the same one-line assignment at the same kind of decision point.
- **Some success envelopes now carry `where` too, where they previously said `null`.** `nodes *`, `models *`, `validate`, and `assets push` never passed `where=` to `emit()` explicitly (they carry the target as a *payload* field instead), so the stamp fills the envelope-level fallback on success as well as failure. This is additive and matches the schema's own description of the field; no test asserted the old `null`. Commands that pass `where=` explicitly (`workflow`, `jobs`, `run`, `logs`) are byte-identical on success.
- **Out of scope, unchanged:** `env`, `which`, `feedback`, `nodes refresh` (its `--where` is documented as accepted-and-ignored), `setup`, and `templates` — none of them route, so their envelopes correctly stay `where: null`.
- **Known bound (not introduced here):** `Renderer` is a process-wide singleton, and these stamps are not unset at command exit the way `run`'s `finally` unsets its own. On the CLI this cannot leak — `cmdline.entry()` installs a *fresh* `Renderer` per invocation. It is only observable in a long-lived process that calls two different comfy-cli command functions in-process without reinstalling a renderer, where the second one fails *before* it routes; each entry point re-resolves and re-stamps before any post-routing error, so the window is limited to pre-routing failures. Left as-is rather than adding a `try/finally` to a dozen command bodies; happy to follow up if that in-process pattern matters to comfy-mcp.

## Verification of the acceptance criteria

- Failing `comfy workflow list --where cloud --json` → `where: "cloud"`; `--where local` → `"local"` (`TestWorkflow`).
- Failing `comfy jobs status --json` → `"cloud"`/`"local"` per routing (`TestJobs`).
- Failing `comfy upload --where cloud --json` → `"cloud"`, including the `--host` + cloud rejection that never reaches `execute_upload` (`TestTransfer`).
- `where_invalid` still emits `where: null` (`TestPreDecisionErrorsStayNull`).
- Full suite green; `comfy_cli/schemas/envelope.json` untouched.

### DIFF
--- a/comfy_cli/cmdline.py
+++ b/comfy_cli/cmdline.py
@@ -1188,6 +1188,11 @@ def validate(
     # made by string-comparing user input.
     decision = where_module.resolve_default_or_exit(flag=where)
     mode = decision.target.value
+    # Routing resolved — stamp it so the envelopes below (the object_info load
+    # failure and the UI-conversion errors) name the target this validate ran
+    # against. The file-read errors above stay `where: null`: they precede the
+    # decision, as does the `where_invalid` `resolve_default_or_exit` raises.
+    renderer.where = mode
 
     # Resolve the local object_info server the same way `comfy run` does —
     # flag > COMFY_LOCAL_URL > config.background > 127.0.0.1:8188. Without the
@@ -1403,6 +1408,11 @@ def upload(
         raise typer.Exit(code=1)
 
     effective_where = "cloud" if decision.target is where_module.WhereTarget.CLOUD else "local"
+    # Routing is decided, so every error envelope from here down can name the
+    # target — including the `host_flag_cloud` rejection immediately below,
+    # which never reaches `execute_upload`. The `where_invalid` above stays
+    # `where: null`: it failed before there was a decision to report.
+    renderer.where = effective_where
     # --host/--port address a local ComfyUI; the cloud target's address comes
     # from the signed-in account's base URL and ignores them entirely
     # (``Target.host``/``Target.port`` are documented local-only). Rejecting
@@ -1464,6 +1474,10 @@ def download(
         raise typer.Exit(code=1)
 
     effective_where = "cloud" if decision.target is where_module.WhereTarget.CLOUD else "local"
+    # As in `upload`: stamped once routing resolves, so the preflight failure
+    # below and `execute_download`'s stdin-parsing errors (which run before it
+    # resolves its own target) all name the backend this invocation routed to.
+    renderer.where = effective_where
     if effective_where == "cloud":
         where_module.cloud_preflight_or_exit()
 
--- a/comfy_cli/command/jobs.py
+++ b/comfy_cli/command/jobs.py
@@ -682,8 +682,9 @@ def ls_cmd(
 
     # Resolve the routing target once: per-command --where flag > COMFY_WHERE
     # env (how the top-level `comfy --where` arrives) > config default. Both
-    # the server query and the state-file scope key off this single decision.
-    target_where = "cloud" if _is_cloud(where) else "local"
+    # the server query and the state-file scope key off this single decision,
+    # and it is stamped on the renderer so error envelopes carry it too.
+    target_where = _stamp_where(renderer, where)
 
     # --orphaned only makes sense for state files (the server doesn't know
     # whether a watcher crashed), so skip the server query in that mode.
@@ -1099,7 +1100,7 @@ def status_cmd(
     ] = None,
 ):
     renderer = get_renderer()
-    if _is_cloud(where):
+    if _stamp_where(renderer, where) == "cloud":
         return _cloud_status(prompt_id)
 
     with report_usage_error(renderer, command="jobs status"):
@@ -1523,7 +1524,8 @@ def wait_cmd(
     wait_all: Annotated[bool, typer.Option("--all", help="Wait on all locally-tracked non-terminal jobs.")] = False,
 ):
     renderer = get_renderer()
-    cloud = _is_cloud(where)
+    where_label = _stamp_where(renderer, where)
+    cloud = where_label == "cloud"
 
     ids = list(prompt_ids or [])
     if wait_all:
@@ -1587,7 +1589,6 @@ def wait_cmd(
         "elapsed_seconds": round(time.time() - start, 2),
         "jobs": jobs_list,
     }
-    where_label = "cloud" if cloud else "local"
 
     if renderer.is_pretty():
         _render_wait_pretty(summary)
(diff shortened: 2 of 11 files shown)


bigcat88 (approved):
Approving — verified empirically against the real CLI, not by reading.

## What I ran

A 65-probe matrix driving `python -m comfy_cli --json …` as a real subprocess (isolated `HOME`, `COMFY_LOCAL_URL`/`COMFY_CLOUD_BASE_URL` pointed at a dead port, then at a stub backend answering HTTP 500 so the *handled* error paths were reached), run once on this branch and once with `comfy_cli/` reverted to `origin/main`. Then 26 more probes against the 500-stub for the paths connection-refused can't reach.

**Result: 44 probes flip `where: null` → the correct label; `rc` and `error.code` are identical on both trees in every single probe.** No behavioral change beyond the field, which is what a fix like this should look like.

Covered: `jobs ls/status/cancel/watch/wait`, `workflow list/get/delete`, `workflow slots` + `decompose` (via `COMFY_WHERE`), `nodes ls/show`, `models list-folders/list-folder/search/show`, `system-stats`, `free`, `upload`, `download`, `validate`, `logs`, `assets push` — each on both targets.

The other half of the contract holds too: all 9 pre-routing probes stay `where: null` — `where_invalid` on `upload`/`validate`, `workflow_not_found`, `workflow_invalid_json`, `invalid_argument` from both unsafe-path-segment rejections, and `logs --where cloud`.

## Non-vacuity

Reverted `comfy_cli/` to `origin/main` and ran the new file: **31 of 36 fail**. The 5 that pass are exactly the ones asserting `where is None` — i.e. every positive assertion is load-bearing.

## Other checks

- Full suite: **4654 passed, 31 skipped**. The one failure is `test_non_fast_deps_uses_global_python`, which fails identically on plain `main` in this environment.
- `ruff check` + `ruff format --diff` at the CI-pinned **0.15.15** — clean.
- No cross-command contamination: `run` is the only other stamper, it brackets its stamp with `renderer.where = None`, and it reaches none of the newly-stamped helpers (it calls `resolve_target` directly). `entry()` installs a fresh `Renderer` per invocation, so the singleton bound you documented is real but unreachable from the CLI.
- Events are untouched — only `emit()` and `error()` read `self.where`.
- `docs/json-output.md:100` already promised "Set as soon as routing resolves … `null` for commands that don't route". This PR makes that sentence true rather than changing it; no doc update needed.

## One nit (not blocking)

`nodes ls --input <dump>` now reports `where: "local"`/`"cloud"` on a run where nothing routed at all — mild tension with that same doc line's "`null` for commands that don't route". Your in-code comment makes the case (the mode still selects which catalog the dump is read as) and I'm not asking you to change it; flagging so the choice is on the record.

## Two pre-existing holes this surfaced (both identical on `main` — follow-ups, not this PR)

1. **`jobs` silently swallows an invalid per-command `--where`.** `comfy --json jobs ls --where nowhere` → `ok:true`, routed local. `_is_cloud`'s `except ValueError: return False` comment says "the validating command (cmdline.py top-level option) will surface `where_invalid`" — that's true for `comfy --where nowhere jobs ls` (which does emit `where_invalid`) but not for the per-command flag, which never reaches that validation. This PR now labels it `where: "local"` — truthful about the routing that happened, and it makes the underlying bug easier to see, which is a point in its favour.

2. **`--where nowhere` on `nodes ls`, `models search`, `system-stats`, `workflow list` dies with an unhandled `ValueError` traceback** from `where.py:169 _parse` — exit 1, zero bytes on stdout, no envelope. Same commands with an unreachable/500 backend (`jobs ls --where cloud`, `download`) also escape as a raw urllib traceback. Three different behaviours for one invalid input, and two of them break the envelope/1 "always a terminating envelope" guarantee. Worth a ticket.

PR #719, opened 2026-08-14
feat(nodes): include is_api_node in `nodes ls` and `nodes search` rows

mattmillerai:
## ELI-5

Some ComfyUI nodes are free (they run open weights locally); others are paid partner-API nodes that call an external provider and cost money. `comfy nodes show` already tells you which is which, but `comfy nodes ls` and `comfy nodes search` did not — so an agent scanning search results had to run a second `nodes show` call per candidate just to find out whether picking a node would spend money. This adds one boolean, `is_api_node`, to every `ls` and `search` row so the answer is right there in the result you already have.

## What changed

- `comfy_cli/command/nodes.py` — `"is_api_node": m.is_api_node` added to the row dict in `ls_cmd` and in `search_cmd`. The value comes from `Morphism.is_api_node`, which the CQL engine already parses out of object_info's `api_node` flag and already emits from `morphism_to_dict` (what `nodes show` renders).
- `comfy_cli/skills/comfy/SKILL.md` — a short note in the Nodes section telling agents the flag is on the row, so they stop doing a `nodes show` per candidate. No existing doc enumerated the row field list, so there was nothing else to update.
- `tests/comfy_cli/command/test_nodes_introspect.py` — a new `TestIsApiNodeRows` class following the file's existing fake-`object_info` fixture pattern.

**JSON only.** The pretty rich tables in `ls` and `search` are untouched — no new column — so interactive terminal output is byte-for-byte what it was. The JSON payload is what scripts and MCP wrappers consume, and that is where the field lands.

**Additive only.** No row field is renamed or removed. The close-match ("did you mean…") fallback rows carry the field too, since they are built from the same projection. `--api-only` on `ls` is unchanged — it already filtered on this same model field, and a test now pins that.

`comfy_cli/schemas/nodes.json` types `rows` as a bare array with no per-item schema, so no schema change was required.

## Why the flag matters

Two node classes can share a display name and differ only in whether they are paid. The known case is a partner-API "MiniMax H3 Reference to Video" node and a free open-weights node with the identical display name — indistinguishable in a search row before this change. The new tests model exactly that pair: same `display_name`, one entry with `api_node: true`, one with the key absent entirely.

## Verification

Lint/format with the CI-pinned ruff (`ruff==0.15.15`, per `.github/workflows/ruff_check.yml`): `ruff check .` → all checks passed; `ruff format --diff .` → 283 files already formatted. (Heads-up for anyone reproducing locally: an older ruff in a stale venv reports 19 pre-existing findings in files this PR does not touch — use the pinned version.)

Targeted tests green — `tests/comfy_cli/command/test_nodes_introspect.py`, `test_nodes_cli.py`, `tests/comfy_cli/cql/`, `test_pretty_print_sanitize.py`: **412 passed**. New coverage: the flag on `ls` rows, on `search` rows, `false` (not missing) when object_info omits `api_node`, the flag surviving the close-match fallback path, and `--api-only` still selecting exactly the API nodes.

Beyond the hand-rolled fixtures, I ran the shipped commands end-to-end against a **real captured object_info** already in the repo (`tests/comfy_cli/fixtures/nodes_path_object_info.json`, which contains a genuine partner node, `ByteDanceImageNode` with `api_node: true`):

- `nodes ls --input …` → `ByteDanceImageNode` `is_api_node: true`; the other 11 nodes `false`.
- `nodes search ByteDance --input …` and `nodes search image --input …` → same, correct per row.
- `nodes search BiteDanceImageNode --input …` (typo → close-match fallback) → row carries both `close_match: true` and `is_api_node: true`.

**Full-suite caveat:** the repo's whole `pytest` run exceeded the time budget I had for this change and was still going when I opened this PR, so I am not claiming it green from a local run — CI is the gate on that. Everything I did run is reported above.

## Unexercised artifacts / judgment calls

- **The two specific node classes** that motivated this (`MinimaxHailuo03ReferenceNode` and its free twin `MiniMaxH3ReferenceToVideo`) were **not** exercised by name: neither appears in any object_info fixture or bundled catalog in this repo, and I had no live ComfyUI instance with partner-API nodes loaded to query. What I verified instead is the mechanism they exemplify — a real partner-API node from a real captured object_info (above), plus a fixture pair reproducing their exact shape (shared display name, one `api_node: true`, one with the key absent). Anyone with a live catalog can confirm in one call: `comfy --json nodes search "MiniMax H3 Reference to Video"` and read `is_api_node` off the two rows.
- **`nodes upstream` / `nodes downstream` rows still omit the flag.** Those projections were deliberately out of scope here; this PR changes only `ls` and `search`. Worth a separate look if agents plan graphs off those rows.
- The SKILL.md paragraph is a small addition beyond the letter of the request (no existing docstring or help text enumerated the row fields, so there was no field list to update). It seemed strictly better than shipping a field no agent-facing doc mentions — easy to drop if you'd rather keep the doc untouched.
- No spend/consent code path is touched anywhere in this diff.

### DIFF
--- a/comfy_cli/command/nodes.py
+++ b/comfy_cli/command/nodes.py
@@ -284,6 +284,9 @@ def ls_cmd(
                 "display_name": m.display_name,
                 "output_types": m.output_types(),
                 "output_node": m.is_output_node,
+                # Paid partner-API node vs free open-weights node. JSON only —
+                # the pretty table is deliberately left unchanged.
+                "is_api_node": m.is_api_node,
             }
             for m in nodes
         ],
@@ -553,6 +556,10 @@ def search_cmd(
                 "display_name": m.display_name,
                 "description": m.description,
                 "output_types": m.output_types(),
+                # Paid partner-API node vs free open-weights node — two nodes can
+                # share a display name and differ only here (MiniMax H3). JSON
+                # only; the pretty table is deliberately left unchanged.
+                "is_api_node": m.is_api_node,
                 **({"close_match": True} if close_match else {}),
             }
             for m in matched
--- a/comfy_cli/skills/comfy/SKILL.md
+++ b/comfy_cli/skills/comfy/SKILL.md
@@ -350,6 +350,11 @@ comfy --json nodes ls --produces VIDEO --exclude-deprecated --limit 10
 comfy --json nodes ls --pack core --produces MASK --limit 5
 ```
 
+Every `nodes ls` and `nodes search` row carries `is_api_node` — `true` for a paid
+partner-API node, `false` for a free open-weights one. Two nodes can share a
+display name and differ only in this flag, so read it off the row instead of
+running `nodes show` per candidate.
+
 If no local server is running and you're not signed into cloud, pass
 `--input <object_info.json>` to query against a saved dump.
 
--- a/tests/comfy_cli/command/test_nodes_introspect.py
+++ b/tests/comfy_cli/command/test_nodes_introspect.py
@@ -458,6 +458,87 @@ class TestSearch:
         assert "KSampler" in [r["name"] for r in env["data"]["rows"]]
 
 
+class TestIsApiNodeRows:
+    """`is_api_node` on `nodes ls` / `nodes search` rows.
+
+    The twin-family trap: a paid partner-API node and a free open-weights one can
+    carry the same display name, so an agent picking off a search row needs the
+    flag inline rather than a follow-up `nodes show` per candidate.
+    """
+
+    @pytest.fixture
+    def patched_loader(self, monkeypatch: pytest.MonkeyPatch):
+        from comfy_cli.cql.engine import Graph
+
+        info = {
+            # Paid partner-API twin: object_info carries `api_node: true`.
+            "MinimaxHailuo03ReferenceNode": {
+                "input": {"required": {}},
+                "output": ["VIDEO"],
+                "output_name": ["VIDEO"],
+                "category": "partner/video",
+                "display_name": "MiniMax H3 Reference to Video",
+                "description": "Generate video from a reference image.",
+                "output_node": False,
+                "api_node": True,
+                "python_module": "comfy_api_nodes",
+            },
+            # Free open-weights twin: no `api_node` key at all.
+            "MiniMaxH3ReferenceToVideo": {
+                "input": {"required": {}},
+                "output": ["VIDEO"],
+                "output_name": ["VIDEO"],
+                "category": "video",
+                "display_name": "MiniMax H3 Reference to Video",
+                "description": "Generate video from a reference image.",
+                "output_node": False,
+                "python_module": "nodes",
+            },
+        }
+        graph = Graph.from_object_info(info)
+        monkeypatch.setattr(nodes_cmd, "_get_graph", lambda *a, **kw: graph)
+
+    def test_ls_rows_carry_the_flag(self, patched_loader, capsys):
+        env = _run(["ls"], capsys)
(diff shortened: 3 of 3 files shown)


bigcat88 (approved):
Approving. I had the thing you said you couldn't get — a live ComfyUI with partner-API nodes loaded — so I ran this against it rather than against the fixtures.

## The catalog you didn't have: 792 classes, 204 of them paid

`/object_info` from a real `--cpu` server here carries `api_node` on 655 of 792 classes, 204 of them true. Every row `nodes ls` returned was checked against that ground truth:

| | |
|---|---|
| rows returned | **792** |
| rows carrying `is_api_node` | **792** (0 absent) |
| non-boolean values | **0** |
| disagreements with object_info's `api_node` | **0** |
| true / false | 204 / 588 |

`--api-only` returns exactly those 204, every row `is_api_node: true`, and the set is *equal* to the ground-truth set — not merely a subset. Both halves of your `--api-only` test hold at catalog scale.

The `bool(raw.get("api_node", False))` in `Morphism` is what makes this a contract rather than a passthrough: 137 of those classes omit the key entirely and still emit `false`, never `null` and never absent. That is the property your third test pins, and it survives real data.

## No regression: 14 verbs, everything except the new field

Ran `ls`/`search`/`show`/`upstream`/`downstream`/`path`/`packs` — with `--limit`, `--api-only`, `--produces`, `--exclude-deprecated`, a typo triggering the close-match path, and a no-match query — on `main` and on this branch against the same catalog, then stripped `is_api_node` from both and compared:

```
14 verbs compared, 0 differ outside the new field
```

rc, `ok`, and error code identical on every one.

**Pretty output is byte-identical**, which I checked properly rather than by piping: piping makes the CLI auto-select JSON (non-TTY rule), so a naive capture never reaches the table you're claiming is untouched. Under a real pty with `--no-json`, `ls` / `ls --api-only` / `search` / `show` / `upstream` are byte-for-byte equal to `main` — 5536, 3122, 6898, 1637 and 2141 bytes, all `cmp`-clean.

Your schema claim checks out too: `rows` in `nodes.json` is `{"type": "array"}` with no item schema, and there is no `additionalProperties` anywhere in the file — so an added field cannot fail validation.

Suite: **4654 passed**, same two failures as clean `main` in this environment (`test_non_fast_deps_uses_global_python`, and a `test_onboarding` full-suite ordering failure that reproduces identically on `main`). `ruff check` + `ruff format --diff` clean at the CI-pinned 0.15.15.

## One follow-up, not blocking

You flag `upstream`/`downstream` as deliberately out of scope, and I'd normally leave it — except those are the projections your own rationale argues hardest for. `nodes upstream KSampler` returned **254 rows** here and `downstream` **57**, none carrying the flag; that is the exact "agent scans rows and picks a candidate" moment, at 20x the row count of a search. And it is the same one-liner in `_morphism_row`, which both verbs already share:

```python
def _morphism_row(m) -> dict[str, Any]:
    return {
        "name": m.id,
        "category": m.category,
        "display_name": m.display_name,
        "output_types": m.output_types(),
        "is_api_node": m.is_api_node,
    }
```

`path_cmd` doesn't use that helper, so the blast radius is just those two verbs. Additive either way, so it loses nothing to land separately — but the inconsistency is user-visible right now: `rows` carries the field or not depending on which verb produced it.

Worth noting your motivating pair (`MinimaxHailuo03ReferenceNode` / `MiniMaxH3ReferenceToVideo`) isn't in this catalog, and in fact it has **zero** display-name collisions spanning a paid and a free node — so I can't confirm that specific twin, only the mechanism. Your fixtures model the shape correctly, and the real ByteDance/Beeble/Bria/Flux partner nodes exercise the same path.

PR #722, opened 2026-08-17
fix(workflow_to_api): apply subgraph instance promoted-widget values; primitive-fed inputs beat stale widget residue

guillaume86:
## Problem

Two related value-resolution bugs in `convert_ui_to_api` made conversion silently diverge from the frontend's `graphToPrompt`:

**1. Subgraph instance promoted-widget values were dropped.** A curated subgraph exposes unlinked widget-type def inputs as instance widgets; the frontend substitutes the instance's `widgets_values` into the interior nodes at queue time. The converter never did — interior *definition defaults* ran instead. Anything edited on the instance (prompt, seed, model names) was silently ignored.

This bites every official template built as a curated subgraph. Concrete repro: run any official MiniMax H3 video template through `comfy run --workflow` and inspect the executed prompt in `/history` — the definition's placeholder prompt ("Vaporwave title sequence look…") executes instead of the instance prompt the template actually ships ("Realistic live-action cinematic look…" for t2v; for i2v the instance prompt references the example input image, so the generated clip ignores the image entirely). Found in the wild while reproducing the H3 templates on a local install; execution-history diffs confirmed every subgraph-template render used placeholder values.

**2. Primitive-fed inputs lost to stale widget residue.** Assembly preferred `widget_inputs` over `primitive_inputs`, so an input converted-to-input and wired from a `PrimitiveNode` still took the consuming node's leftover `widgets_values` entry whenever one existed. Frontend semantics: the incoming value wins; the widget slot is residue.

## Fix

- During subgraph expansion, synthesize a virtual `PrimitiveNode` per promoted instance value (one `widgets_values` entry per widget-type def input, in def order — connection-only types contribute no entry) and wire it to the def input's interior targets via `linkIds`. The existing primitive-value machinery injects the value and excludes the virtual node from output, exactly as for user-placed primitives. Def inputs that are externally linked on the instance are skipped (link wins).
- Assembly precedence flipped: `primitive_inputs` before `widget_inputs` (also in the unknown-order preservation loop). This is what makes both the synthesized and ordinary user-placed primitives override widget residue.

## Tests

`TestSubgraphPromotedWidgets` (4 cases): instance value overrides interior default; virtual primitive not emitted; external link beats stale instance widget; missing instance widgets keep interior defaults. Full `test_workflow_to_api.py` (101), `test_subgraph_gallery_templates.py` + `cql` suites pass; `ruff check`/`format` clean on touched files.

https://claude.ai/code/session_01WDDgJAus5uLUGWNsLcy3zr

### DIFF
--- a/comfy_cli/workflow_to_api.py
+++ b/comfy_cli/workflow_to_api.py
@@ -473,11 +473,13 @@ def _inject_promoted_widget_values(
     # length disagreement means we cannot say which value belongs to which
     # input. Falling back to the interior defaults is wrong but self-consistent
     # and executable; a misaligned guess submits a prompt string into an INT
-    # slot. Prefer the recoverable failure.
+    # slot. Prefer the recoverable failure — but warn, because every value the
+    # instance carries is being dropped and the run would otherwise look fine.
     if len(promoted) != len(inst_widgets):
-        logger.debug(
-            "Subgraph instance %s: %d promoted widget(s) but %d widgets_values — "
-            "skipping instance-value injection, interior defaults kept",
+        logger.warning(
+            "Subgraph instance %s exposes %d promoted widget(s) but carries %d "
+            "widgets_values; cannot align them, so the instance's values are "
+            "ignored and the subgraph's interior defaults run instead",
             outer_id,
             len(promoted),
             len(inst_widgets),
--- a/comfy_cli/workflow_to_api.py
+++ b/comfy_cli/workflow_to_api.py
@@ -402,13 +402,39 @@ def _expand_one_subgraph(
     return expanded_nodes, expanded_links, input_targets, output_sources
 
 
-# Def-input types the frontend materializes as widgets on a subgraph instance.
-# The instance's ``widgets_values`` array holds one entry per def input of
-# these types, in def-input order (connection-only types such as IMAGE/MODEL
-# contribute no entry). Entries whose def input also has an external link on
-# the instance are stale residue — the link wins, same as any widget-behind-a-
-# link on a regular node.
-_PROMOTED_WIDGET_TYPES = frozenset({"STRING", "INT", "FLOAT", "BOOLEAN", "COMBO"})
+def _promoted_widget_def_indices(
+    sg_def: dict,
+    input_targets: dict[int, list[tuple[Any, int]]],
+    node_by_id: dict[Any, dict],
+    outer_id: Any,
+) -> list[int]:
+    """Def-input indices that materialize as widgets on a subgraph instance.
+
+    The instance's ``widgets_values`` array holds one entry per promoted
+    widget, in def-input order. A def input is widget-backed exactly when the
+    interior input it feeds is itself widget-backed — which the serialized
+    graph marks with a ``widget`` key on that interior input slot. Deriving it
+    from the interior (rather than guessing from the def input's declared
+    type) is what keeps the indices aligned: a promoted widget can carry any
+    type at all, including node-pack types like ``COMFY_DYNAMICCOMBO_V3``, and
+    an allowlist that misses one shifts every later value into the wrong slot.
+    """
+    indices: list[int] = []
+    for def_idx, in_def in enumerate(sg_def.get("inputs") or []):
+        if not isinstance(in_def, dict):
+            continue
+        for target_id, target_slot in input_targets.get(def_idx) or []:
+            target = node_by_id.get(f"{outer_id}:{target_id}")
+            if not isinstance(target, dict):
+                continue
+            inputs = target.get("inputs") or []
+            if not (isinstance(target_slot, int) and 0 <= target_slot < len(inputs)):
+                continue
+            slot = inputs[target_slot]
+            if isinstance(slot, dict) and slot.get("widget"):
+                indices.append(def_idx)
+                break
+    return indices
 
 
 def _inject_promoted_widget_values(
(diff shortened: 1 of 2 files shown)


guillaume86:

bigcat88 (changes requested):
Requesting changes. The bug is real and worth fixing, and the mechanism you chose — synthesize a virtual `PrimitiveNode` and let the existing primitive machinery inject it — is the right shape. But **the index derivation is wrong**, and on real templates it doesn't produce the definition defaults instead of instance values; it produces *other inputs'* values.

## What it does to the official template set

I converted all **459** official templates (`comfyui_workflow_templates_json`, 153 of them carrying subgraph definitions) on `main` and on this branch and diffed the resulting API prompts.

**17 templates change. All 17 are corrupted. None are fixed.**

The clearest one, `video_ltx2_i2v.json`:

```
CLIPTextEncode.text  : 'A close-up shot of a y…'  ->  '121'
PrimitiveInt.value   : '121'                      ->  'A close-up shot of a young waitress…'
```

The prompt and the frame count are **swapped**. Type-mismatch count across the corpus: `main` 8 (in 4 templates, pre-existing), this branch **23 (in 14)**. Ten templates newly acquire a value whose type contradicts the declared input type — `CFGGuider.cfg <- 'capybara_v0.1.safetensors'`, `ControlNetApplyAdvanced.strength <- 'qwen_2.5_vl_7b_fp8_scaled.safetensors'`, `EmptyImage.width <- "squish it -- a woman's voice…"`.

That count is a **lower bound**, because a wrong value of the *right* type is invisible to it:

```
audio_ace_step_1_5_split_llm.json
   timesignature : '4'  -> 'en'
   language      : 'en' -> '4'
```

So this is worse than the bug it fixes. Today those templates run with the interior defaults — wrong values, but self-consistent ones that execute. With this change they'd be submitted with a prompt string in an INT slot.

## Why the index is wrong

The assumption is that the instance's `widgets_values` holds one entry per def input of a promoted type, in def-input order. Measured against the 207 real subgraph instances:

| | |
|---|---|
| `widgets_values` empty → early return, harmless | 159 |
| length matches promoted-def-input count | 34 |
| **length disagrees** | **14** |

But length agreement isn't sufficient — `audio_ace_step_1_5_split_llm` above is in the *aligned* 34 and is still swapped. So it isn't an off-by-N; **def-input order is simply not the order the instance stores its widget values in.**

Concretely, `video_ltx2_i2v.json`'s def inputs are `value(INT), text(STRING), image(IMAGE), ckpt_name, text_encoder, lora_name, model_name` — 6 promoted — while the instance carries **8** `widgets_values`, starting `['A close-up shot…', 121, None, None, …]`. The string comes first, so the frontend is not walking def inputs in this order, and there are two more values than promoted inputs to receive them.

Two things in the data that may point at the real rule: the subgraph definition has its own top-level **`widgets`** key (empty `[]` in these templates, so it isn't populated here but looks like the intended home for a promotion list), and the instance's `inputs[]` entries carry an explicit **`{"widget": {"name": …}}`** marker distinguishing widget-backed inputs from link inputs. Deriving the mapping by *name* from one of those, rather than positionally from def-input order, is what would make this robust — and it would fail closed rather than silently mis-assign when the shapes disagree.

## Why the tests pass anyway

`TestSubgraphPromotedWidgets` builds a definition with two def inputs — `pixels` (IMAGE, not promoted) and `text` (STRING, promoted). With exactly **one** promoted input, every possible ordering rule produces the same answer, so no ordering bug can surface. The four cases are good ones for the behaviours they target (override, virtual node not emitted, link wins, missing values), but none of them can fail on order.

A fixture with ≥3 promoted inputs of *different* types, in an order that differs from the instance's, would have caught this — and asserting the resulting value types match the declared input types would catch the whole class.

## The second fix looks fine, but nothing here exercises it

The `primitive_inputs` before `widget_inputs` flip changed **zero** of the 459 templates — every one of the 17 diffs traces to the subgraph injection. So the reordering is not a regression risk on this corpus, but it also isn't confirmed by it; your own unit tests are the only evidence for it. Worth splitting into its own PR — it's independently correct-looking and would land immediately, instead of waiting on the subgraph work.

## Suite

`pytest tests/comfy_cli` on this branch: ****4763 passed**, the only failure being `test_non_fast_deps_uses_global_python`, which fails identically on clean `main` here — so your four new tests do pass, which is rather the point**. `ruff check` + `ruff format --diff` clean at the CI-pinned 0.15.15.

Happy to re-run the full-corpus differential once the ordering is reworked — it's scripted now, and "17 templates change, all 17 corrupted" turning into "N change, all N correct" is exactly the signal this needs.

guillaume86:
Thanks — this was exactly the right call, and the differential is what made it findable. Pushed a rework in a47806c.

## You're right that it's broken. The diagnosis is off by one step

It isn't that def-input order is the wrong order. The order is right; the **set of def inputs that count as promoted** was wrong.

Promotion was decided from an allowlist of def-input *types* — `{STRING, INT, FLOAT, BOOLEAN, COMBO}`. A promoted widget can carry any type at all, so `InfiniteTalk`'s `mode` (`COMFY_DYNAMICCOMBO_V3`) fell through, consumed no slot, and shifted every later value by one. That is your `audio_scale <- "two_speakers"` and `text <- 1`, exactly.

Embarrassingly, this module already has a careful widget classifier — `_is_widget_spec`, which handles `COMFY_*COMBO*`, `forceInput`, wildcards, lowercase custom types. The subgraph path was reimplementing a worse copy of it.

## The rule

A def input is promoted exactly when the interior input it feeds is itself widget-backed — which the serialized graph already marks with a `widget` key on that interior input slot. So read it off the interior rather than guessing from the declared type. No object_info needed, and no second classifier to keep in sync.

Measured over every subgraph instance in the corpus:

| rule | `len(promoted)` == `len(widgets_values)` |
|---|---|
| type allowlist (old) | 76 / 78 |
| interior `widget` marker (new) | **78 / 78** |

Length agreement isn't sufficient — your `audio_ace_step_1_5_split_llm` point stands — so I also checked the assignments themselves: **519 marker-derived assignments, zero type mismatches** against the declared def-input types.

## Differential

Same method as yours, 499 templates, real `object_info` from a live instance:

|  | main | previous revision | now |
|---|---|---|---|
| templates changed vs main | — | 13 | **12** |
| type mismatches | 8 (in 4) | 12 (in 5) | **8 (in 4)** |

Back to main's baseline. Those 8 are the pre-existing `Load3D` ones (`width <- "upload3dmodel"`), untouched by this PR and a separate bug.

All 12 changed templates are corrections: **34 changed values, all 34 traceable to an instance `widgets_values` entry**, none to a definition default.

`video_wan2_1_infinitetalk` now converts identically to main — not because injection is skipped, but because that template's instance values happen to equal its interior defaults. Verified value-by-value.

One caveat I can't close from here: I'm on `comfyui_workflow_templates_json` 0.1.47 (499 convertible templates), you were on a corpus of 459 with 153 carrying subgraph defs. Our `video_ltx2_i2v.json` differ structurally — mine has 8 def inputs / 7 promoted / 7 `widgets_values`, yours 7 / 6 / 8 — so I verified the *class* of bug you found, on a different snapshot. Your offer to re-run stands very welcome.

## Fail closed

Added, and it's the answer to the shape I can't see in your corpus: on any count disagreement, injection is skipped entirely and the interior defaults are kept. Wrong but self-consistent and executable, which is the trade you argued for. So the worst case on an unfamiliar shape is "no fix applied", not "prompt string in an INT slot".

## Tests

`TestSubgraphPromotedWidgetOrdering` — five promoted inputs of five distinct, mutually type-incompatible values, interleaved with connection-only inputs so a miscount shifts the tail. Plus the type-conformance assertion you asked for, the external-link-still-consumes-its-slot case, and the fail-closed case.

All five fail on the previous revision. Your read on why the old four couldn't catch it was correct — with one promoted input every ordering rule agrees.

## On splitting the precedence flip

Can't go second — it's load-bearing for the subgraph fix. With the injection but without the flip, five tests fail: the synthesized primitive loses to the interior node's stale `widgets_values` entry, which is the whole mechanism.

It also isn't unexercised on real data, though your corpus was right that it's nearly invisible. Flip-only changes exactly one template here — `template_contact_sheet-step_3.app.json`, `KlingFirstLastFrameNode.duration` 3 → 5, where 5 comes from a user-placed primitive reaching the interior through a Reroute and a promoted input, and 3 is the residue.

Happy to land it *first* as its own PR and rebase this on top if you'd prefer the smaller review — just not the other way around.

## Suite

`pytest tests/comfy_cli` — 4776 collected. Everything passes except six pre-existing failures that reproduce identically on clean `main` here (`test_file_utils` 2, `test_global_python_install` 1, `test_host_port` 3), and `command/test_run.py::TestLocalExecuteItemMapAndGroupedOutputs`, which SIGKILLs the runner on this box on clean `main` too; the other 132 in that file pass. `ruff check` + `ruff format --diff` clean at 0.15.15.

guillaume86:
Closing this one — upstream landed the same fix independently in ce5c2a4 / 088d151 / f0d5c4b (2026-08-28). `main` now carries `_overlay_promoted_host_values()` alongside `tests/comfy_cli/test_workflow_to_api_promoted.py`, which covers the case this PR was aimed at, so there's nothing left here worth rebasing.

@bigcat88 — thanks for the review. The call on the index derivation was correct, and I never circled back; sorry for leaving the thread hanging.

PR #731, opened 2026-08-18
feat(distribution): serverless distribution commands (limited beta)

sundar-svg:
## Description

Adds the `comfy distribution` command group — packages a local ComfyUI
environment into a serverless distribution and drives builds on the Developer
Platform builder. The same builder operations are exposed in two shapes: a
coarse `create` for a human at a terminal, and granular one-op-per-command tools
an agent composes. Every command emits the structured JSON envelope and
registers a `discover` output schema.

Commands:
- `scan` — walk `models/` + `custom_nodes/` into a builder-ready definition JSON
  (streaming SHA-256 per model, git `repo@ref` per node, pip freeze + a
  source-platform header). Fully local, no backend.
- `create --from <def>` — resolve public models to URLs (hash-verified), upload
  private ones as blobs, create the distribution, and cut a build. Preview by
  default; `--execute` goes live. Fails fast when the definition has no
  `baseComfyVersion` (which the build requires).
- `list` / `get <id>` / `validate <id>` (dry-run resolve) / `delete <id>`
  (soft-delete; confirmation or `--yes`).
- `version create <id>` (cut a build) / `version list <id>` /
  `version get <vid>` / `version logs <vid> [--os --gpu]`.

Auth reuses the `comfy cloud login` OAuth token (the builder validates the Cloud
JWT by signature). Builder access is **limited beta**, enforced server-side: a
caller whose account isn't enabled gets a clear `distribution_not_enabled`
("limited beta") message rather than the command being hidden.

## How tested

- 60 unit tests (scan/hash helpers, definition mapping, builder-client parsing +
  URLs, error mapping incl. the beta-403 and the presigned-upload header,
  the missing-version preflight, delete confirmation) — all green.
- `ruff check` / `ruff format` clean; the `discover` schema<->command bijection
  and the error-code registry both pass their CI gates.
- Exercised end-to-end against the staging builder: scan -> resolve -> a 319 MB
  private-blob upload -> create -> cut -> list/get/version/logs -> validate ->
  delete.

## Not in this change

`update`, `version manifest`, `artifact download`, `blob list`, and the reference
lookups (`base-images` / `build-targets` / `model-dirs`) are the next tranche.
Local (non-git) custom-node upload is deferred (v1 assumes git-based nodes).

---

https://claude.ai/code/session_01KV7wUb2hyzKHsR52ZJTvAm


james00012:
LGTM once the first two are addressed. This is a lot of surface and it reads carefully, and the scan half in particular is the part only this tool can do.

**What this does, as I read it**: `scan` reads the local install into a definition, `create` resolves what is public, uploads what is not, creates the distribution and cuts a build, and the read verbs wrap the rest of the builder's `/v1`.

Blockers:

* **A single hand-dropped custom node wastes the whole upload.** `plan_create` appends every model upload before any `node_zip` upload, and `execute_create` walks that list in order, so a definition with one non-git node uploads every private model first and only then raises `NotImplementedError`. Nothing is created, and on a real install that is potentially hours and hundreds of gigabytes. Preflighting the list before any bytes move is a few lines next to the `baseComfyVersion` check that already exists, and it also covers the missing-`--models-dir` case that today fails partway through the same loop:

  ```python
  paths = []
  for u in plan["uploads"]:
      if u["kind"] != "model":
          raise NotImplementedError(f"uploading a local custom node ({u.get('name')}) isn't supported yet")
      paths.append(locate_bytes(u))
  ```

  Both exceptions are already mapped to `distribution_upload_unavailable`, so no new error code. `comfy_cli/command/distribution.py:487`. The test at `tests/comfy_cli/command/test_distribution.py:412` passes `models: []`, so it never sees the ordering; one model plus one local node, asserting `create_blob` was never called, would.
* **`create --execute` emits a payload its own registered schema rejects.** `distribution_create.json` requires `plan`, and the execute path emits `{name, executed, distributionId, versionId, statusUrl, uploaded}` with no `plan`. That schema is what an agent reads out of `comfy discover` to know the shape, so the one path that changes state is the one that does not match. Either make the executed envelope carry `plan` too, or split it with a `oneOf` on `executed`. `comfy_cli/schemas/distribution_create.json:6`, `comfy_cli/command/distribution.py:887`.

Nits / non-blockers:

* **The limited-beta 403 branch never fires, and `create` could not use it anyway.** comfy-builder has no `FEATURE_NOT_ENABLED`: the only string of that name in `cloud` belongs to ingest's provider-policy governance. What actually blocks an unentitled caller is `402 SUBSCRIPTION_REQUIRED`, at `services/comfy-builder/apiserver/distributions/distributions.go:72` and `versions/versions_cut.go:133`. Separately, `_create_execute` does not route through `_builder_call` at all, so even a matching 403 would land in the generic handler, and today a builder `400 INVALID_DEFINITION` reaches the user as urllib's `HTTP Error 400: Bad Request` rather than the builder's own message. Routing create through `_builder_call` fixes the second half regardless of what you decide about the first.
* **A failed cut leaves an orphan the message does not name.** `execute_create` creates the distribution, then cuts. If the cut fails, and `402 SUBSCRIPTION_REQUIRED` is raised at exactly that point, the distribution and every uploaded blob exist but the error carries no `distributionId`. Recoverable through `comfy distribution list`, so this is only about putting the id in `details`.
* **`_load_definition` reads without an encoding.** `atomic_write_text` writes UTF-8 (`comfy_cli/file_utils.py:145`) and `path.read_text()` reads the platform default, so a model filename with non-ASCII characters round-trips fine on macOS and Linux and fails on Windows as "not valid JSON". Windows is in the test matrix. `comfy_cli/command/distribution.py:743`.
* **`_freeze_env` calls `uv` by bare name.** `_git_output` in the same file goes through `resolve_required_binary`, and `_safe_exec`'s docstring gives the Windows CWD-search reason; `comfy_cli/uv.py:71` already uses `shutil.which("uv")`. `comfy_cli/command/distribution.py:313`.
* **Every scanned filename goes to HuggingFace and CivitAI.** `resolve_models_via_builder` sends the whole set, and the builder's own spec says the value becomes a live provider query, so the names of private fine-tunes leave the machine on `create --execute`. Worth confirming that is intended rather than changing.

Checked and not a problem:

* **The builder URL and the auth story hold.** `platformapi.comfy.org/builder` plus `/v1/<path>` is exactly what the platform website's proxy builds (`platform/server/api/builder/[...path].ts`, `.env`), and the builder accepts audience `comfy-cloud` (`services/comfy-builder/config/config.go:222`), which the CLI's OAuth session is bound to via RFC 8707. Verified by reading both sides, not by a live prod call.
* **Every request and response field matches the builder's OpenAPI**, including the 32-filename resolve cap and the `x-goog-if-generation-match: 0` header being signature-bound.
* **`validate` claiming "Definition resolves" on a 200 is accurate.** The handler only ever returns `Ok: true` on 200 (`apiserver/validate/validate.go:48`); warnings ride alongside and print with the rest of the JSON.
* **Three CodeRabbit findings I would ignore.** Its "skip candidates lacking a truthy `sourceUri`" is against the spec, where `sourceUri` is required on `ModelSourceCandidate` (`services/comfy-builder/openapi.yaml:606`). Its redirect finding overstates the risk: the PUT carries no credential and the host comes from our own builder's authed response, so `allow_redirects=False` is tidy, not a data-handling hole. And the 120-character line in `error_codes.py` is a formatting note that ruff already passed.

james00012 (approved):
Approving so this isn't sitting on me.

The two blockers from [my earlier comment](https://github.com/Comfy-Org/comfy-cli/pull/731#issuecomment-5332414726) still stand and I'd like them in before merge: the upload preflight, and the `create --execute` payload matching `distribution_create.json`. Both are small and I don't need to see them again.

The nits are yours to take or leave.

james00012:
Follow-up, not a merge blocker: my approval stands. This is about what has to happen next for the agent to be able to use any of this.

**Nothing server-side can drive these commands today.** Both client constructions go through `BuilderClient.from_session` (`distribution_api.py:54`, used at `distribution.py:846` and `:901`), which reads only the on-disk `comfy cloud login` session. There is no flag and no environment override, so a process holding a caller's token has no way in.

**The gap is wider than the builder client, which is why I would not fix it here.** `resolve_cloud_credential` has four slots: an explicit flag, the OAuth session, `COMFY_CLOUD_API_KEY`, and the stored key. A forwarded Cloud JWT fits none of them, and the explicit flag is typed `api_key` so it would go out as `X-API-Key` — which comfy-builder ignores outright, since `httpkit/identity.go` reads `Authorization: Bearer` only. So this is not "add a `--token` flag to the distribution commands"; it is a fifth slot in the shared resolver returning `kind="oauth"`, which every cloud command gains at once. Doing it inside `distribution_api` instead would also fight the ratchet test that rejects new direct environment and session reads outside `credentials.py`.

**One thing worth someone confirming, because it may already bite.** The agent forwards the caller's credential by type: a `comfyui-` key as `COMFY_CLOUD_API_KEY`, and anything else — a Firebase or Cloud JWT — as `COMFY_CLOUD_AUTH_TOKEN` (`cloud/services/agent/internal/loop/loop.go:1087`). That second name does not appear anywhere in comfy-cli. If forwarded credentials are in practice always `comfyui-` keys then nothing is broken and this is only about the new commands; if they are not, the cloud path is already failing closed for reasons that have nothing to do with this pull request. I did not verify which, and whoever owns the agent will know in a minute.

**And the piece that makes the result usable:** `GET /v1/blobs` (`listBlobs`, with `limit` and `kind`) already exists on the builder, so `blob list` is a thin read on top of the client this pull request adds. Worth pairing with the auth change, since a server that can upload private models but cannot enumerate them has no way to reconcile what it already sent.

Happy for all of this to be a follow-up. Flagging it here so it lands on someone rather than being discovered when the agent first tries to build a distribution.

sundar-svg:
Thanks @james00012 — both blockers are in, plus the nits. Ready for another look.

**Blockers**
- **Upload preflight** (`587afb5`) — `execute_create` now validates the whole upload list (an unsupported local node, or a model whose file isn't found via `--models-dir`) *before* any bytes move, so one hand-dropped node no longer uploads every private model first and then raises. Test added for the one-model-one-node case.
- **`--execute` payload vs `distribution_create.json`** (`04fe7dc`) — the schema now uses a `oneOf` on `executed`, so both the preview (`plan`) and executed (`distributionId`/`versionId`/…) shapes validate.

**On the 403 / error surfacing**
`FEATURE_NOT_ENABLED` exists now — [cloud#6952](https://github.com/Comfy-Org/cloud/pull/6952) (merged) adds a server-side `403 FEATURE_NOT_ENABLED` on comfy-builder's `/v1` group (the `distributions_enabled` private-beta gate), which is exactly what this branch keys on. Separately, `create --execute` now routes through the same `_report_builder_error` the read verbs use (`587afb5`), so a builder `400 INVALID_DEFINITION` / `402 SUBSCRIPTION_REQUIRED` / `403` surfaces the builder's own message instead of urllib's opaque `HTTP Error 400`.

**Other nits** — orphan `distributionId` now rides in the error `details`, definitions/version-markers read as utf-8, and `uv` resolves through `resolve_required_binary` (same rule as `git`).

**Filenames → HF/CivitAI** — confirmed intended. Only the model *filenames* go (via the builder's resolve search), never bytes or hashes, and only on `create --execute`; it's how we skip re-uploading public models. Noted as a compromise; happy to add a `--no-resolve` opt-out if a user with sensitive filenames asks.

**Passing your own JWT (from the Slack thread)** — done (`3664985`): `COMFY_BUILDER_TOKEN` lets the agent service / CI inject the caller's Cloud JWT and skip `comfy cloud login`. Env var (not a flag) so it stays out of process listings.

james00012:
Last one from me, and it is the only thing left that I think has to change here rather than on my side. Preflight and the `oneOf` both look right, and carrying `distributionId` on a failed cut was a nice extra.

**`version logs` cannot return a large build log.** The builder serves up to 8 MiB (`maxServedBuildLogBytes = 8 << 20`, `apiserver/versions/versions_logs.go:18`, matching the 8 MiB persisted cap in `temporal/activities/sandbox/buildlog.go:155`). `_MAX_JSON` in `distribution_api.py` caps every builder response at 5 MB. Anything between the two is served by the builder and thrown away by the client. Because the log is a JSON string, escaping makes it bite earlier than 5 MB of actual log: every newline costs two bytes, and a build transcript is mostly newlines.

**The failure mode is worse than the cap.** `read_capped` raises `ResponseTooLarge` (`http.py:219`, raised at `:265`), which subclasses bare `Exception`. `_builder_call` catches `URLError`, `RequestException` and `KeyError`, so this escapes as an unhandled traceback and a non-envelope exit rather than the coded error every other failure produces.

Two asks, and the second matters more than the first:

* **Give the logs call its own larger cap**, above 8 MiB so the builder's own bound is the only one that applies. The other endpoints return ids and status and are fine at 5 MB.
* **Add `ResponseTooLarge` to `_builder_call`** (and to `_create_execute`'s except clause). Then whatever the cap is, an oversize response degrades to an error envelope instead of a stack trace.

Why I care about this one specifically: [Comfy-Org/cloud#6973](https://github.com/Comfy-Org/cloud/pull/6973) gives the agent six read tools over these commands, and reading a failed build's log is the case the whole thing exists for. That PR bounds the log to a 64 KiB tail, but it bounds what the CLI hands back, so it never gets the chance. A build noisy enough to be worth asking about is exactly the one that trips this.

Everything else on my side is mine to fix, not yours: I will match the token variable name and drop `--limit`, since the builder's default of 20 newest is enough for what the agent does with it.

PR #732, opened 2026-08-18
feat(distribution): remaining builder commands (update, references, manifest, artifact, blob)

sundar-svg:
## Description

Completes the CLI's coverage of the comfy-builder `/v1` surface, on top of the scan / create + read / version verbs merged in #731:

| Command | Builder op |
|---|---|
| `distribution update <id> --from <def>` | `updateDistribution` (PATCH the stored definition) |
| `distribution resolve <filenames…>` | `resolveModelSources` (name → public download candidates) |
| `distribution base-images` / `build-targets` / `model-dirs` | the reference lookups (authoring inputs) |
| `distribution version manifest <vid>` | `getDistributionVersionManifest` |
| `distribution artifact download <id>` | `getBuildArtifactDownload` (signed link) |
| `distribution blob list` | `listBlobs` |

New `artifact` and `blob` sub-apps follow the same resource-verb shape as `version`. Every command emits the JSON envelope and registers a `discover` schema (the schema↔command bijection gate passes); client methods mirror the openapi field names.

**Deferred (follow-ups):** `version add-target` (build another target for a cut version) and poll-to-completion after `version create`.

## How tested

- New unit test covers the reference/list/update/manifest/artifact client methods (URLs + methods + parsing).
- `ruff check` / `ruff format` clean; discovery schema↔command bijection and the error-code registry pass their CI gates. 66 tests green.

https://claude.ai/code/session_01KV7wUb2hyzKHsR52ZJTvAm


james00012 (approved):
LGTM once the first one is fixed. Read at `15f973e`. Seven of the eight commands are exact passthroughs and I checked every path, method and envelope key against the builder's spec; the one that writes is the one that is broken.

**What this does, as I read it**: adds the eight builder operations #731 left out, being the definition edit, the three authoring menus, a version's manifest, an artifact's signed download and the blob listing, each with its envelope and its discovery schema.

Blockers:

* **`distribution update` can never succeed; it returns 409 STALE on every call.** `UpdateDistributionRequest` requires `expectedUpdatedAt` and the client sends only `{"definition": ...}` (`comfy_cli/distribution_api.py:182`). There is no request-validating middleware in front of the builder, so the missing field decodes to Go's zero time (the generated field is a bare `time.Time`, not a pointer, so absent and year 1 are indistinguishable). The handler then puts that value straight into a predicate-guarded update, `Where(distribution.UpdateTimeEQ(0001-01-01))`, which matches no row, and ent's not-found is mapped to `409 STALE`: `services/comfy-builder/apiserver/distributions/distributions_update.go:64`. The builder's own test picks a deliberately impossible timestamp for that path and comments it `// never matches` (`distributions_db_test.go:82`). Fix is a read-modify-write: `GET /v1/distributions/{id}`, take `updatedAt` off the response, send it back as `expectedUpdatedAt`. That is last-writer-wins for a CLI, which is the right semantic here; the check exists for the website, where a human sits between the read and the save. Worth noting the new test pins the bug rather than catching it: `test_distribution.py:619` now asserts the PATCH body is exactly `{"definition": {"models": []}}`.

Nits / non-blockers:

* **`update` inherits `_load_definition`'s `models` requirement.** `comfy_cli/command/distribution.py:1167` rejects any definition file without a `models` key (`distribution.py:767`). That is a fair guard on `create`, which came from `scan`, but a nodes-only or pins-only edit is a normal thing to push through `update`, and it fails client-side with "not a distribution definition".
* **`blob list` drops the spec's `kind` filter.** `GET /v1/blobs` takes `kind=node_zip|model`; a `--kind` flag is a one-liner and is the difference between listing models and listing everything. Not the `--limit` question from last time, which stays settled.
* **The reference lookups print raw JSON while `model-dirs` prints a list.** `base-images` and `build-targets` go through `print_json`, `model-dirs` prints one line per entry (`distribution.py:1197`, `1208`, `1219`). Both are fine; picking one would be nicer.

Checked and not a problem:

* **Every path, method and envelope key matches the spec.** `/v1/base-images` to `baseImages`, `/v1/build-targets` to `targets`, `/v1/model-directories` to `directories`, `/v1/blobs` to `blobs`, plus `/v1/distribution-versions/{id}/manifest` and `/v1/build-artifacts/{id}/download` returning `{downloadUrl, expiresAt}`, and `update` is the right verb on the right path. Read against `services/comfy-builder/openapi.yaml` on `origin/main`, not against a live deploy.
* **Nothing here opens a tenancy hole.** Both new id-taking reads are workspace-scoped server-side and return 404 rather than 403 for another workspace's id: `versions/artifacts_download.go:33` walks the artifact's owning version, and the manifest handler does the same. The CLI passing a bare id is correct, because the guard is the builder's, and it holds.
* **The resolve batching fix is right.** `_chunks(..., _RESOLVE_BATCH)` matches the builder's own `maxResolveFilenames = 32` (`apiserver/menus/model_resolve.go:27`), and results stay in request order because the flatten preserves batch order.
* **The discovery gate really passes.** Ran `test_distribution.py` and `test_discovery.py` at `15f973e` locally: 60 passed, schema to command bijection included.
* **`artifact download` is usable.** The artifact id it wants is in what `distribution version get` already emits, so there is no dead end where you cannot get the id.

PR #734, opened 2026-08-20
fix(utils): apply the stdlib data filter on both tarball extraction paths

### LINKED ISSUE #725: `tar.extractall()` runs with the extraction filter disabled on both paths
**What** — `extract_tarball()` calls `tar.extractall(filter=None)` on the non-progress path, and on the progress path passes a custom `_filter` that returns `tinfo` unmodified. Both bypass Python's `data_filter`, so a tarball member with an absolute path or `../` traversal is written wherever it points.

**Why it matters here** — This is the CVE-2007-4559 class. It matters here because `StandalonePython.FromTarball(fpath)` accepts an arbitrary caller-supplied path, and `FromDistro()` downloads from a configurable `asset_url_prefix` rather than a pinned source. The default path is a trusted HTTPS distro, which is why this is `recommended` and not `critical` — but the entry point is general.

Worth being clear: this was a knowing tradeoff, not an oversight. The code carries `# TODO: ideally we'd use data_filter here, but it's busted: https://github.com/python/cpython/issues/107845`. The ask is to recheck whether that upstream bug is still open on the Python versions comfy-cli now supports, not to fix a mistake.

**Evidence**

```
$ ruff check --select S202 --output-format concise comfy_cli
comfy_cli/utils.py:174:13: S202 Uses of `tarfile.extractall()`
comfy_cli/utils.py:195:13: S202 Uses of `tarfile.extractall()`

$ sed -n '173,175p;184,195p' comfy_cli/utils.py
        with tarfile.open(inPath) as tar:
            tar.extractall(filter=None)
...
        def _filter(tinfo: tarfile.TarInfo, _path: PathLike):
            ...
            # TODO: ideally we'd use data_filter here, but it's busted: cpython#107845
            # return tarfile.data_filter(tinfo, _path)
            return tinfo

$ grep -n 'extract_tarball' comfy_cli/standalone.py
128:        extract_tarball(inPath=fpath, outPath=rpath, show_progress=show_progress)
```

**Fix**

Recheck cpython#107845 against the minimum supported Python. If it is fixed, switch both paths to `data_filter`. If not, keep the progress callback for UI but have `_filter` delegate to `tarfile.data_filter(tinfo, _path)` for the safety decision — the progress bar and the filter are separable concerns, and the current code conflates them.

---

Found by `repo-audit` during repo improvement sweep 2026-08-17. Parent: #723

christian-byrne:
Closes #725

## The problem

`extract_tarball()` bypassed Python's tarball extraction filter on **both** of its paths:

- non-progress path: `tar.extractall(filter=None)`
- progress path: a custom `_filter` that updated the progress bar and then `return tinfo` unmodified

So a member named `../evil.txt` or `/etc/evil` was written wherever it pointed — the CVE-2007-4559 class. It is reachable from caller-supplied input: `StandalonePython.FromTarball(fpath)` takes an arbitrary path, and `FromDistro()` downloads from a configurable `asset_url_prefix`.

## The crux: is cpython#107845 still a problem?

The filter was disabled deliberately, not by accident. The code carried:

```python
# TODO: ideally we'd use data_filter here, but it's busted: https://github.com/python/cpython/issues/107845
```

That TODO is now stale. What the bug actually was:

`data_filter` resolved a **symlink's** target against the destination root rather than against the directory containing the link. `TarInfo.linkname` is relative to the link's own directory for symlinks but relative to the archive root for hardlinks, and PEP 706's original implementation used the hardlink rule for both. Result: `LinkOutsideDestinationError` raised on perfectly valid archives.

Two things follow, and both matter here:

1. **It was a false-rejection bug, never an escape.** The buggy target was `realpath(dest/linkname)`; the correct one is `realpath(dest/dirname(name)/linkname)`. `dirname(name)` is always inside `dest` (the member name itself is checked earlier), so the buggy computation starts strictly shallower and can only reject *more* than the correct one — never accept a link that actually escapes. Upstream carries no security label and no CVE; the NEWS entry says it "will no longer **reject some valid** tarballs".
2. **It is fixed on every version we support.** Fixed by [cpython#107846](https://github.com/python/cpython/pull/107846) and backported to every live branch, all released 2023-08-24:

| Branch | First fixed release |
| --- | --- |
| 3.10 | **3.10.13** |
| 3.11 | **3.11.5** |
| 3.12 | **3.12.0rc2** → all 3.12.x finals are clean |
| 3.13+ | clean from 3.13.0a1 |

The affected set was exactly the releases that *introduced* the filter: 3.8.17, 3.9.17, 3.10.12, 3.11.4, plus 3.12.0b1–rc1.

`requires-python = ">=3.10"`, and CI runs 3.10 (pytest/build) and 3.12 (mac/windows/GPU). Intersecting that with the affected set leaves exactly **3.10.12 and 3.11.4** — two patch releases superseded three years ago. And since `extractall(filter=...)` does not exist *at all* before 3.10.12, the existing code already could not run below that line, so that two-release window is the entire residual exposure. On those two the worst case is a loud `LinkOutsideDestinationError`, not a silent escape — it fails closed.

So: **use `data_filter` directly, no version gate.** The `pypa/build`-style version gate is the wrong trade here, because its fallback branch reverts to *unfiltered* extraction on precisely those versions — reintroducing the hole this PR closes, to dodge a bug that only ever produced a loud error.

## The fix

The progress bar and the safety decision were conflated in one callback. They are separable concerns:

- non-progress path → `tar.extractall(filter="data")`
- progress path → `tar.extractall(members=_reporting_members(tar), filter="data")`, where the generator drives the two progress bars while yielding members

`members` drives the UI; `filter` stays the literal `"data"` filter. Using the string literal rather than `filter=tarfile.data_filter` also clears ruff's `S202` — `ruff check --select S202 comfy_cli` now passes clean across the package, which the callable form would not have.

## Tests

`TestExtractTarballFiltering` in `tests/comfy_cli/test_utils.py`, parametrised over both `show_progress` values so neither path can regress. The malicious tarball is built in-test; no binary fixture.

- `test_rejects_path_traversal_member` — a tarball carrying `../evil.txt` must not write outside the extraction directory, and must be rejected.
- `test_allows_internal_symlinks` — the control. python-build-standalone tarballs (the only thing `StandalonePython` extracts) are full of relative symlinks like `bin/python3 -> python3.12`. Those stay inside the destination and must still extract. This is the case cpython#107845 would have broken.

Verified the regression test genuinely fails against the unfixed code, on both paths:

```
FAILED tests/comfy_cli/test_utils.py::TestExtractTarballFiltering::test_rejects_path_traversal_member[False]
FAILED tests/comfy_cli/test_utils.py::TestExtractTarballFiltering::test_rejects_path_traversal_member[True]

E  AssertionError: traversal member escaped the extraction directory:
E    /tmp/pytest-of-c_byrne/pytest-76/test_rejects_path_traversal_me1/evil.txt
E  assert not True
```

With the fix, `tests/comfy_cli/test_utils.py` is 8 passed, and `tests/comfy_cli/test_standalone.py` (the caller) is 11 passed / 3 skipped.

## Notes for the reviewer

- **#511 also touches `comfy_cli/utils.py`**, but only to make `requests` a lazy import inside `download_url`. It does not touch `extract_tarball`, so these should not conflict.
- Scope note: `filter="data"` is not an absolute guarantee on an unpatched interpreter — there is a separate, later cluster of filter-bypass bugs ([cpython#135034](https://github.com/python/cpython/issues/135034), CVE-2025-4517 / 4330 / 4138 / 4435, fixed June 2025). That is an argument for keeping interpreters current, not against this change: filtered extraction is strictly better than the `filter=None` it replaces.

### DIFF
--- a/comfy_cli/utils.py
+++ b/comfy_cli/utils.py
@@ -169,31 +169,47 @@ def extract_tarball(
     shutil.rmtree(extractPath, ignore_errors=True)
     shutil.rmtree(outPath, ignore_errors=True)
 
+    # Both extraction paths below use the stdlib "data" filter so a member with an
+    # absolute path or a `../` traversal is rejected instead of being written
+    # wherever it points (CVE-2007-4559).
+    #
+    # This used to be skipped because of https://github.com/python/cpython/issues/107845,
+    # where data_filter resolved symlink targets against the destination root rather
+    # than against the directory holding the link, and so falsely raised
+    # LinkOutsideDestinationError on valid archives. That was a false-rejection bug,
+    # never an escape, and it was fixed in 3.10.13 / 3.11.5 / 3.12.0rc2 (2023-08-24).
+    # The only affected releases in our range are 3.10.12 and 3.11.4 — and since
+    # `extractall(filter=...)` does not exist before 3.10.12 at all, that is the whole
+    # window. On those two, the worst case is a loud error rather than a silent escape.
     if not show_progress:
         with tarfile.open(inPath) as tar:
-            tar.extractall(filter=None)
+            tar.extractall(filter="data")
         shutil.move(extractPath, outPath)
         return
 
     fileSize = inPath.stat().st_size
 
-    _size = 0
-
     with _tarball_progress("extracting tarball...", fileSize) as (barProg, barTask, pathProg, pathTask):
 
-        def _filter(tinfo: tarfile.TarInfo, _path: PathLike):
-            nonlocal _size
-            pathProg.update(pathTask, description=tinfo.path)
-            barProg.advance(barTask, _size)
-            _size = tinfo.size
+        def _reporting_members(tar: tarfile.TarFile):
+            """Yield every member, driving the progress bars as we go.
 
-            # TODO: ideally we'd use data_filter here, but it's busted: https://github.com/python/cpython/issues/107845
-            # return tarfile.data_filter(tinfo, _path)
-            return tinfo
+            Progress reporting used to ride on the ``filter`` argument, which
+            meant the extraction filter had to be a custom callable that
+            returned members unmodified — silently disabling the CVE-2007-4559
+            checks. The two concerns are separable: ``members`` drives the UI
+            and ``filter`` stays the stdlib ``"data"`` filter.
+            """
+            size = 0
+            for tinfo in tar:
+                pathProg.update(pathTask, description=tinfo.path)
+                barProg.advance(barTask, size)
+                size = tinfo.size
+                yield tinfo
+            barProg.advance(barTask, size)
 
         with tarfile.open(inPath) as tar:
-            tar.extractall(filter=_filter)
-        barProg.advance(barTask, _size)
+            tar.extractall(members=_reporting_members(tar), filter="data")
         pathProg.update(pathTask, description="")
 
     shutil.move(extractPath, outPath)
--- a/tests/comfy_cli/test_utils.py
+++ b/tests/comfy_cli/test_utils.py
@@ -1,4 +1,5 @@
 import io
+import tarfile
 from unittest.mock import MagicMock, patch
 
 import pytest
@@ -82,3 +83,79 @@ class TestTarballRoundTrip:
 
         assert (dest / "hello.txt").read_text() == "hello world"
         assert (dest / "sub" / "nested.txt").read_text() == "nested content"
+
+
+def _write_member(tar: tarfile.TarFile, name: str, data: bytes) -> None:
+    tinfo = tarfile.TarInfo(name)
+    tinfo.size = len(data)
+    tar.addfile(tinfo, io.BytesIO(data))
+
+
+def _write_symlink(tar: tarfile.TarFile, name: str, target: str) -> None:
+    tinfo = tarfile.TarInfo(name)
+    tinfo.type = tarfile.SYMTYPE
+    tinfo.linkname = target
+    tar.addfile(tinfo)
+
+
+class TestExtractTarballFiltering:
+    """Regression tests for CVE-2007-4559 (see issue #725).
+
+    ``extract_tarball`` extracts into the current working directory, so a member
+    named ``../evil.txt`` lands one level above it unless the stdlib ``data``
+    filter rejects the archive.
+    """
+
+    @staticmethod
+    def _traversal_tarball(workdir):
+        tarball = workdir / "payload.tgz"
+        with tarfile.open(tarball, "w:gz") as tar:
+            _write_member(tar, "payload/keep.txt", b"benign")
+            _write_member(tar, "../evil.txt", b"pwned")
+        return tarball
+
+    @pytest.mark.parametrize("show_progress", [False, True])
+    def test_rejects_path_traversal_member(self, tmp_path, monkeypatch, show_progress):
+        """Both extraction paths must refuse to write outside the destination."""
+        workdir = tmp_path / "work"
+        workdir.mkdir()
+        monkeypatch.chdir(workdir)
+
+        tarball = self._traversal_tarball(workdir)
+        escaped = tmp_path / "evil.txt"
+
+        rejection = None
+        with patch("comfy_cli.utils.Live"):
+            try:
+                extract_tarball(tarball, workdir / "out", show_progress=show_progress)
+            except tarfile.FilterError as exc:
+                rejection = exc
+
+        assert not escaped.exists(), f"traversal member escaped the extraction directory: {escaped}"
+        assert rejection is not None, "the traversal member was extracted instead of being rejected"
+
+    @pytest.mark.parametrize("show_progress", [False, True])
+    def test_allows_internal_symlinks(self, tmp_path, monkeypatch, show_progress):
+        """Control: the filter must not reject the layout real payloads use.
+
+        python-build-standalone tarballs (the only thing ``StandalonePython``
+        extracts) are full of relative symlinks such as ``bin/python3 ->
+        python3.12``. Those stay inside the destination and must survive.
+        """
+        workdir = tmp_path / "work"
+        workdir.mkdir()
+        monkeypatch.chdir(workdir)
+
+        tarball = workdir / "python.tgz"
+        with tarfile.open(tarball, "w:gz") as tar:
+            _write_member(tar, "python/bin/python3.12", b"#!/bin/sh\n")
+            _write_symlink(tar, "python/bin/python3", "python3.12")
+            _write_symlink(tar, "python/bin/python", "python3")
+
+        dest = workdir / "out"
+        with patch("comfy_cli.utils.Live"):
+            extract_tarball(tarball, dest, show_progress=show_progress)
+
+        assert (dest / "bin" / "python3.12").read_bytes() == b"#!/bin/sh\n"
+        assert (dest / "bin" / "python3").is_symlink()
+        assert (dest / "bin" / "python").is_symlink()


christian-byrne:
Verified this branch does what it says, and found one thing it does **not** cover — details and repro in https://github.com/Comfy-Org/comfy-cli/issues/725#issuecomment-5370560201.

**What this branch fixes, confirmed.** Ran all three filter forms against a tarball with a `../../ESCAPED.txt` member:

```
origin/main  filter=None            (utils.py:177)     -> extracted; escaped to /tmp/tar-.../ESCAPED.txt
origin/main  identity callable      (utils.py:187-195) -> extracted; escaped to /tmp/tar-.../ESCAPED.txt
c0fe495      filter="data"                             -> REFUSED: OutsideDestinationError
```

So both paths really were open and both are really closed. The commit message's read of cpython#107845 matches mine — false-rejection, never an escape, fixed in 3.10.13 / 3.11.5 / 3.12.0rc2.

**What it does not cover.** `utils.py:162-172` derives an `rmtree` target from the **first tar member's** name, before `extractall` runs:

```python
old_name = info.name.split("/")[0]
extractPath = inPath.with_name(old_name)     # with_name("..") is legal
shutil.rmtree(extractPath, ignore_errors=True)
```

A first member named `../evil` gives `old_name == ".."`, and the CLI recursively deletes the parent of the download directory. Same repro on both branches:

```
                    'python/bin/python3'   '../evil'
origin/main         workspace intact       workspace after: []
c0fe495 (this PR)   workspace intact       workspace after: []
```

The filter can't help here — the delete is header-derived and happens first.

Since this PR closes #725, merging as-is would retire the ticket with the delete primitive still live. Two options: extend this branch to also scope the extraction (`extractall(path=...)` into a temp dir, then move the top-level entry — note neither call passes `path=` today, so extraction currently lands in `os.getcwd()`), or explicitly de-scope it and drop the auto-close so #725 stays open. Either is fine, but it should be a decision rather than a side effect.

Also worth knowing while this is open, from the same review: the download host is not pinned (`standalone.py:82-88` takes both `tag` and `asset_url_prefix` out of a fetched `latest-release.json`), and the tarball is never checksummed even though `SHA256SUMS` is already downloaded two functions away (`standalone.py:30-59` parses it only for version numbers). Both are in the #725 comment.

christian-byrne:
Re-verified the above independently against `origin/main` and `pull/734/head` (`c0fe4956dd`). The
region is byte-identical on both — this PR adds `filter="data"` to the `extractall` calls, and the
`rmtree` happens before either of them:

```python
old_name = info.name.split("/")[0]
extractPath = inPath.with_name(old_name)
shutil.rmtree(extractPath, ignore_errors=True)   # <-- filter never gets a vote
```

One detail worth adding, because the obvious way to check this gives the wrong answer. On 3.12.3:

```
'..'       -> /home/u/dl/..
'.'        -> RAISES ValueError Invalid name '.'
'../evil'  -> RAISES ValueError Invalid name '../evil'
```

`with_name` rejects a traversal — so trying `with_name("../evil")`, seeing the `ValueError`, and
concluding the report is a false positive is a very easy mistake to make. `..` is the single string
that gets through, and `split("/")[0]` is precisely what turns the rejected form into the accepted
one. A member named `../evil` yields `old_name = ".."`, and `rmtree` then takes the parent of the
download directory.

Not asking for scope creep on this PR — just flagging that merging it as-is would auto-close #725
with the `rmtree` path still live, so the ticket would stop tracking the part that deletes files.
Either extend the fix to validate `old_name` (reject anything that isn't a plain single component),
or land this and leave #725 open against the remainder.

christian-byrne:
Following up on [the earlier comment](https://github.com/Comfy-Org/comfy-cli/pull/734#issuecomment-5370564455) about the `rmtree` half — adding only the thing it didn't have: a **runnable test**, and a measured red/green.

The two tests this PR adds both build their tarball with `payload/keep.txt` **first**:

```python
_write_member(tar, "payload/keep.txt", b"benign")
_write_member(tar, "../evil.txt", b"pwned")
```

So `tar.next()` returns `payload/keep.txt`, `old_name` is `"payload"`, and `extractPath` is benign. The tests exercise the *write* primitive that `filter="data"` closes, and never reach the *delete* primitive — which runs off the first header, before `extractall`, where the filter has no vote.

Making the malicious member **first** is the whole difference. Measured with the same harness on three variants:

| utils.py from | attack case | control (benign first member) |
|---|---|---|
| `origin/main` @ 3ff9f55 | **2 failed** | 2 passed |
| this PR's head @ c0fe495 | **2 failed** | 2 passed |
| this PR's head + the 2-line guard below | **2 passed** | 2 passed |

Failure message on both unfixed refs, for both `show_progress` values:

```
E       AssertionError: workspace settings were deleted by a tar header
```

The control passing on every row is what rules out a broken harness.

The guard that turns it green, at `comfy_cli/utils.py` immediately before `extractPath = inPath.with_name(old_name)`:

```python
if old_name in ("", ".", "..") or "/" in old_name or "\\" in old_name:
    raise ValueError(f"refusing to extract tarball with unsafe top-level member name: {old_name!r}")
```

Drop-in test, written against this PR's own file so it can be appended to `TestExtractTarballFiltering`:

```python
@pytest.mark.parametrize("show_progress", [False, True])
def test_first_member_name_cannot_select_an_rmtree_target(self, tmp_path, monkeypatch, show_progress):
    """The first tar header must not be able to pick what gets deleted.

    extractPath is derived from tar.next().name and rmtree'd BEFORE extraction,
    and Path.with_name("..") is legal, so a first member named "../evil"
    resolves the delete target to the parent of the download directory.
    """
    workspace = tmp_path / "workspace"
    downloads = workspace / "downloads"
    downloads.mkdir(parents=True)
    (workspace / "comfy.settings.json").write_text('{"real": "settings"}')

    tarball = downloads / "python.tgz"
    with tarfile.open(tarball, "w:gz") as tar:
        _write_member(tar, "../evil", b"x")          # FIRST — this is the point
        _write_member(tar, "payload/keep.txt", b"x")

    monkeypatch.chdir(downloads)
    with patch("comfy_cli.utils.Live"):
        try:
            extract_tarball(tarball, downloads / "out", show_progress=show_progress)
        except Exception:
            pass  # the delete already happened; how extraction ends is not the point

    assert (workspace / "comfy.settings.json").exists(), "workspace settings were deleted by a tar header"
```

Filing note: this came out of a closed-bug regression audit whose lens was "closed bugs whose regression test tests the wrong half". This PR is the cleanest live example of that shape, which is why it got a test rather than another issue. No objection to the PR's actual change — `filter="data"` is correct and measurably closes the write primitive; the only ask is that it not auto-close #725 while the delete primitive is still reachable.

comfy_cli/utils.py
```diff
@@ -169,31 +169,47 @@ def extract_tarball(
...
+    # than against the directory holding the link, and so falsely raised
+    # LinkOutsideDestinationError on valid archives. That was a false-rejection bug,
+    # never an escape, and it was fixed in 3.10.13 / 3.11.5 / 3.12.0rc2 (2023-08-24).
+    # The only affected releases in our range are 3.10.12 and 3.11.4 — and since
+    # `extractall(filter=...)` does not exist before 3.10.12 at all, that is the whole
+    # window. On those two, the worst case is a loud error rather than a silent escape.
     if not show_progress:
         with tarfile.open(inPath) as tar:
-            tar.extractall(filter=None)
+            tar.extractall(filter="data")
```
bigcat88, line 186: Not this PR's fault (pre-existing in both paths), but worth a follow-up: `extractall` is called without `path=`, so extraction lands in the CWD, while `extractPath` is computed next to the tarball (`inPath.with_name(...)`). When CWD != the tarball's parent directory, the subsequent `shutil.move` fails with `FileNotFoundError` — I hit this in practice while testing. Current callers happen to satisfy the assumption, but passing `path=inPath.parent` to `extractall` would make the function self-consistent. Happy to see that as a separate PR.

bigcat88 (approved):
Verified this end-to-end:

- Reproduced the vulnerability class: the new `test_rejects_path_traversal_member` fails against the unfixed `extract_tarball` (progress path) and passes with the fix.
- Ran a real-world extraction test with an actual python-build-standalone tarball (cpython-3.12.3 aarch64-apple-darwin `install_only`), both with and without progress: extraction succeeds, internal symlinks are preserved, and the extracted interpreter runs. So the `filter="data"` concerns from cpython#107845 don't bite on real PBS payloads.
- `ruff check`, targeted tests (`test_utils.py`, `test_standalone.py`), and full CI are green.

Nice touch driving progress via the `members=` generator instead of abusing the filter callback — that keeps reporting and security concerns separate.

PR #740, opened 2026-08-20
feat(workflow): add node mode overrides

### LINKED ISSUE #721: [Enhancement] workflow set-slot: support toggling node mode (bypass/mute)
## Motivation

`comfy workflow slots` / `set-slot` cover widget values, which makes value-level presets (resolution, steps, checkpoint name) fully scriptable against one canonical frontend-format file. But a very common preset dimension is **structural**: switching a node on/off — the canonical example being a LoRA loader that a "draft" preset enables and a "finals" preset disables.

The GUI answer is bypass (Ctrl+B, litegraph `mode` 4) and `workflow_to_api.py` already resolves bypassed/muted nodes correctly at conversion — the only missing piece is a CLI verb to set the mode.

Setting a LoRA's `strength_model=0` is not a substitute: ComfyUI registers zero-strength patches unconditionally (`model_patcher.add_patches` has no short-circuit), so under weight streaming you still pay the full per-step re-merge for a patch with zero effect. On an 11 GB card we measured that tax at ~20% of a whole render — a silent perf trap that mode toggling avoids.

## Proposal

Either of:

1. Extend the `set-slot` address space with a pseudo-input, e.g.
   `comfy workflow set-slot wf.json '105/200._mode=bypass'` (accepting `normal|mute|bypass`), or
2. A sibling verb: `comfy workflow set-mode wf.json 105/200=bypass`.

Subgraph-interior addressing (`instance/inner`) already exists in `slots`, so either form composes with it naturally. `slots` output could also list each node's current mode so agents can discover toggleable structure.

## Current workaround

We generate preset variants by running `set-slot --stdout` for values and then patching `node.mode = 4` for a declared `bypass:` list in a ~40-line script — it works, but it reaches into the file format where everything else goes through the CLI, and we'd happily delete it.

SiaoZeng:
## Summary

Add `comfy workflow set-mode` so scripts and agents can switch frontend workflow nodes between `normal`, `mute`, and `bypass` without patching workflow JSON directly.

Closes #721.

## What changed

- Add `comfy workflow set-mode <workflow> NODE=MODE...`.
- Accept top-level node IDs and nested `INSTANCE_ID/INNER_ID` paths.
- Preserve `--stdout` and JSON-envelope behavior used by `set-slot`.
- Validate every override before writing, so a failed batch leaves the source untouched.
- Reuse CQL's copy-on-write subgraph resolver so an instance-scoped edit does not mutate sibling instances.
- Fix CQL isolation for mixed UUID-typed and legacy name-typed subgraph instances by counting resolved definition identity and canonicalizing aliases before a fork.
- Register the command, error code, output schema, and bundled agent guidance.

## Why

Setting a LoRA strength to zero is not equivalent to bypassing the node: the patch can still be registered and re-merged during execution. Structural draft/final variants need a supported CLI operation for the LiteGraph node mode instead of ad-hoc JSON mutation.

## Testing

- TDD checkpoints: missing-command contract first, then implementation; review regressions were also committed RED before their fixes.
- `246 passed` across workflow, CQL, command-registration, and discovery suites.
- Targeted `ruff check` passed.
- Real CLI smoke: `workflow set-mode ... 6=bypass` emitted a successful `envelope/1` and wrote `mode: 4`.

## Repository-wide checks

- `ruff check .` passes.
- `ruff format --check .` is blocked by a pre-existing formatting difference in `docs/DESIGN-uv-compile.md`; this branch does not modify that file.
- The full local pytest run reaches the pre-existing failure `tests/comfy_cli/command/generate/test_spend_gate.py::test_interactive_prompt_accept_proceeds`; the same test fails on clean `origin/main` in this environment. Focused affected suites are green.

## Safety

- File writes use the existing atomic writer.
- Pretty-mode paths and addresses are sanitized before Rich rendering.
- Shared and deeply nested subgraph definitions are forked before mutation.
- Invalid modes, missing nodes, and late batch failures produce a structured error and no partial write.

### DIFF
--- a/comfy_cli/command/workflow.py
+++ b/comfy_cli/command/workflow.py
@@ -1,11 +1,11 @@
 """``comfy workflow`` — slot-based editing of ComfyUI frontend-format workflows.
 
-Three primitives:
+Four editing primitives:
 
     comfy workflow slots <file>                        # what can I tweak?
-    comfy workflow set-slot <file> ADDR=VALUE [...]    # tweak one or more
+    comfy workflow set-slot <file> ADDR=VALUE [...]    # tweak widget values
+    comfy workflow set-mode <file> NODE=MODE [...]     # normal/mute/bypass nodes
     comfy workflow vary <file> --slot ADDR='[v1,v2]'   # produce N variants
-
 Plus one read-only reader that needs no object_info at all:
 
     comfy workflow notes <file>                        # what did the author write?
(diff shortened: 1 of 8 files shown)


SiaoZeng:

PR #741, opened 2026-08-20
feat(nodes): add object-info snapshots

### LINKED ISSUE #739: Add a command to save object_info for offline validation

Could we add `comfy nodes snapshot --output object_info.json`?

It should use the normal target settings, stream the response, validate the JSON, and write the file atomically. This would give MCP clients a clean way to cache large node catalogs for offline validation.

**Describe alternatives you've considered**

Today users need to call `/object_info` with curl and manage the file manually.

**Additional context**

This gives comfy-mcp a CLI command it can wrap without adding direct HTTP logic.

SiaoZeng:
## Summary

Add `comfy nodes snapshot --output object_info.json` for repeatable offline node discovery and workflow validation.

Closes #739.

## What changed

- Resolve local/cloud targets through the normal routing chain.
- Stream `/object_info` into an exclusive temporary sibling and atomically replace the destination only after validation.
- Bound the response to 128 MiB and validate the top-level catalog incrementally, with an 8 MiB per-node-entry cap instead of materializing the complete catalog in memory.
- Preserve an existing destination and clean temporary files on invalid JSON, non-catalog JSON, size violations, network failures, and write failures.
- Reject credential-bearing plaintext HTTP targets in the shared authenticated request builder while preserving the existing unsupported-scheme error contract.
- Add structured `where_invalid` / `nodes_snapshot_failed` envelopes, output schema fields, discovery registration, and bundled skill guidance.

## Why

Large ComfyUI installations can exceed the existing 64 MiB buffered `/object_info` ceiling. `comfy validate --input` already supports saved catalogs, but users currently have to call the HTTP endpoint manually. The new command provides the CLI-owned, target-aware, atomic producer that `comfy-mcp` can wrap without adding a direct HTTP client.

## Testing

- TDD checkpoints cover the missing command, review-found security/resource boundaries, and the incremental entry bound.
- `187 passed` across snapshot, node CLI, shared HTTP, opener, command-registration, and discovery suites.
- Targeted `ruff check` passed.
- Live smoke against a local ComfyUI: streamed `1,805,169` bytes, validated `1,058` node classes, emitted a successful `envelope/1`, and removed the smoke artifact afterward.

## Repository-wide checks

- `ruff check .` passes.
- `ruff format --check .` is blocked by a pre-existing formatting difference in `docs/DESIGN-uv-compile.md`; this branch does not modify that file.
- The full local pytest run reaches the pre-existing failure `tests/comfy_cli/command/generate/test_spend_gate.py::test_interactive_prompt_accept_proceeds`; the same test fails on clean `origin/main` in this environment. Focused affected suites are green.

## Security and failure behavior

- OAuth/API-key headers are never sent to plaintext non-loopback HTTP.
- Redirects remain refused by the shared authenticated opener.
- Network bytes and the largest decoded catalog entry are independently bounded.
- No invalid or partial snapshot replaces a prior valid destination.

### DIFF
--- a/comfy_cli/command/nodes.py
+++ b/comfy_cli/command/nodes.py
@@ -20,14 +20,20 @@ Backed by the pure-Python CQL engine (``comfy_cli.cql.engine.Graph``).
 from __future__ import annotations
 
 import difflib
+import json
+import os
+import tempfile
+from pathlib import Path
 from typing import Annotated, Any
 
 import typer
 
 from comfy_cli import tracking
 from comfy_cli.cql.engine import Graph, LoadError
+from comfy_cli.http import authed_urlopen
 from comfy_cli.output import get_renderer, rprint
 from comfy_cli.output.sanitize import sanitize_markup
+from comfy_cli.target import Target, resolve_target
 
 app = typer.Typer(no_args_is_help=True, help="Introspect ComfyUI node classes (inputs, outputs, categories).")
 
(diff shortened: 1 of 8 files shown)


SiaoZeng:

SiaoZeng:
recheck

SiaoZeng:
Fixed the valid pretty-output markup finding in fe149d0. The snapshot destination now passes through `sanitize_markup`, with a regression test.

skishore23 (changes requested):
The core of this is solid — verified with a real loopback server that every failure path (invalid JSON, non-catalog, >8 MiB entry, >128 MiB body, HTTP 500, mid-body disconnect, connection refused, output-is-directory, missing parent, read-only dir) leaves the existing destination intact with zero `.tmp` files; the validator is genuinely incremental (52.8 MB / 55k-class catalog at 71 MB peak RSS); and the `http.py` plaintext gate regresses nothing (loopback `http://127.0.0.1` / `[::1]` / `localhost` still carry credentials, non-loopback http refused, `ftp://` keeps the unsupported-scheme error). Two things before merge:

**1. Rebase — it no longer merges.** `git merge-tree origin/main pr/741` conflicts in `comfy_cli/command/nodes.py` (add/add at the same insertion point as main's `widget-catalog` block from #789). Trivial to resolve by keeping both; with that done the full suite on the merged tree is 6172 passed / 1 failed, and the failure (`test_node_deps.py::test_non_pep440_installed_version…`) fails identically on plain main.

**2. (high) `nodes.py:1184` — local mode bypasses the `nodes` family's SSRF guard.** `engine._load_from_target` (`cql/engine.py:1804`) refuses `--where local` fetches to any non-loopback host — `nodes ls --where local --host 10.255.255.1` → `cql_no_graph: Refusing to fetch object_info from non-loopback host … (potential SSRF)`. `_stream_object_info_snapshot` calls `authed_urlopen(target.url("object_info"), target)` directly with no such check, so `nodes snapshot --where local --host 10.255.255.1 --output x.json` attempts the connection and writes whatever comes back to disk. The PR's own `test_snapshot_target_uses_normal_routing` (host `gpu-box`) enshrines the bypass. No credentials leak (local targets carry none), but this makes `snapshot` the one fetch-any-host-to-disk primitive its siblings deliberately refuse to be — especially relevant since the stated consumer is `comfy-mcp`. Apply the same `is_loopback_host` check and flip that test.

Non-blocking, in the same area:
- `nodes.py:1205-1212` collapses every failure into `nodes_snapshot_failed` with the hint "check the target, destination directory, and available disk space" — a cloud 401 tells an agent to check disk space, where `nodes ls` on the same target says `run comfy cloud login` with `details.status`. Map HTTP/auth failures the way the sibling commands do.
- The stream validator requires *every* entry to be a dict with `input`/`category`; the loader and engine tolerate a malformed entry (`any(...)` / `continue`), so one broken pack makes `snapshot` refuse a catalog `nodes ls` serves. Fail-closed, so low.
- The temp sibling leaks on SIGTERM (only `KeyboardInterrupt` unwinds the `finally`); output is written 0600 and replacing an existing 0644 file silently tightens it; `--output` at a symlink replaces the link rather than the target.

Also: `where_invalid` was already registered on main (`resolve_default_or_exit`), so the body's "add" overstates; only `nodes_snapshot_failed` is new.

PR #762, opened 2026-08-23
feat(knowledge): capability reachability — miss nudge and block vocabulary (BE-8810)

annehe9:
Linear: BE-8810 (parent BE-8296). Parts A and C of SPEC-capability-reachability. The comfy-knowledge side is Comfy-Org/comfy-knowledge#4, which is independent of this one.

**Draft, and stacked.** The base is `anne/be-8559-knowledge-enrichment`, the branch behind #756, not `main`.

The spec puts Part A inside #756 and Part C on a new branch off `main` once #756 merges. #756 is still open, and Part C rewrites the nudge that Part A introduces, so Part C cannot sit on `main` today. Both parts are here as separate commits so they can be split:

- `f9c5310` Part A, belongs in #756
- `bb870ca` Part C, belongs in its own PR once #756 lands
- `138d6ca` a review fix on Part A's behavior, belongs with `f9c5310`

## Part A — nudge on a query miss

`attach()` only built a nudge when the block ended up completely empty and the command's own result was empty. A browse that returns rows but misses the capability produced no signal at all.

`templates ls --tag "FLF2V"` matches 28 templates, so `thin` is False. Those 28 ids go through the reverse index and fill `models`. The agent got model rows and no indication that a ranked table for this job exists anywhere. A total miss was loud. A partial miss was silent, and the partial miss is the common one.

`attach()` now captures whether the query strings themselves resolved, before the reverse-index pass appends to `model_hits`, and nudges on a non-empty block whose query resolved to nothing.

`zero_hit` is unchanged. It still means an empty block on a thin result, which is what the query/miss log reads. Overloading it would change the log's meaning retroactively.

The narrowness is deliberate. `generate schema kling` resolves "kling" to a model, so it gets no nudge: the agent asked about a named thing and got it. `--tag FLF2V` resolves to neither a model nor a capability, so it does.

### Review fix on top

The spec's `query_resolved` rule counted only model-alias and capability lookups. A node class or template id passed as both a query and a reverse-index key therefore produced a model row plus a nudge denying that same term:

```
attach(queries=['KlingImage2VideoNode'], nodes=['KlingImage2VideoNode'])
  models: [('kling', matched_on='KlingImage2VideoNode')]
  nudge:  "no curated knowledge for 'KlingImage2VideoNode'; ..."
```

`nodes search <ClassName>` and `templates ls --name-sub <template id>` both take that path. A query now also counts as resolved when a shipped row's `matched_on` is one of the query strings. This is a hole in the spec's rule rather than a drift from it, and it contradicted the spec's own stated intent that a query reaching a named thing gets no nudge.

## Part C — capability vocabulary in the block

The agent was asked to hit a target it had never been shown. The 12 capability ids appeared nowhere it could read before forming a query, and Part A's nudge only helps after a miss, at the cost of a round trip.

Every enriched block now carries `capabilities_available`, read from the bundle. That matters: the set of capability ids is bundle content, which updates without a release, while comfy-cli code does not. Hardcoding the list into `SKILL.md` would put churning data on the slow release path. It is roughly 150 bytes for the current 12 ids, against a `MAX_BLOCK_BYTES` of 8192.

`_fit` counts it and evicts it last, after models and picks are exhausted. It is the only thing telling an agent which search terms reach a ranked table, so it outlives the answers to the current query.

Both nudges now point at it rather than repeating the id list, so the ids live in one place.

One addition beyond the spec: in the `zero_hit` branch, `capabilities_available` is popped if the block still overruns after the nudge degrades to its head. Without it a zero-hit block measured 251 bytes against a 220-byte ceiling once Part C added the key.

## Docs and schemas

`comfy_cli/schemas/knowledge_block.json` declares `capabilities_available`, and the `nudge` description no longer ties it to an empty result.

`comfy_cli/skills/comfy/SKILL.md` gains a rule that `capabilities_available` is the search vocabulary, and rule 2 now covers a nudge on a block that still carries rows. The ids themselves are deliberately not listed in `SKILL.md`.

## Verification

```
uv run --extra dev pytest tests/comfy_cli/test_knowledge_attach.py \
                          tests/comfy_cli/command/test_knowledge_enrichment.py -q
84 passed

uvx ruff@0.15.15 check .           All checks passed!
uvx ruff@0.15.15 format --check .  336 files already formatted
```

End to end against the live gallery, with `COMFY_KNOWLEDGE_FILE` pointed at a bundle built from Comfy-Org/comfy-knowledge#4:

```
FLF2V          hit_ids ['seedance', 'cap:first-last-frame']   picks 3   nudge None
Text to Audio  hit_ids ['cap:audio-generation']               picks 3   nudge None
Image Upscale  hit_ids ['magnific', 'cap:upscale']            picks 3   nudge None
Inpainting     hit_ids ['wan', 'cap:inpaint']                 picks 3   nudge None
Outpainting    hit_ids ['flux-1-fill', 'cap:inpaint']         picks 3   nudge None
Video Upscale  hit_ids ['topaz', 'seedvr2', 'cap:upscale']    picks 3   nudge None

ControlNet     hit_ids ['ltx', 'wan']   picks []   zero_hit False
               nudge "no curated knowledge for 'ControlNet'; see capabilities_available"
```

`capabilities_available` carried all 12 ids on every one of those, hits included.

New tests cover: an unresolved query on a non-empty block, a query resolving to a model, to a capability, and through the reverse index only, the nudge degrading to its head then disappearing under byte pressure, the vocabulary on every enriched block, the vocabulary matching the bundle's capability keys sorted, and `_fit` dropping the vocabulary only after models and picks.

The full comfy-cli suite has pre-existing failures on `main` from local `~/.config` and `FORCE_COLOR` leakage, and one hanging test in `test_run.py`. Both are unrelated to this branch.

https://claude.ai/code/session_012fFbkjpCcwnAFFGfMioG6u

### DIFF
--- a/comfy_cli/knowledge.py
+++ b/comfy_cli/knowledge.py
@@ -603,8 +603,9 @@ def attach(
     and ``nodes`` go through the reverse index. ``catalog_*`` are the ids the
     command actually loaded, used to drop rows and picks that do not resolve
     locally. ``thin`` marks a command whose own result was empty, which is the
-    only case that earns a nudge. Fail-open: any exception leaves ``payload``
-    exactly as it was.
+    only case that earns ``zero_hit``; a query that resolved to nothing earns a
+    nudge on its own. Fail-open: any exception leaves ``payload`` exactly as it
+    was.
     """
     try:
         bundle = load_bundle(cache_only=True)
@@ -612,6 +613,7 @@ def attach(
             return
         query_list = [q.strip()[:MAX_QUERY_CHARS] for q in queries if isinstance(q, str) and q.strip()]
         model_hits, cap_hits = _lookup(bundle, query_list)
+        query_resolved = bool(model_hits or cap_hits)
         seen = {mid for mid, _ in model_hits}
         for ids, index in ((templates, bundle.templates), (nodes, bundle.nodes)):
             for ident in ids:
@@ -655,6 +657,16 @@ def attach(
             block["nudge"] = f"{head}; covered capabilities: {', '.join(sorted(bundle.capabilities))}"
             if _block_bytes(block) > MAX_BLOCK_BYTES:
                 block["nudge"] = head
+        elif query_list and not query_resolved:
+            # A browse that returns rows the reverse index enriched, while the
+            # query itself matched neither a model nor a capability. Not a
+            # zero_hit: that feeds the miss log and means an empty block.
+            head = f"no curated knowledge for {query_list[0]!r}"
+            block["nudge"] = f"{head}; covered capabilities: {', '.join(sorted(bundle.capabilities))}"
+            if _block_bytes(block) > MAX_BLOCK_BYTES:
+                block["nudge"] = head
+            if _block_bytes(block) > MAX_BLOCK_BYTES:
+                del block["nudge"]
         payload["knowledge"] = block
     except Exception:  # noqa: BLE001 — knowledge is additive; the payload ships unchanged on any failure
         return
--- a/comfy_cli/knowledge.py
+++ b/comfy_cli/knowledge.py
@@ -646,6 +646,12 @@ def attach(
             entries.append(_model_entry(bundle, mid, row, matched_on=matched_on, brief=brief))
         entries.sort(key=lambda e: e.get("tier") != "law")
         entries = entries[: MAX_MODELS_BRIEF if brief else MAX_MODELS]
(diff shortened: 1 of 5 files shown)


annehe9:
Superseded. Part A landed in #756 directly, and Part C now lives in #763 as a Part C-only branch off the same base. Closing this to avoid double-applying the Part A commits.

PR #764, opened 2026-08-23
chore(hygiene): remove internal ticket IDs from comments and docstrings

annehe9:
## What

`public-repo-hygiene` was adopted in #758 and has failed on `main` ever since, with 252 findings. Every open PR inherits the red mark, so it says nothing about the PR showing it, and a genuine leak would sit unnoticed among 252 lines.

191 of those findings were internal tracker IDs cited in comments and docstrings. This removes all of them.

The prose around each one stays. What a comment explains about the code earns its place; the tracker number it came from does not, and nobody reading this repo can resolve that number anyway.

Every reference lived in a comment or a docstring, so nothing here changes behaviour. Verified by grepping for the ID shape outside comments before starting: no string literals, no identifiers, no test names.

Two supporting edits:

- A cql engine test wrote the sentinel `VALUE-FOR-10`, which matches the checker's ticket shape. Lowercased to `value-for-10` — it is arbitrary test data either way, written and then asserted back.
- `HAILUO-03` is MiniMax's Hailuo 03 model, named in the curated knowledge bundle and asserted in tests. It is a product name that happens to match the ticket shape, so it goes in the workflow's `ticket_allowlist` input rather than being renamed.

Also carries the two `ERA001` fixes that make `ruff check` pass. #735 enabled the rule with two hits still in the tree, so `main` is red on lint as well. `# status (billing / plan / concurrency)` and `# set-widget (name-addressed)` both parse as call expressions; reworded the way #735 handled `# browse: types / categories`.

## Verification

Ran the real checker locally, loaded from the SHA this repo pins in `.github/workflows/public-repo-hygiene.yml`, so these numbers are the ones CI computes:

| | findings |
| --- | --- |
| before | 252 |
| after | 58 |

The 58 that remain are all `Comfy-Org/<repo>` references. None is a ticket ID. They are triaged and tracked separately, and split three ways: repos that are public and missing from the org-wide allowlist, Hugging Face model paths the checker reads as GitHub repos, and a small number of genuine references to non-public repos.

- Full suite: 5640 passed, 42 failed.
- Those 42 are pre-existing. I ran the same four files on a clean `origin/main` worktree and diffed the failure lists: **identical**, so this change introduces zero failures. They come from `FORCE_COLOR` and `~/.config` leaking into the local environment.
- `ruff check` and `ruff format --diff` clean at the pinned 0.15.15.

## Notes

The remaining `Comfy-Org/<repo>` findings are not fixable from this repo alone. The allowlist is deliberately not a caller input, so the public-repo half needs a change in the shared workflows repo.

https://claude.ai/code/session_01L86eLQSKkxRnc7NchcqSzc

### DIFF
--- a/tests/comfy_cli/command/test_transfer_download.py
+++ b/tests/comfy_cli/command/test_transfer_download.py
@@ -285,7 +285,7 @@ GLB_RECORD = {
 
 
 class TestSaveGlb3dDownload:
-    """Regression (BE-4417): a SaveGLB job emits its output under the "3d"
+    """Regression: a SaveGLB job emits its output under the "3d"
     key. Shape-based extraction must resolve it so `comfy download` saves the
     `.glb` instead of erroring `download_no_outputs`."""
 
@@ -964,7 +964,7 @@ class TestSSRFGuardIntact:
 
 class TestDownloadHelperExtraction:
     """Direct, guard-pinning unit tests for the per-URL helpers extracted from
-    ``execute_download`` (BE-3273): ``_download_one_url`` orchestrates naming +
+    ``execute_download``: ``_download_one_url`` orchestrates naming +
     the symlink-dest refusal + branch dispatch; ``_copy_local_one`` and
     ``_stream_http_one`` carry the two download branches. The behavior-preserving
     extraction must keep every guard, envelope code, and exit intact — and the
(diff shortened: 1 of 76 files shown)


tests/comfy_cli/command/test_transfer_download.py
```diff
@@ -1274,11 +1274,11 @@ def get(self, key):
...
-    query param (a compromised/malicious server). Unlike the `item` token, it
-    was never sanitized, so control/ANSI bytes reached the on-disk name and
-    the echoed path — a terminal-injection vector in human mode. `_sanitize_ext`
-    whitelists it to a safe charset while preserving the no-traversal guarantee
-    (BE-3326)."""
+     query param (a compromised/malicious server). Unlike the `item` token, it
+     was never sanitized, so control/ANSI bytes reached the on-disk name and
+     the echoed path — a terminal-injection vector in human mode. `_sanitize_ext`
+     whitelists it to a safe charset while preserving the no-traversal guarantee
+    ."""
```
bigcat88: Non-blocking: removing `(BE-3326)` left this docstring slightly mangled — these four lines picked up an extra leading space, and the closing line is now a lone `.`. Suggest rejoining as `…while preserving the no-traversal guarantee."""` with the original 4-space indent.

bigcat88 (approved):
Verified this end-to-end:

- Reproduced the findings numbers with the real checker, loaded from `Comfy-Org/github-workflows` at the pinned `workflows_ref` SHA: origin/main → **252**, this branch with `--ticket-allow HAILUO-03` → **58**. All 58 remaining are `Comfy-Org/<repo>` allowlist findings; a residual grep for the ticket-ID shape finds nothing.
- Mechanically scanned every changed non-comment line in the diff: all are docstring prose, the self-consistent `VALUE-FOR-10` → `value-for-10` sentinel rename (write + assert, only two occurrences), and the two ERA001 comment rewrites. No behavior changes.
- Confirmed origin/main fails `ruff check` with exactly the two ERA001 errors this fixes; the branch passes.
- Ran all 55 changed test files (2709 tests): 14 failures, byte-identical failure list on origin/main — all pre-existing local-env issues, zero introduced by this PR.

One cosmetic nit inline — non-blocking.

PR #767, opened 2026-08-24
chore(hygiene): drop org-qualified private repo paths from two docstrings

mattmillerai:
## Summary

Two docstrings carried a provenance note that named a private repo by its full `org/name` path. Neither path was doing any work: the unqualified names they describe already appear elsewhere in the same files, and an unqualified name is not what the hygiene check denies. The fact survives, the path does not.

- `comfy_cli/cql/engine.py`: the port-provenance line now names the Go package without the repo path.
- `comfy_cli/knowledge.py`: the bundle producer is now described by pipeline rather than by repo path.

Comments only, no behaviour change.

## Testing

`ruff format --check` and `ruff check` clean on both files. `tests/comfy_cli/cql/` passes, 337 tests.

## Note on the remaining findings

This clears 2 of the repo's 29 hygiene findings. Of the rest, 19 are public model download URLs that the checker was misreading as GitHub references, fixed upstream in Comfy-Org/github-workflows#217, and 8 are a genuine private-repo reference in `.github/workflows/refresh-cql-catalogs.yml` that needs a design decision rather than a scrub, since the workflow authenticates against that repo and the name is load-bearing. Two markdown link texts in a captured upstream test fixture also remain and are correctly denied.

### DIFF
--- a/comfy_cli/cql/engine.py
+++ b/comfy_cli/cql/engine.py
@@ -4,7 +4,7 @@ Parses ComfyUI's ``object_info.json``, builds an indexed compatibility graph,
 and exposes upstream/downstream, path-finding, validation, annotations,
 and widget-order resolution.
 
-Port of ``github.com/Comfy-Org/cql/nodegraph`` (Go).
+Port of the upstream Go ``nodegraph`` package.
 """
 
 from __future__ import annotations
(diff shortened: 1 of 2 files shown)


mattmillerai:
Superseded by #768, which collapses this with the private-source-coordinate move and the gallery-fixture exclusion so a single PR demonstrates a green hygiene check. The docstring commit is carried over unchanged.

PR #773, opened 2026-08-25
perf(cli): keep psutil and GitPython off the startup import path

skishore23:
## What

Move three module-level imports that sit on every `comfy` invocation's startup path into the functions that use them, and add a test that keeps them there.

- `comfy_cli/utils.py`: `psutil` → inside `kill_all()` and `is_running()`.
- `comfy_cli/hardware.py`: `psutil` → inside `_detect_ram_bytes()` (reached eagerly via `env_checker` from `cmdline`).
- `comfy_cli/command/install.py`: `git` (GitPython) → inside the one branch that opens a repo.
- `tests/comfy_cli/test_import_budget.py`: runs `python -X importtime -m comfy_cli --json --version` in a subprocess and asserts `psutil`, `git`, `gitdb` are not imported at startup. On the previous `main` this test fails: the baseline loads all three. (A module-count budget test was in an earlier revision and was dropped in review: 100-200 modules of headroom is no signal, and it fires on dependency bumps instead.)

`stop_port.py` also imports `psutil` at module level but is already imported lazily by `cmdline` (only on `comfy stop-port`), so it is untouched.

## Why

Agents (comfy-agent, the MCP server) pay the CLI's import graph on every tool call. Measured on `main` before this change: `comfy --json version` loads 883 modules in ~0.65 s wall; `python -c pass` is 0.03 s, so it is the import graph, not the interpreter. GitPython and `psutil` were the two largest third-party items in `-X importtime` (GitPython about five times the cost of psutil) and only two helpers and one install branch need them.

## Measured

Same machine, `python -m comfy_cli --json version`, three runs each:

| | modules | wall |
|---|---|---|
| before | 883 | 0.64–0.73 s |
| after | 809 | 0.57–0.60 s |

About 10%. The remaining floor is comfy-cli's own ~80 modules pulled in by `cmdline.py`'s 21 eager `add_typer` groups; a follow-up makes those groups lazy (import on first use), which is where the larger win is. This PR lands the safe part and the guard.

## Behaviour change

None for any command that runs. A missing `psutil`/`git` would now surface at first use instead of at startup; both are hard dependencies in `pyproject.toml`, so this cannot happen in a normal install.

## Tests

- `pytest` on the touched modules (utils, hardware, env_checker, install, github/pr, stop_port): 252 passed.
- New `test_import_budget.py`: passes; fails on the previous `main`.
- Full `pytest`: 54 failed / 5774 passed on this branch vs 54 failed / 5772 passed on a pristine `origin/main` checkout in the same environment; the 54 failing test IDs are identical on both (mostly `tests/comfy_cli/command/test_logs.py` and `test_host_port.py`: `CliRunner(mix_stderr=...)` with the installed click, `where_invalid` routing, spend-gate prompt capture). Pre-existing and unrelated to this change; the +2 passes are the new budget tests.

Linear: BE-9172.

### DIFF
--- a/comfy_cli/command/install.py
+++ b/comfy_cli/command/install.py
@@ -7,7 +7,6 @@ import sys
 from typing import TypedDict
 from urllib.parse import urlparse
 
-import git
 import requests
 import semver
 import typer
@@ -209,7 +208,10 @@ def execute(
             raise typer.Exit(code=1) from e
 
     elif not check_comfy_repo(repo_dir)[0]:
-        # Get actual remote URL for better error message
+        # Get actual remote URL for better error message. GitPython is imported
+        # here so `comfy` commands that never touch a repo do not pay for it.
+        import git
+
         try:
             repo = git.Repo(repo_dir)
             remote_urls = [r.url for r in repo.remotes]
--- a/comfy_cli/command/install.py
+++ b/comfy_cli/command/install.py
@@ -208,8 +208,7 @@ def execute(
             raise typer.Exit(code=1) from e
 
     elif not check_comfy_repo(repo_dir)[0]:
-        # Get actual remote URL for better error message. GitPython is imported
-        # here so `comfy` commands that never touch a repo do not pay for it.
+        # Get actual remote URL for better error message
         import git
 
         try:
--- a/comfy_cli/hardware.py
+++ b/comfy_cli/hardware.py
@@ -292,11 +292,11 @@ def _detect_gpu(system: str, machine: str, cpu: str | None) -> dict | None:
 
 
 def _detect_ram_bytes() -> int | None:
-    # Lazy: psutil costs more to import than the rest of this module put
-    # together, and only RAM detection needs it.
-    import psutil
-
     try:
+        # Lazy: psutil costs more to import than the rest of this module put
+        # together, and only RAM detection needs it.
+        import psutil
+
         return int(psutil.virtual_memory().total)
     except Exception:
         logger.debug("RAM detection failed", exc_info=True)
--- a/comfy_cli/hardware.py
+++ b/comfy_cli/hardware.py
@@ -18,8 +18,6 @@ import os
 import platform
 import subprocess
 
-import psutil
-
 from comfy_cli import _safe_exec, cuda_detect
 
 logger = logging.getLogger(__name__)
@@ -294,6 +292,10 @@ def _detect_gpu(system: str, machine: str, cpu: str | None) -> dict | None:
 
 
 def _detect_ram_bytes() -> int | None:
+    # Lazy: psutil costs more to import than the rest of this module put
+    # together, and only RAM detection needs it.
+    import psutil
+
     try:
         return int(psutil.virtual_memory().total)
     except Exception:
--- a/comfy_cli/utils.py
+++ b/comfy_cli/utils.py
@@ -10,7 +10,6 @@ from contextlib import contextmanager
 from pathlib import Path
 from typing import BinaryIO, cast
 
-import psutil
 from rich import progress
 from rich.live import Live
 from rich.table import Table
@@ -69,6 +68,10 @@ def get_not_user_set_default_workspace():
 
 
 def kill_all(pid):
+    # Imported here, not at module level: psutil is the single most expensive
+    # import on the CLI's startup path and only these two helpers need it.
+    import psutil
+
     try:
         parent = psutil.Process(pid)
         children = parent.children(recursive=True)
@@ -80,6 +83,8 @@ def kill_all(pid):
 
 
 def is_running(pid):
+    import psutil
+
     try:
         psutil.Process(pid)
         return True
--- a/comfy_cli/utils.py
+++ b/comfy_cli/utils.py
@@ -68,8 +68,7 @@ def get_not_user_set_default_workspace():
 
 
 def kill_all(pid):
-    # Imported here, not at module level: psutil is the single most expensive
-    # import on the CLI's startup path and only these two helpers need it.
+    # Imported here, not at module level: only kill_all/is_running need psutil.
     import psutil
 
     try:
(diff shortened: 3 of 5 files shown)


tests/comfy_cli/test_import_budget.py
```diff
@@ -0,0 +1,80 @@
...
+import subprocess
+import sys
+from collections import Counter
+
+# Libraries that must stay off the startup path. Add to this list when a
+# lazy-import fix lands so the fix is guarded.
+FORBIDDEN_AT_STARTUP = ("psutil", "git", "gitdb")
+
+# Loose ceiling; the eager command-group imports keep us near 800 today.
+MODULE_BUDGET = 900
```
annehe9: I'd drop this test and keep only `test_heavy_libraries_stay_lazy`.

The budget is 900 against 692 actual on my machine, and 809 on yours. That is 100-200 modules of headroom, so it will not fire on the regression it is meant to catch. It passed on `main` unchanged too, so today it caught nothing. What it will actually do is fail on a dependency bump or on someone's env picking up an extra `.pth`, which sends a person chasing a number that has nothing to do with our code.

The forbidden-import test is the part with real signal. Suggest deleting `test_startup_module_count_within_budget` along with the `_breakdown` helper, which only exists to make this test's failure message readable.

annehe9 (commented):
Does what it says, and I verified it locally rather than taking the numbers on trust. Nothing here blocks merge.

On my machine (repo `.venv`, PR merged onto current `main`):

| | modules | `psutil`/`git`/`gitdb` at startup |
|---|---|---|
| `main` @ 8b9c8b0 | 766 | all three loaded |
| this PR merged onto `main` | 692 | none |

- The new test fails on `main` (`imported at startup but should be lazy: ['git', 'gitdb', 'psutil']`) and passes here, so the guard is real.
- `147 passed` across hardware / env_checker / install / install_gpu_flags / stop_port / utils.
- Passes with `--cov=comfy_cli` on, so the pytest-cov env stripping in b616413 does its job. That mattered, since `pytest.yml` runs with `--cov`.
- Pinned `ruff==0.15.15` on the four changed files: `All checks passed!` and `4 files already formatted`.
- Merges clean with `main` (the branch is 3 commits behind); module count and tests are unchanged after the merge.
- The `elif not check_comfy_repo(...)` branch has no test in the repo, so I drove it by hand both ways (a real git repo with a foreign remote, and a plain directory). Both reach `typer.Exit 1` with no `NameError`, so the lazy `import git` is placed correctly.

One thing outside the diff: the PR description says the measured command is `comfy_cli --json version`, but the test runs `--json --version`. Those are different paths, and `version` became a deprecated alias in 6a5e9d7. Worth making the description match what the test actually measures.

Four inline notes below, only the first is worth acting on before merge.

comfy_cli/utils.py
```diff
@@ -69,6 +68,10 @@ def get_not_user_set_default_workspace():
 
 
 def kill_all(pid):
+    # Imported here, not at module level: psutil is the single most expensive
```
annehe9: This claim is not right. Measuring `-X importtime` on `main`, `psutil` is 518 us self / 6883 us cumulative, while `git` is 26429 us self / 37280 us cumulative. GitPython is about five times the cost, which matches how the PR description frames them as the two largest.

Either fix the number or drop the comment. `import psutil` sitting inside a function already tells a reader it is deliberate.

comfy_cli/hardware.py
```diff
@@ -294,6 +292,10 @@ def _detect_gpu(system: str, machine: str, cpu: str | None) -> dict | None:
 
 
 def _detect_ram_bytes() -> int | None:
+    # Lazy: psutil costs more to import than the rest of this module put
+    # together, and only RAM detection needs it.
+    import psutil
```
annehe9: Small thing: move this one line down, inside the `try`.

The module docstring says everything except RAM degrades to `None`, and the `except Exception` right below is what delivers that. With the import outside it, an `ImportError` escapes `_detect_ram_bytes` instead of returning `None`. That is not a regression against `main` (a missing `psutil` crashed at import there too), but the fix is free and it makes the function match the contract the docstring states.

comfy_cli/command/install.py
```diff
@@ -209,7 +208,10 @@ def execute(
             raise typer.Exit(code=1) from e
 
     elif not check_comfy_repo(repo_dir)[0]:
-        # Get actual remote URL for better error message
+        # Get actual remote URL for better error message. GitPython is imported
```
annehe9: The second half of this comment explains the change rather than the code. `import git` inside the branch already reads as deliberate. I'd keep the original `# Get actual remote URL for better error message` and drop the rest.

comfy_cli/hardware.py
```diff
@@ -294,6 +292,10 @@ def _detect_gpu(system: str, machine: str, cpu: str | None) -> dict | None:
...
+
```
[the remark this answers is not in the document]
skishore23: Fixed. Moved the psutil import inside the try block in _detect_ram_bytes so an import failure falls back to None instead of escaping detect_hardware. Added tests that force the import to fail via sys.modules and assert both _detect_ram_bytes and detect_hardware degrade cleanly. 077eacc2.

tests/comfy_cli/test_import_budget.py
skishore23: agreed, dropped in 64ae07e9.

comfy_cli/utils.py
skishore23: fixed the claim in 64ae07e9.

comfy_cli/hardware.py
skishore23: already in 077eacc, the import now sits inside the try. No code change needed.

comfy_cli/command/install.py
skishore23: trimmed in 64ae07e9.

annehe9 (approved):

PR #774, opened 2026-08-25
perf(cli): import subcommand groups and heavy commands on first use

skishore23:
Stacked on #773 (base branch `skishore23/cli-lazy-imports`). Second half of BE-9172.

## What

Cut the CLI's startup import graph from 809 modules / 0.58 s (after #773) to 456 modules / 0.24 s, measured on `comfy --json --version`. `main` before #773 was 883 / 0.65 s. What is left is typer, rich and the stdlib: 29 of the 456 modules are comfy-cli's own.

Three commits, each standalone:

1. **`comfy_cli.command` resolves submodules lazily.** The package `__init__` imported every subcommand module eagerly (~200 ms, +340 modules on every invocation). It is now a PEP 562 `__getattr__`. The eager import had been added after one startup flake (`cannot import name 'transfer' from 'comfy_cli.command'`); the real cause was an import cycle that the eager, ordered import masked: `launch → custom_nodes.cm_cli_util → custom_nodes/__init__ → command → bisect_custom_nodes → launch`. Whoever imported `launch` first lost. `bisect_custom_nodes` now imports `launch` at the call site. `test_command_init.py` keeps both original checks and adds two: importing the package imports no submodule, and each submodule imports on its own in a fresh interpreter (the check that catches this class of cycle).
2. **`urllib.request` and `install.py` off the startup path.** `utils` imported `comfy_cli.http` for one timeout constant (`urllib.request`, ~60 ms). The two `--version` typer callbacks moved from `command/install.py` (which imports `requests` + the custom-nodes machinery) to `command/version_validators.py`; `install.py` re-exports them, so existing imports and tests are unchanged.
3. **Subcommand groups and heavy commands import on first use.** New `comfy_cli/_lazy.py`:
   - `LazyModule` / `lazy_attr`: proxies that import on first use and forward get/set/del to the real module. `cmdline.py` keeps its flat module-level names (`run_inner`, `custom_nodes`, `install_inner`, ...), so the ~70 `patch("comfy_cli.cmdline.run_inner.execute")`-style test patches work unchanged.
   - `LazyTyperGroup`: the root click group. The 22 `add_typer` groups and three heavy top-level commands (`generate`, `run-template`, `preview`) are declared in a table at the bottom of `cmdline.py` and built on first lookup with typer's own `get_group_from_info` / `get_command_from_info`, so help text, hidden flags and group callbacks (the `distribution` deprecation alias) behave exactly as before. `list_commands` keeps the eager commands first and the table order after. This also removes the ~75 ms typer spent building the click tree for all 48 nested groups on every run.

## Not done (from the ticket)

- Step 4 (lazy `WorkspaceManager()` / logging setup): measured 0.3 ms and 0.0 ms. Not worth code.
- `comfy_cli.tracking` (~25 ms, pulls `ui`/`workspace_manager`/`yaml`) stays eager: every command is decorated with `@tracking.track_command()`.

## Verification

- `comfy --help`, `comfy --help-json` (217 KB) and `comfy <group> --help` for node, nodes, cs, manager, model, models, cloud, jobs, build: byte-identical to the eager wiring (captured on the base commit, diffed after).
- New/updated tests: `test_lazy.py` (9: proxy forwarding incl. `mock.patch` and `monkeypatch.setattr` through a proxy; lazy group/command render identically to `add_typer`/`app.command`), `test_command_init.py` (19), `test_import_budget.py` (2, now on a real command and forbidding `requests`, `urllib.request`, `comfy_cli.http`, `comfy_cli.standalone`, and the heavy `comfy_cli.command.*` modules at startup).
- Targeted run of every test file that patches `comfy_cli.cmdline.*` plus help/discovery/distribution/generate: passes except the two `test_spend_gate` cases that fail identically on `origin/main` in this environment.
- Full `pytest`: see the last comment (same pre-existing failure set as `origin/main`).

## Measured

`python -m comfy_cli --json --version`, same machine, three runs each:

| | modules | wall |
|---|---|---|
| `main` (before #773) | 883 | 0.64–0.73 s |
| #773 | 809 | 0.57–0.60 s |
| this PR | 456 | 0.24–0.25 s |

Note: #773's numbers were taken on `comfy --json version`, which is the no-such-command error path (`version` is the `--version` flag). Startup cost is the same either way; the budget test now uses the real command.

Linear: BE-9172.


comfy_cli/_lazy.py
```diff
@@ -0,0 +1,214 @@
...
+        # Same TyperInfo ``add_typer`` would have appended: only the keywords
+        # that were given, so the sub-app's own settings fill the rest.
+        kwargs: dict[str, Any] = {"name": name}
+        if spec.help is not None:
+            kwargs["help"] = spec.help
+        if spec.hidden:
+            kwargs["hidden"] = True
+        if spec.callback is not None:
+            kwargs["callback"] = spec.callback
+        return typer.main.get_group_from_info(
```
annehe9, line 181: This call passes `pretty_exceptions_short` and `rich_markup_mode` but not `suggest_commands`, which typer 0.20 made required and defaultless. On the locked `typer==0.21.1` every lazy group raises `TypeError` here on first lookup.

`TyperGroup.__init__` already stores the value, so `suggest_commands=self.suggest_commands` is the fix.

One wrinkle: `pyproject.toml:57` still allows `typer>=0.12.5`, where the argument does not exist, so passing it unconditionally breaks the floor instead. Either raise the floor to `>=0.20`, or select the kwarg from `inspect.signature(typer.main.get_group_from_info)`.

comfy_cli/cmdline.py
```diff
@@ -2053,27 +1989,79 @@ def _warn_deprecated() -> None:
...
+    "manager": LazySubcommand("comfy_cli.command.custom_nodes", attr="manager_app", help="Manage ComfyUI-Manager."),
+    "pr-cache": LazySubcommand("comfy_cli.command.pr_command", help="Manage PR cache."),
+    "code-search": LazySubcommand("comfy_cli.command.code_search", help="Search code across ComfyUI repositories."),
+    "cs": LazySubcommand("comfy_cli.command.code_search", hidden=True),
+    "tracking": LazySubcommand("comfy_cli.tracking", help="Manage analytics tracking settings."),
+    "cloud": LazySubcommand("comfy_cli.cloud.command", help="Comfy Cloud — sign in, route commands, inspect session."),
+    "auth": LazySubcommand("comfy_cli.auth.command", help="Manage API tokens for model hosts (Civitai, Hugging Face)."),
+    "jobs": LazySubcommand("comfy_cli.command.jobs", help="List, inspect, and live-watch ComfyUI prompts."),
+    "build": LazySubcommand(
+        "comfy_cli.command.distribution", help="Package a local ComfyUI environment into a serverless build."
```
annehe9: `comfy_cli.command.distribution` was renamed to `comfy_cli.command.build` by `e4f661f` on `main`. This line and 2049 both still name the old path, and the file is gone.

Since these are lazy strings the failure is deferred to `comfy build` at runtime rather than caught at import. `git merge origin/main` also conflicts in this file at these exact hunks, so this needs a rebase either way.

annehe9 (commented):
The win here is real, and I verified it rather than trusting the numbers: 668 to 332 modules, with `comfy --help`, all 22 group helps, and the 217KB `--help-json` byte-identical to the base.

Two things block it though, and the first one is not visible unless you test on the pinned typer.

**1. It does not run on the typer version CI installs.**

```
uv.lock typer version: 0.21.1
signature: (group_info, *, pretty_exceptions_short, suggest_commands, rich_markup_mode)
suggest_commands: REQUIRED

$ comfy --help
TypeError: get_group_from_info() missing 1 required keyword-only argument: 'suggest_commands'
```

typer 0.20 made `suggest_commands` required with no default. `uv.lock` pins `0.21.1` and `.github/workflows/pytest.yml:32` runs `uv sync --extra dev --locked`, so CI gets exactly that. Every lazy group dies on first lookup, which is `comfy --help`, `node`, `build`, `cloud`, all of them. This reads like it was developed against 0.19.x, where the argument does not exist yet.

**2. Two lazy module paths point at a file that no longer exists.**

`e4f661f` on `main` renamed `comfy_cli/command/distribution.py` to `command/build.py`. Because these are lazy import *strings*, nothing fails at import time and no test notices. It surfaces as `ModuleNotFoundError` the first time someone runs `comfy build`.

Worth calling out as a general risk of this approach: moving an eager import to a string moves the failure from startup to first use, and out of the reach of the import-time checks. The new `test_each_submodule_imports_on_its_own` test is good and does what it claims, but it cannot catch a typo in a lazy path string. Some check that every registered lazy module path actually resolves would close that gap.

The rest is small stuff, inline.

comfy_cli/_lazy.py
```diff
@@ -0,0 +1,214 @@
...
+    def __repr__(self) -> str:
+        return f"<LazyModule {object.__getattribute__(self, '_lazy_path')!r}>"
+
+
+def lazy_module(path: str) -> Any:
+    """``run_inner = lazy_module("comfy_cli.command.run")``."""
+    return LazyModule(path)
+
+
+class _LazyAttr:
```
annehe9: I would drop `_LazyAttr` and `lazy_attr` entirely. It is 29 lines of descriptor machinery serving four call sites, and no test patches any of the attributes it defers (`launch_command`, `logs_command`, `normalize_cm_cli_exit_code`, `StandalonePython` all return zero hits), so the indirection is not buying testability either.

This same PR already solves the identical problem twice with a plain function-local import, in `bisect_custom_nodes.py:196` and `utils.py:114`. Four local imports would do it here and delete the class.

comfy_cli/_lazy.py
```diff
@@ -0,0 +1,214 @@
...
+    """One deferred ``app.command(name, ...)(fn)``. ``attr`` names the function
+    in ``module``. ``register`` instead names a ``register_with(app)`` hook on
+    the module that registers the command onto a Typer app itself (``generate``
+    does this because its command is a closure)."""
+
+    module: str
+    attr: str | None = None
+    register: str | None = None
+    help: str | None = None
+    context_settings: dict[str, Any] | None = None
```
annehe9: `context_settings` is never set by any caller. `grep -n context_settings comfy_cli/cmdline.py` comes back empty, and the value the test sees arrives through the `register_with` hook instead. Suggest deleting the field and the kwarg at `:191`.

Same idea for `lazy_module` at `:82` and `lazy_attr` at `:107`, whose entire bodies are `return LazyModule(path)` and `return _LazyAttr(path, name)`. Calling the classes directly reads the same and is two fewer names to learn.

comfy_cli/_lazy.py
```diff
@@ -0,0 +1,214 @@
...
+
+        class RootGroup(LazyTyperGroup):
+            lazy_subcommands = {"nodes": LazySubcommand("comfy_cli.command.nodes", help="...")}
+
+        app = typer.Typer(cls=RootGroup)
+    """
+
+    lazy_subcommands: dict[str, LazySubcommand | LazyCommand] = {}
+    # Mirrors ``typer.Typer()``'s default; the root app in cmdline.py uses it.
+    pretty_exceptions_short: bool = True
```
annehe9: Small one: this hardcodes typer's current default rather than reading it off the root app, so it quietly diverges the day someone writes `typer.Typer(pretty_exceptions_short=False)`. Same line as the `suggest_commands` fix, so probably worth doing in one pass.

comfy_cli/_lazy.py
skishore23: Fixed in cd27d62b without moving the floor. The group now builds the kwargs from inspect.signature of the factory: pretty_exceptions_short, rich_markup_mode and suggest_commands=getattr(self, "suggest_commands", True) are only passed when the installed typer accepts them. Added a test that asserts the kwarg set matches the factory signature, so a future typer that adds or drops one is caught.

Verified on both versions. typer 0.16.0 (miniconda): 214 passed for test_lazy.py, test_command_init.py, test_import_budget.py, test_help_json.py, test_command_mentions.py, test_build.py. typer 0.21.1 (uv run --locked --extra dev): same six files, 214 passed. Also confirmed the old code failed on the locked version: get_command(ctx, "nodes") raised TypeError: get_group_from_info() missing 1 required keyword-only argument: "suggest_commands", and now returns the group.

comfy_cli/cmdline.py
skishore23: Merged origin/main in ed551c71 and resolved cmdline.py there. Both entries now name comfy_cli.command.build, and the distribution alias still points at that same module with the deprecation callback. Nothing else in the tree referenced comfy_cli.command.distribution; command/__init__.py and test_command_init.py had no entry for it.

After the merge: test_lazy.py, test_command_init.py, test_import_budget.py, test_help_json.py, test_command_mentions.py and the renamed test_build.py all pass on typer 0.16 and on the locked 0.21.1. comfy --help and comfy --help-json are byte-identical to origin/main.

comfy_cli/_lazy.py
skishore23: Agreed, deleted in cd27d62b. _LazyAttr and lazy_attr are gone, and the four call sites are plain function-local imports inside the command bodies: launch, logs, normalize_cm_cli_exit_code in the update-all error path, and StandalonePython in standalone. Same shape as bisect_custom_nodes.py. The lazy_attr test in test_lazy.py is removed too.

comfy_cli/_lazy.py
skishore23: Both done in cd27d62b. LazyCommand.context_settings is gone along with the kwarg in _build_command, so CommandInfo is built with name, callback and help only. lazy_module and lazy_attr wrappers are deleted; cmdline.py now calls LazyModule(path) directly and the tests import LazyModule.

comfy_cli/_lazy.py
skishore23: Done in the same pass, cd27d62b. cmdline.py sets _RootGroup.pretty_exceptions_short = app.pretty_exceptions_short right after the app is created, so the value comes off the root Typer. The class attribute is now only a fallback and it reads typer Typer signature default instead of restating True, so it cannot drift either. Added a test asserting _RootGroup mirrors app.

skishore23:
Superseded by #789 (base branch deleted on #773's merge closed this one).

PR #780, opened 2026-08-25
fix(build): answer every builder failure with an envelope a create can trust (BE-9432)

james00012:
**TL;DR:** Every way a builder call can fail now produces an envelope, and a create that may already have written a build says so instead of inviting the retry that duplicates it.

Linear: [BE-9432](https://linear.app/comfyorg/issue/BE-9432/comfy-cli-a-builder-call-that-times-out-escapes-the-error-envelope-and)

**Why:** urllib wraps only the request phase in `URLError`, so a stalled read, a reset, a server hanging up mid-response and a truncated body all escaped as tracebacks: a human saw a stack, a `--json` caller saw nothing at all. The builder writes the build row before it answers, so on a create that silence costs the id of a build that already exists, and the retry it invites makes a second one.

**Validation**
* **Unit**: 156 pass in `test_distribution.py`. Nineteen of the new cases fail on `origin/main` (`e4f661f`) with this branch's test file copied in. `ruff@0.15.15 check` and `format --diff` both clean. Two whole-suite failures and the hygiene check are red here and equally red on a clean `origin/main` checkout, in files this change never opens.
* **Local stack, end to end**: the real CLI at a socket that accepts and stalls, both `from-snapshot` and `from-workflow`. On `origin/main` stdout is empty and stderr ends `TimeoutError: timed out`; here stdout carries one `build_builder_error` envelope with the `comfy build list` hint and `details.effectUnconfirmed`. Exit 1 on both.
* **Local stack, a build id the builder made up**: a response whose id is `dist-9\n\x1b[2KCreated build dist-0` prints a forged `Created build dist-0` line on the base. Here `from-workflow` refuses the id and says a build exists it cannot name. A list, a number and `dist-9; rm -rf ~` take the same path.
* **Local stack, the cases that must not change**: a blob PUT answering 412 before any create, and a create refused with 400, both keep the registered hint and no `effectUnconfirmed`. A cut answering 500 after the create now names the build and says to cut that one rather than retry. A builder answering 201 gives a byte-identical success envelope on both branches.
* **Not run**: the real builder behind `platformapi.comfy.org`, which needs a Cloud JWT and would spend a build slot. Every run above used a local HTTP double, so none of them proves the builder's own behaviour.

**What changed**
* **[comfy-cli] distribution.py, one `_BUILDER_TRANSPORT_ERRORS` tuple at five handlers**: every read-phase failure becomes an envelope, `update_cmd`'s resolve included, and an oversize answer goes through the same reporter instead of its own bare `renderer.error`.
* **[comfy-cli] the three create verbs**: carry a hint and `details.effectUnconfirmed`, released only where the create may really have landed. `execute_create` marks the create attempt, a status under 500 other than 408 is a refusal that wrote nothing, and a failure that already knows the id supersedes both and says to cut that build.
* **[comfy-cli] a build id from a response**: gets the grammar this file already gives a registry id, and is refused rather than printed. Seen and left: `create --execute`'s success path still prints its id unguarded ([BE-9472](https://linear.app/comfyorg/issue/BE-9472)), and a presigned upload URL still reaches the error message ([BE-9469](https://linear.app/comfyorg/issue/BE-9469)).

**Contradicts:** nothing. `details.effectUnconfirmed` is an additive optional field, which the envelope's own bump rule exempts (`comfy_cli/output/renderer.py:28-31`).

### DIFF
--- a/comfy_cli/command/build.py
+++ b/comfy_cli/command/build.py
@@ -29,11 +29,13 @@ from __future__ import annotations
 
 import contextlib
 import hashlib
+import http.client
 import io
 import json
 import os
 import re
 import subprocess
+import urllib.error
 from collections.abc import Iterator
 from pathlib import Path
 from typing import Annotated
@@ -46,6 +48,7 @@ from comfy_cli._safe_exec import BinaryNotFoundError, resolve_required_binary
 from comfy_cli.command.pack_scan import read_pyproject
 from comfy_cli.constants import DEFAULT_COMFY_MODEL_PATH, SUPPORTED_PT_EXTENSIONS
 from comfy_cli.file_utils import atomic_write_text
+from comfy_cli.http import ResponseTooLarge
 from comfy_cli.output import get_renderer
 from comfy_cli.registry.api import sanitize_error_body
 from comfy_cli.workspace_manager import WorkspaceManager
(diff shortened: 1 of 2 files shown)


james00012:
Closing this. The failure it fixes is rare, and the CLI is being revamped, so hardening the current error path is not worth carrying.

**One thing worth keeping when that work happens**, because it is not about timeouts. The CLI printed whatever arrived in the builder's id field. An answer containing `\n\x1b[2KCreated build dist-0` — a newline then an ANSI erase-line — renders as a line the user reads as the CLI's own output, and the id then lands unquoted in a shell command the CLI tells them to paste. The fix was a grammar check that refuses an id it cannot vouch for rather than printing it. Whatever replaces this code should not print a server-supplied id straight to a terminal.

The other half, that a create can time out after the builder has written the row and leave a build nobody can name, is the rarer case and is genuinely covered by `comfy build list`.

Branch left in place if any of it is wanted; the 21 tests describe the behaviours precisely.

PR #782, opened 2026-08-25
feat(workflow): add `comfy workflow print --format py`

skishore23:
Part 1 of BE-9191 (the agent-side `print_workflow` tool is a follow-up in the cloud repo after the pin bump).

## What

`comfy workflow print --format py`: render a save-format (UI) workflow as one Python-like line per node, with a binding-name → node-id map. Read-only; nothing parses the text back.

```
$ comfy workflow print wf.json --input object_info.json
checkpoint_loader_simple = CheckpointLoaderSimple(ckpt_name="v1-5-pruned-emaonly-fp16.safetensors")  # 4
empty_latent_image = EmptyLatentImage(width=512, height=512, batch_size=1)  # 5
clip_text_encode = CLIPTextEncode(clip=checkpoint_loader_simple.CLIP, text="beautiful scenery nature glass bottle landscape, purple galaxy bottle,")  # 6
clip_text_encode_2 = CLIPTextEncode(clip=checkpoint_loader_simple.CLIP, text="text, watermark")  # 7
ksampler = KSampler(model=checkpoint_loader_simple.MODEL, positive=clip_text_encode, negative=clip_text_encode_2, latent_image=empty_latent_image, seed=685468484323813, control_after_generate="randomize", steps=20, cfg=8, sampler_name="euler", scheduler="normal", denoise=1)  # 3
vae_decode = VAEDecode(samples=ksampler, vae=checkpoint_loader_simple.VAE)  # 8
save_image = SaveImage(images=vae_decode, filename_prefix="SD1.5")  # 9
# note 11 "Note: Model link": [Tutorial](https://docs.comfy.org/tutorials/basic/text-to-image) (+12 lines)
```

`--json` carries the same text as `source` plus `bindings` (`{"ksampler": "3", ...}`), `node_count`, `skipped` (notes and spliced UI-only nodes) and `warnings`; `--select` works like `slots`.

## Why

The agent cannot read a workflow today: it reasons about the graph through `ls-nodes` (rows, no edges) and N × `slots` calls, and edits against a topology it has not seen. One read that shows nodes, edges and every widget value in ~7 lines is the cheapest fix for that. Humans get a diffable artifact for free. Design: Obsidian `Comfy-Agent — workflow print (design, 2026-08-25)`.

## Rules (the ones that matter)

- One bound line per node, topological order, id inline. Class names that are not identifiers print as `Node["KSampler (Advanced)"]`; no mangling table.
- Widget names come from the catalog (`_expand_widget_entries`, so dynamic combos and `control_after_generate` behave as `slots` does); a slot absent from `widgets_values` prints the catalog default (the effective value), non-identifier names (`model.size_preset`) go in a trailing `**{...}`. A class the catalog does not know prints `widgets=[...]` positionally with a `# class not in catalog` note instead of refusing (the agent's catalog lags custom nodes).
- Edges: `binding` for single-output sources, `binding.NAME` otherwise, `binding.out[i]` when the output name is not an identifier. A widget-backed input with a live link prints the link, not the stale static value.
- Bypass/mute print as `mode=` annotations; links print as wired (printing preserves, compiling lowers), and a consumer of a bypassed node gets ` <input> via bypassed <id>` on its own line.
- `Reroute` and `GetNode`/`SetNode` are spliced through; `PrimitiveNode` is inlined with `# text from primitive 12`; notes print as one-line comments (full text is `workflow notes`).
- Subgraphs: the instance prints as `Subgraph["<title>"](...)`, each definition once as an indented block with `IN.<slot>` / `OUT.<slot> = ...` and fully qualified interior addresses (`10/3/<id>`, the form `slots` uses; the header also states the binding-key form). A missing definition prints opaquely with a warning (a real gallery template in the fixtures has one).
- Refuse and report, listing every reason (`workflow_print_unsupported`): legacy group nodes, duplicate node ids, links to missing nodes/slots or with non-integer slots, link cycles. Other malformed input degrades to a warning; nothing escapes as a traceback.

## Errors as navigation

The hints on `workflow_edit_invalid` / `workflow_slot_invalid` and the six edit-verb failure sites now lead with `comfy workflow print <file>` (they pointed at `workflow slots` / `ls-nodes`, which list rows without edges; the failure that dominates in prod is an edit against a node id that is not in the live graph). `slots` stays the pointer for exact `<node_id>.<input>` addresses; the misleading `nodes types` (connection types, not class names) is gone from the registry default. MCP inherits this; the agent's own steering is BE-9538.

## Files

- `comfy_cli/workflow_print.py` — the printer, pure (workflow dict + optional `Graph` → `PrintResult` or `PrintUnsupported`).
- `comfy_cli/command/workflow.py` — `print_cmd`, same shape as `notes`/`slots`; imports the printer lazily.
- Registrations: `discovery.py`, `schemas/workflow.json`, `error_codes.py`, `help_json.py`.
- Tests: `tests/comfy_cli/test_workflow_print.py` (printer: golden over the sd15 fixture, splicing, subgraph fixture, refusals), `tests/comfy_cli/command/test_workflow_print_cmd.py` (envelope, `--select`, pretty mode, schema validation, refusal envelope).

## Tests

- New tests: 57 (48 printer, incl. goldens over the sd15 and seedream V3 dynamic-combo fixtures, + 9 CLI). Contract suites (`test_discovery`, `test_error_code_registry`, `test_command_mentions`, `test_help_json`, `test_workflow_notes`, `test_workflow_slots`) plus the new files: 202 passed.
- Full `pytest`: 54 failed / 5862 passed; the 54 failing test IDs are identical to a pristine `origin/main` run in the same environment (env-only: `CliRunner(mix_stderr=...)` with the installed click, `where_invalid` routing, spend-gate prompt capture). All pass in CI on `main`.

Linear: BE-9191.

### DIFF
--- a/comfy_cli/command/workflow.py
+++ b/comfy_cli/command/workflow.py
@@ -609,6 +609,66 @@ def notes_cmd(
     renderer.emit(payload, command="workflow notes")
 
 
+@app.command(
+    "print", help="Render a workflow as one line of Python-like source per node, with a binding -> node id map."
+)
+@tracking.track_command("workflow")
+def print_cmd(
+    file: Annotated[str, typer.Argument(help="Frontend-format workflow JSON.")],
+    fmt: Annotated[str, typer.Option("--format", help="Output form. Only `py` today.")] = "py",
+    input_path: Annotated[
+        str | None, typer.Option("--input", help="Offline object_info.json dump for widget names.")
+    ] = None,
+    host: Annotated[str | None, typer.Option("--host")] = None,
+    port: Annotated[int | None, typer.Option("--port")] = None,
+    where: Annotated[str | None, typer.Option("--where", help="Routing for the catalog fetch: local or cloud.")] = None,
+    select: Annotated[
+        str | None, typer.Option("--select", help="Project the payload (see `comfy workflow slots --select`).")
+    ] = None,
+):
+    """Read-only. Reuses the catalog for widget names; refuses (with every reason) anything it cannot print faithfully."""
+    from comfy_cli.workflow_print import PrintUnsupported, render_py
+
+    renderer = get_renderer()
+    if fmt != "py":
+        renderer.error(
+            code="workflow_print_unsupported",
+            message=f"unknown --format {fmt!r}; only `py` is supported",
+            details={"reasons": [f"format {fmt!r}"]},
+        )
+        raise typer.Exit(code=1)
+    p, workflow = _load_workflow_or_fail(renderer, file)
+    graph = _get_graph(input_path, host, port, where=where)
+    try:
+        res = render_py(workflow, graph)
+    except PrintUnsupported as e:
+        renderer.error(
+            code="workflow_print_unsupported",
+            message="workflow contains something `print` cannot render faithfully: " + "; ".join(e.reasons),
+            details={"reasons": e.reasons, "path": str(p)},
+        )
+        raise typer.Exit(code=1) from e
+    payload = {
+        "workflow": str(p),
+        "format": "py",
+        "source": res.source,
+        "bindings": res.bindings,
+        "node_count": res.node_count,
+        "skipped": res.skipped,
+        "warnings": res.warnings,
+    }
+    if renderer.is_pretty():
+        typer.echo(_strip_terminal_controls(res.source), nl=False)
+        for w in res.warnings:
+            typer.echo(f"warning: {_strip_terminal_controls(w)}", err=True)
+    if select:
+        from comfy_cli.selector import emit_selected
+
+        emit_selected(renderer, payload, select, command="workflow print")
+        return
+    renderer.emit(payload, command="workflow print")
+
+
 # ---------------------------------------------------------------------------
 # Saved workflows — list, get, save, delete.
 # ---------------------------------------------------------------------------
--- a/comfy_cli/command/workflow.py
+++ b/comfy_cli/command/workflow.py
@@ -622,9 +622,13 @@ def print_cmd(
     ] = None,
     host: Annotated[str | None, typer.Option("--host", show_default=False)] = None,
     port: Annotated[int | None, typer.Option("--port", show_default=False)] = None,
-    where: Annotated[str | None, typer.Option("--where", help="Routing for the catalog fetch: local or cloud.")] = None,
+    where: Annotated[
+        str | None,
+        typer.Option("--where", show_default=False, help="Routing for the catalog fetch: local or cloud."),
+    ] = None,
     select: Annotated[
-        str | None, typer.Option("--select", help="Project the payload (see `comfy workflow slots --select`).")
+        str | None,
+        typer.Option("--select", show_default=False, help="Project the payload (see `comfy workflow slots --select`)."),
     ] = None,
 ):
     """Read-only. Reuses the catalog for widget names; refuses (with every reason) anything it cannot print faithfully."""
@@ -639,7 +643,18 @@ def print_cmd(
         )
         raise typer.Exit(code=1)
     p, workflow = _load_workflow_or_fail(renderer, file)
-    graph = _get_graph(input_path, host, port, where=where)
+    # Mirrors slots_cmd: a stale-cache fallback is a fact about the rendered
+    # widget names, so it has to reach the caller. `print`'s `warnings` is a
+    # list of plain strings (slots' is a list of objects), so it is reported
+    # as one, appended to whatever the printer itself warned about.
+    _stale: dict = {}
+    graph = _get_graph(
+        input_path,
+        host,
+        port,
+        on_stale=lambda key, err: _stale.update(source=key, reason=err),
+        where=where,
+    )
     try:
         res = render_py(workflow, graph)
     except PrintUnsupported as e:
@@ -649,6 +664,9 @@ def print_cmd(
             details={"reasons": e.reasons, "path": str(p)},
         )
         raise typer.Exit(code=1) from e
+    warnings = list(res.warnings)
+    if _stale:
+        warnings.append(f"object_info_stale: {_stale['source']}: {_stale['reason']}")
     payload = {
         "workflow": str(p),
         "format": "py",
@@ -656,7 +674,7 @@ def print_cmd(
         "bindings": res.bindings,
         "node_count": res.node_count,
         "skipped": res.skipped,
-        "warnings": res.warnings,
+        "warnings": warnings,
     }
     if select is not None:
         from comfy_cli.selector import emit_selected
@@ -664,7 +682,7 @@ def print_cmd(
         return emit_selected(renderer, payload, select, command="workflow print")
     if renderer.is_pretty():
         typer.echo(_strip_terminal_controls(res.source), nl=False)
-        for w in res.warnings:
+        for w in warnings:
             typer.echo(f"warning: {_strip_terminal_controls(w)}", err=True)
     renderer.emit(payload, command="workflow print")
 
--- a/comfy_cli/command/workflow.py
+++ b/comfy_cli/command/workflow.py
@@ -679,7 +679,8 @@ def print_cmd(
     if select is not None:
         from comfy_cli.selector import emit_selected
 
-        return emit_selected(renderer, payload, select, command="workflow print")
+        select_payload = _sanitize_selected_strings(payload) if renderer.is_pretty() else payload
+        return emit_selected(renderer, select_payload, select, command="workflow print")
     if renderer.is_pretty():
         typer.echo(_strip_terminal_controls(res.source), nl=False)
         for w in warnings:
@@ -804,6 +805,24 @@ def _strip_terminal_controls(text: str) -> str:
         if (ch in "\t\n" or 0x20 <= ord(ch) < 0x7F)
         or (ord(ch) >= 0xA0 and unicodedata.category(ch) not in _SPOOFING_CATEGORIES)
     )
+
+
+def _sanitize_selected_strings(value: Any) -> Any:
+    """Recursively strip terminal control chars from string leaves of a ``--select`` payload.
+
+    Only needed for pretty mode: ``emit_selected`` there can write a selected
+    bare string (or a JSON-pretty-printed slice) straight to the terminal, so
+    an escape sequence buried in a node title or widget value would reach the
+    terminal unescaped. JSON mode's encoder already escapes control chars, so
+    this is a no-op there and is never called for it.
+    """
+    if isinstance(value, str):
+        return _strip_terminal_controls(value)
+    if isinstance(value, dict):
+        return {k: _sanitize_selected_strings(v) for k, v in value.items()}
+    if isinstance(value, list):
+        return [_sanitize_selected_strings(v) for v in value]
+    return value
 
 
 def _reject_unsafe_workflow_key(renderer, key: str) -> str:
--- a/comfy_cli/command/workflow.py
+++ b/comfy_cli/command/workflow.py
@@ -617,10 +617,11 @@ def print_cmd(
     file: Annotated[str, typer.Argument(help="Frontend-format workflow JSON.")],
     fmt: Annotated[str, typer.Option("--format", help="Output form. Only `py` today.")] = "py",
     input_path: Annotated[
-        str | None, typer.Option("--input", help="Offline object_info.json dump for widget names.")
+        str | None,
+        typer.Option("--input", show_default=False, help="Offline object_info.json dump for widget names."),
     ] = None,
-    host: Annotated[str | None, typer.Option("--host")] = None,
-    port: Annotated[int | None, typer.Option("--port")] = None,
+    host: Annotated[str | None, typer.Option("--host", show_default=False)] = None,
+    port: Annotated[int | None, typer.Option("--port", show_default=False)] = None,
     where: Annotated[str | None, typer.Option("--where", help="Routing for the catalog fetch: local or cloud.")] = None,
     select: Annotated[
         str | None, typer.Option("--select", help="Project the payload (see `comfy workflow slots --select`).")
@@ -657,15 +658,14 @@ def print_cmd(
         "skipped": res.skipped,
         "warnings": res.warnings,
     }
+    if select is not None:
+        from comfy_cli.selector import emit_selected
+
+        return emit_selected(renderer, payload, select, command="workflow print")
     if renderer.is_pretty():
         typer.echo(_strip_terminal_controls(res.source), nl=False)
         for w in res.warnings:
             typer.echo(f"warning: {_strip_terminal_controls(w)}", err=True)
-    if select:
-        from comfy_cli.selector import emit_selected
-
-        emit_selected(renderer, payload, select, command="workflow print")
-        return
     renderer.emit(payload, command="workflow print")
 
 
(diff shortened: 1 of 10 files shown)


comfy_cli/command/workflow.py
```diff
@@ -609,6 +609,84 @@ def notes_cmd(
...
+        "source": res.source,
+        "bindings": res.bindings,
+        "node_count": res.node_count,
+        "skipped": res.skipped,
+        "warnings": warnings,
+    }
+    if select is not None:
+        from comfy_cli.selector import emit_selected
+
+        return emit_selected(renderer, payload, select, command="workflow print")
```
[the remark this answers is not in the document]
skishore23: kept typer.echo for the source path on purpose: Rich console markup would interpret the literal `[...]` in things like `Node["KSampler (Advanced)"]` and in user strings as rich markup, and notes_cmd/slots_cmd already use the same no-op Renderer.emit pattern in pretty mode, so switching this one command would be inconsistent.

fixed the real gap you flagged: `--select` in pretty mode now runs the selected payload through a small recursive sanitizer (`_sanitize_selected_strings` in workflow.py) before `emit_selected`, stripping terminal control chars from every string leaf. JSON mode is untouched since the encoder already escapes control chars. Added a test with a note containing `\x1b[31m` selected via `--select source` in pretty mode. a20f387e

comfy_cli/workflow_print.py
```diff
@@ -0,0 +1,1126 @@
...
+    out_name = None
+    if 0 <= slot < len(outputs):
+        out_name = outputs[slot].get("name")
+    if not out_name and graph is not None:
+        m = graph.node(src_type)
+        if m is not None and 0 <= slot < len(m.outputs):
+            out_name = m.outputs[slot].name
+    if out_name and _IDENT.match(out_name) and not keyword.iskeyword(out_name):
+        return f"{binding}.{out_name}"
+    return f"{binding}.out[{slot}]"
```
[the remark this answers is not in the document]
skishore23: fixed. `_edge_ref` now guards `isinstance(outputs[slot], dict)` before calling `.get("name")` on it — a non-dict entry falls back to the `.out[<slot>]` index form and appends a warning `node <id>: malformed outputs entry at slot <n>; referenced by index`. Checked every other `outputs[...]`/`inputs[...]` `.get(` site reachable from render_py — the rest already guard with isinstance before indexing (e.g. the Reroute source collector, _build_args, the subgraph proxy name maps), this was the only unguarded one. Added a test with `outputs: ["junk", "junk2"]` and a link on slot 0: no exception, `.out[0]` in the source, warning present. a20f387e

comfy_cli/workflow_print.py
```diff
@@ -0,0 +1,1126 @@
...
+    if extra:
+        inner = ", ".join(f"{json.dumps(k, ensure_ascii=False)}: {v}" for k, v in extra.items())
+        args.append(f"**{{{inner}}}")
+
+    widgets_values = node.get("widgets_values")
+    if isinstance(widgets_values, list) and widgets_values:
+        args.append(f"widgets={py_literal(widgets_values)}")
+
+    warnings.append(f"node {ctx.qualify(nid)}: subgraph definition {type_uuid} missing; printed opaquely")
+    line = f"{binding} = Subgraph[{json.dumps(type_uuid)}]({', '.join(args)})  # {nid} subgraph {type_uuid} definition missing"
```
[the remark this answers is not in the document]
skishore23: fixed, built the comment suffix separately as suggested: `call = f"{binding} = Subgraph[...]({...})"` then `line = f"{call}  # {nid} subgraph {type_uuid} definition missing"`. a20f387e

comfy_cli/error_codes.py
```diff
@@ -544,16 +544,23 @@ class ErrorCode:
...
         "A slot override failed validation (bad shape, unknown address, etc.).",
-        "see `details` — addresses follow `<instance_id>.<input_name>`",
+        "see `details` — addresses follow `<instance_id>.<input_name>`; `comfy workflow print <file>` shows every node with its id and widget names",
     ),
     ErrorCode(
         "workflow_edit_invalid",
         "A structured edit (add-node/connect/set-widget/delete-node) failed: "
         "unknown class_type, missing node, bad slot/widget name, or malformed address.",
-        "run `comfy workflow slots <file>` for widget addresses or `comfy nodes types` for class_types",
+        "run `comfy workflow print <file>` to see every node, edge and widget value with its id in one read (`comfy workflow slots <file>` for exact `<node_id>.<input>` addresses, `comfy nodes search` for class names)",
```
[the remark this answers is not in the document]
skishore23: fixed all 8 sites you listed — split each hint into adjacent string literals at the join point, text unchanged: error_codes.py workflow_slot_invalid (was line 557) and workflow_edit_invalid (was line 563), and the six hints in command/workflow_edit.py (_split_addr, add_node_cmd, set_widget_cmd, connect_cmd, delete_cmd, delete_nodes_cmd). ruff check/format clean. a20f387e

comfy_cli/workflow_print.py
```diff
@@ -0,0 +1,1142 @@
...
+
+    A subgraph instance whose definition is missing is deliberately NOT a
+    refusal here (at any depth): it's printed opaquely instead — see
+    ``_render_missing_subgraph_line`` and D11 in the task brief (amended).
+    """
+    reasons: list[str] = []
+    seen_ids: set[str] = set()
+    reported_dupes: set[str] = set()
+    for n in nodes:
+        t = n.get("type")
```
annehe9, line 151: This is the remaining crash. `_validate` calls `n.get("type")` on every entry of the top-level `nodes[]` with no `isinstance(n, dict)` check, and `_is_frontend_format` only verifies that `nodes` is a list, so a `null` entry sails through to here.

The fix already exists in this file. `_render_definition_block` filters to dicts and emits a warning. Doing the same in `render_py` covers this one, and it is the pattern the rest of the module follows.

comfy_cli/workflow_print.py
```diff
@@ -0,0 +1,1142 @@
...
+
+    return lines, len(printable)
+
+
+def render_py(workflow: dict, graph: Graph | None) -> PrintResult:
+    nodes = workflow.get("nodes") or []
+    links = workflow.get("links") or []
+    defs_by_id = {
+        sg.get("id"): sg
+        for sg in (workflow.get("definitions") or {}).get("subgraphs") or []
```
annehe9: Same class of input, different door. `(workflow.get("definitions") or {}).get("subgraphs")` assumes `definitions` is a dict. With `"definitions": [1]` the `or {}` does not fire, because a non-empty list is truthy, and this raises `AttributeError: 'list' object has no attribute 'get'`.

`_extract_notes` in `command/workflow.py` guards this exact field with `isinstance(..., dict)`. Worth matching.

comfy_cli/schemas/workflow.json
```diff
@@ -30,6 +30,27 @@
...
       }
     },
+    "format": { "type": "string", "enum": ["py"], "description": "print: output form. Only `py` today." },
+    "source": { "type": "string", "description": "print: the rendered Python-like source, one line per node." },
+    "bindings": {
+      "type": "object",
+      "additionalProperties": { "type": "string" },
+      "description": "print: binding name -> node id. Interior subgraph nodes use qualified keys/ids, e.g. \"10/3/primitive_int\" -> \"10/3/6\"."
+    },
+    "node_count": { "type": "integer", "minimum": 0, "description": "print: nodes that got a rendered line (top level + subgraph interiors)." },
```
annehe9: `{"type": "integer", "minimum": 0}` invalidates a payload that already ships. `comfy workflow get` emits `node_count: null` for a non-JSON body (`command/workflow.py:1099` and `:1438`), and `tests/comfy_cli/command/test_workflow_saved.py:307` asserts `env["data"]["node_count"] is None`.

Both commands resolve to this same schema (`discovery.py:98` for print, `:111` for get). Nothing caught it because `test_workflow_saved.py` does not schema-validate.

`"type": ["integer", "null"]` fixes it.

annehe9 (commented):
Went through this against head `a20f387`, and re-ran everything after your latest commit rather than reviewing the older push.

That commit fixed one of the three crash inputs I had found. A non-dict in a node's `outputs[]` now returns a clean envelope. The other two still reproduce, and both end the same way: exit 1, **empty stdout**, raw traceback. Under `--json` that means an agent gets no envelope at all, which is the thing the PR body says cannot happen.

```
$ echo '{"nodes":[null,{"id":1,"type":"KSampler","inputs":[],"outputs":[],"widgets_values":[]}],"links":[],"version":0.4}' > bad.json
$ comfy --json workflow print bad.json --input tests/comfy_cli/fixtures/sd15_object_info.json
EXIT=1
stdout: (empty)
stderr: AttributeError: 'NoneType' object has no attribute 'get'

$ comfy --json workflow notes bad.json      # same file, sibling command
{"schema": "envelope/1", ..., "ok": true, "command": "workflow notes", ...}
```

Same shape with `"definitions": [1]` instead, which gives `AttributeError: 'list' object has no attribute 'get'`.

Everything else checks out. Pinned `ruff==0.15.15` is clean, and the suite is `43 failed, 5873 passed` against `43 failed, 5816 passed` on `main` with identical failure names, so the delta is exactly your new tests. The module is not overbuilt for what it does. Details inline.

comfy_cli/error_codes.py
```diff
@@ -544,16 +544,25 @@ class ErrorCode:
         "got API-format. Auto-convert isn't wired yet.",
         "in ComfyUI, use the regular save (File > Save Workflow) — the API export is for `comfy run`, not for editing",
     ),
+    ErrorCode(
+        "workflow_print_unsupported",
```
annehe9, line 548: The description lists "missing subgraph definition" as a cause of this refusal, but `_render_missing_subgraph_line` deliberately prints those opaquely with a warning instead. `test_workflow_print.py:414` and `:437` pin that behaviour.

An agent reading the registry will hunt for a cause that never fires. Two causes that do refuse are unlisted: duplicate node ids, and non-integer link slots.

comfy_cli/workflow_print.py
```diff
@@ -0,0 +1,1142 @@
...
+        else:
+            value = positional[k] if k < len(positional) else None
+            text = py_literal(value)
+        _place_arg(iname, text, args, extra)
+
+    if extra:
+        inner = ", ".join(f"{json.dumps(k, ensure_ascii=False)}: {v}" for k, v in extra.items())
+        args.append(f"**{{{inner}}}")
+
+    line = f"{binding} = Subgraph[{bracket}]({', '.join(args)})  # {nid} subgraph {type_uuid}"
```
annehe9, line 751: A subgraph instance never picks up the `mode=` suffix that `_render_node_line` adds for regular nodes, here or at the sibling builder just below.

A node with `"mode": 4` and a subgraph type renders as `my_sub = Subgraph["MySub"]()  # 10 subgraph <uuid>`, with nothing saying it is bypassed. The PR body lists bypass/mute printing as `mode=` annotations, so this looks unintended.

comfy_cli/workflow_print.py
```diff
@@ -0,0 +1,1142 @@
...
+_UI_ONLY = frozenset({"Note", "MarkdownNote", "PrimitiveNode", "GetNode", "SetNode", "Reroute"})
+_NOTE_TYPES = frozenset({"Note", "MarkdownNote"})
+_IDENT = re.compile(r"^[A-Za-z_][A-Za-z0-9_]*$")
+
+# Sentinel ids litegraph uses for a subgraph definition's synthetic input/output
+# proxy nodes — matches workflow_to_api._SUBGRAPH_INPUT_NODE_ID / _OUTPUT_NODE_ID.
+_PROXY_IN = "-10"
+_PROXY_OUT = "-20"
+_MAX_SUBGRAPH_DEPTH = 10
+_MAX_SPLICE_HOPS = 100
```
annehe9: Minor, take it or leave it: neither of these two caps can fire usefully. The `seen` set already terminates every Reroute/GetNode chain, and `def_seen` already bounds the definition queue, so a cycle is caught before a counter runs out.

What is left is the bad case. A legitimate chain past the limit returns an unresolved id that then prints as `None` with no warning at all. Deleting both caps would remove that silent-wrong-output path.

comfy_cli/workflow_print.py
skishore23: Fixed. render_py now filters nodes[] to dicts before _validate runs, matching the pattern used elsewhere in the module. Non-dict entries add a warning: "workflow: ignoring N non-object node entries". Test added for a null entry. d42ed868.

comfy_cli/workflow_print.py
skishore23: Fixed. definitions is guarded with isinstance(..., dict) before .get("subgraphs"), same as _extract_notes in command/workflow.py. Non-dict definitions degrade to no definitions plus a warning: "workflow: ignoring non-object definitions block". Test added for definitions: [1]. d42ed868.

comfy_cli/schemas/workflow.json
skishore23: Fixed. node_count is now ["integer", "null"] in schemas/workflow.json, since comfy workflow get emits null for a non-JSON body on the same schema. Also added a schema-validation assert to the existing non-JSON-body test in test_workflow_saved.py so this cannot silently regress again. d42ed868.

comfy_cli/error_codes.py
skishore23: Fixed. Description now lists the causes that actually refuse: legacy group node, duplicate node id, link to a missing node/slot, non-integer link slot, link cycle, unknown --format. Dropped "missing subgraph definition" since that path prints opaquely with a warning, never refuses. Hint unchanged. d42ed868.

comfy_cli/workflow_print.py
skishore23: Fixed, both builders. The resolved-instance line and the missing-definition line now append the same mode=bypass/mode=mute suffix as _render_node_line, right after the id/subgraph comment and before any D9 annotation. Tests added with mode: 4 for both cases. d42ed868.

comfy_cli/workflow_print.py
skishore23: Deleted both. Checked each: _resolve_source loops on Reroute/GetNode chains and checks "key in seen" before every hop, so a cycle always returns splice_cycle before the counter matters. The definition queue is bounded by def_seen too: register_instance only appends a def_id to state.def_order the first time it is seen, so a self- or mutually-referential subgraph chain cannot regrow the queue, and the loop over state.def_order is iterative, not recursive, so there is no stack-depth risk either. Neither cap was catching anything a cycle could cause; both would only misfire on a legitimate long chain or deep nesting. _MAX_SPLICE_HOPS loop is now a plain while True, _MAX_SUBGRAPH_DEPTH check is removed with a comment explaining why. Existing cycle tests still pass. d42ed868.

comfy_cli/workflow_print.py
```diff
@@ -0,0 +1,1142 @@
...
+            warning = (
+                f"node {ctx.qualify(src_node.get('id'))}: malformed outputs entry at slot {slot}; referenced by index"
+            )
+    if not out_name and ctx.graph is not None:
+        m = ctx.graph.node(src_type)
+        if m is not None and 0 <= slot < len(m.outputs):
+            out_name = m.outputs[slot].name
+    if out_name and _IDENT.match(out_name) and not keyword.iskeyword(out_name):
+        return f"{binding}.{out_name}", warning
+    return f"{binding}.out[{slot}]", warning
```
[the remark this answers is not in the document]
skishore23: Fixed. _edge_ref now checks isinstance(out_name, str) before _IDENT.match, so a truthy non-string name like {"name": 1} falls back to the indexed .out[<slot>] form instead of raising TypeError. Test added for that case. d42ed868.

annehe9 (approved):

PR #784, opened 2026-08-25
feat(knowledge): make model choice a lookup the agent runs before building

annehe9:
## Why

An agent eval answered a first-last-frame request with Seedance 2.5 while the
bundle ranks `minimax-h3` first for that capability. Reading the Langfuse trace,
the knowledge was never in front of the model at the moment it chose:

- The turn opened with `templates ls --tag "Image to Video" --select rows.#.name`.
  `--select` returns before enrichment runs, by design, so that call carried no
  picks and the agent guessed two template names instead. Both guesses failed on
  an unrelated doc-host `LoadImage` error.
- The picks did arrive later, on `nodes search`, buried under a full Wan model
  card. One round later that tool result was elided from history, so `picks` was
  gone from context before the recommendation was written.
- Rule 1 in `SKILL.md` already said to read `knowledge.picks` before choosing. It
  was there the whole time and did not hold.

The pattern is that knowledge only ever rides along on a call the agent made for
another reason. A passenger can always be left behind. This PR starts making it
something the agent asks for directly.

## What changed

**`comfy knowledge pick` takes an optional argument.** The capability vocabulary
was previously only reachable by failing on purpose, since `pick` required its
argument and the full list came back on the error path as `details.known`. The
argument help said exactly that. A bare `comfy knowledge pick` now emits the
id-sorted list with descriptions.

**A miss is an answer, not an error.** A miss used to mean exit 1 and
`knowledge_unknown_capability`. That punishes the behaviour the skill now asks
for, because the agent is told to query the user's own phrasing first and prose
misses often. An agent that hits an error abandons the path, which is precisely
what the trace shows it doing after two failed `get_template` calls. A miss now
returns an ok envelope with `zero_hit`, the echoed `query` clipped to
`MAX_QUERY_CHARS`, a `nudge`, and the ids to retry with. That is the shape an
enrichment block already uses, so one concept keeps one shape. A hit carries
`zero_hit: false` so the field is always there to read.

`knowledge_unknown_capability` becomes unreachable and is dropped from the
registry, which the never-raised check requires.

**`SKILL.md` rules 1 and 4.** Rule 1 now routes model choice through
`knowledge pick` with the user's own words, and through `knowledge resolve` when
the user already named a model, since rank is a preference and an explicit
request outranks it. An explicit request is overridden only for correctness,
meaning `status: deprecated` plus `superseded_by`. The rule also says not to dodge
a miss by listing first, because the miss is what records a gap. Rule 4 names the
command that lists the ids.

## Deliberately not here

- No knowledge advisory on `validate`. Worth doing for correctness only, meaning
  deprecated and superseded, never rank. A rank advisory there would make the
  agent tear down a built graph over a preference.
- No capability table baked into `SKILL.md`. Measured at 74 to 638 tokens against
  an already ~17k-token file, so cost is not the blocker. Staleness is, since it
  would refresh only on `comfy skills install`.
- No change to list enrichment. Unfiltered listings already attach nothing via the
  `qualified` gate, and a filtered list is a query, so it should keep attaching.

## A note on measurement

`zero_hit` will under-report coverage gaps. Once the agent has seen the capability
ids it maps requests to the nearest one, so misses get absorbed rather than
reported. Treat `zero_hit` as an agent control signal. For curation data, the
unbiased intent is `agent.turn.input` in Langfuse, joined against the `pick` span
and the model actually built.

## Testing

`85 failed, 5778 passed` here against `85 failed, 5774 passed` on unmodified
`main`. Same 85 failures both sides, all pre-existing on this machine in jobs,
renderer, uv, host_port and onboarding. The +4 is the four new tests.
`ruff check` and `ruff format --check` are clean on every file touched.

https://claude.ai/code/session_01Rh6sB8qV69WuaBT3kdMRaJ


skishore23 (approved):
Approving. The code is right and small: bare `pick` lists capabilities, a miss is an `ok` envelope with `zero_hit: true`, echoed clipped `query`, a `nudge` and `capabilities[]` instead of exit-1, `pick()` gets the `_resolve_tokens` phrase fallback the enrichment path already had, `capability` joins the redaction set, and the dead error code is removed as the registry test requires. Ran with the CI-pinned ruff 0.15.15 (clean) and the two test files (200 passed). Five mutations (single `model` instead of `picks[]`, miss back to error, drop the fallback, `zero_hit` stuck false, un-redact `capability`) each fail a test. All 9 panel/CodeRabbit code comments addressed. The `hygiene` job fails on main too.

Two things to fix before merge, both cheap:

1. The title says a question the agent asks. What ships is a lookup the agent runs; SKILL.md rule 1 still has the agent take rank 1 itself, and nothing references `ask_user`. Either retitle to "make model choice a lookup the agent runs before building", or add one sentence to rule 1: when `picks` has more than one entry with differing caveat or route, show rank 1 and the alternatives to the user before building. I'd take the retitle here and do the behavior change with its own eval.
2. SKILL.md line 66 says a miss "records the gap". CLI-side it doesn't: `_emit_capabilities` emits the envelope and nothing else, `_log_query` is only called from the enrichment path (`knowledge.py:867`), and the only telemetry on the verb is `knowledge:pick_cmd` with a redacted capability, no hit/miss. Call `_log_query` on both paths of `pick` and `resolve` (consent-gated, best-effort). Also pick a policy: `_log_query` ships `queries` verbatim while this PR redacts `capability`; say which one search terms are.

Worth knowing: the Langfuse trace in the body is a cloud run, and comfy-cli's `SKILL.md` isn't in the cloud prefix (cloud embeds its own packs under `internal/loop/skills/` and rejects CLI shell syntax). This reaches Claude Code and comfy-mcp on the next PyPI release; cloud needs the wording ported into `building`.

Follow-ups: `knowledge.json` has no `required`, so `_validate` can't catch a missing `zero_hit`; a conditional `required` or per-verb schema fixes that. The capability listing is uncapped where everything else clips. And pre-existing, not this PR: the triad has no wheel snapshot, and manifest sha256 from the same URL is self-consistency, not authenticity.

PR #791, opened 2026-08-26
test: stabilize config isolation and unit-test memory use

guill:
**TL;DR:** Four test-harness fixes, no command behaviour changes: a `@singleton` that leaked config across tests, an OOM that killed the suite, a live pip bootstrap in a mocked install path, and an unrestored environment.

First of a stack that lands the `comfy-build.yaml` spec surface. It is independent of that work and mergeable on its own.

**What changed**

* **comfy_cli/utils.py**: `singleton` now records each decorated class's instance cache, and `reset_singleton_for_testing()` clears it. `ConfigManager` reads the config dir exactly once, at construction, so repointing the dir per test left the FIRST test's instance serving every later one. Any test that ran the CLI entrypoint then wrote `setup_nudged` into that shared instance, breaking the onboarding tests' "fresh config" premise hundreds of tests later. The cache lives beside the decorator rather than on the wrapper because several tests reach the decorated class through `Wrapper.__closure__[0]`.
* **tests/comfy_cli/conftest.py**: `_isolate_config_path` resets the `ConfigManager` singleton around each test, so the tmp-dir repoint actually takes effect.
* **tests/comfy_cli/command/test_run.py**: stub `validation_warnings` and `workflow_manifest` on the executor mock. `_emit_queued` hands those to `json.dumps` in NDJSON mode, and unstubbed `MagicMock`s made its `default` hook recurse on `isoformat`, retaining every child mock — ~28GB, and the OOM killer took the suite.
* **tests/comfy_cli/test_global_python_install.py**: patch `ensure_pip`. The non-fast-deps path bootstraps pip into the target interpreter for real, which fails wherever the stand-in path is not a real interpreter.
* **tests/comfy_cli/command/github/test_pr.py**: restore the ambient environment the test mutates.
* **pyrightconfig.json**: point the checker at `.venv` so it resolves installed packages.

**Validation**

* `uv run --locked --extra dev pytest -q`: 5970 passed, 38 skipped. The one failure (`test_build.py::test_scan_custom_nodes_records_repo_and_ref`) reproduces identically on a clean `origin/main` checkout in this environment — pre-existing, not this change.
* The OOM fix is the one claim the suite cannot show by passing: before it, a full run was killed by the OOM reaper on this machine; after, it completes in ~6 minutes.

**Contradicts:** nothing.

### DIFF
--- a/comfy_cli/utils.py
+++ b/comfy_cli/utils.py
@@ -8,7 +8,7 @@ import shutil
 import tarfile
 from contextlib import contextmanager
 from pathlib import Path
-from typing import BinaryIO, cast
+from typing import Any, BinaryIO, cast
 
 from rich import progress
 from rich.live import Live
@@ -16,6 +16,11 @@ from rich.table import Table
 
 from comfy_cli.constants import DEFAULT_COMFY_WORKSPACE, OS, PROC
 from comfy_cli.typing import PathLike
+
+#: Instance cache per ``@singleton`` class, keyed by the wrapper the decorator
+#: returns. Kept here rather than on that wrapper because several tests reach
+#: the decorated class through ``Wrapper.__closure__[0]``.
+_SINGLETON_CACHES: dict[Any, dict[type, Any]] = {}
 
 
 def singleton(cls):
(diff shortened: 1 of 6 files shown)


james00012 (approved):
Approving. Reads as: four test-harness fixes, no command behaviour changes.

PR #792, opened 2026-08-26
feat(cli): add shared require_option / confirm interaction helpers

guill:
**TL;DR:** Two helpers for the two refusals every command repeats — a required option that was not supplied, and a destructive action that needs consent — both failing closed in machine mode. No caller yet; the build surface adopts them next in the stack.

Stacked on #791. Review that one first; this PR's diff is the 3 files below.

**What changed**

* **comfy_cli/interaction.py** (new): `require_option()` and `confirm()`. Under a TTY they prompt; with no TTY, or for an agentic caller, they skip the prompt and emit the refusal envelope directly, so an agent never blocks on a prompt it cannot see. Both return through the caller's renderer, so the envelope carries the command's own labels.
* **comfy_cli/interaction.py**: the helpers deliberately hardcode **no** error code. Each caller names its own via `error_code=`. That keeps the register-with-first-call-site rule intact — a command's refusal code is registered by the change that introduces the refusal, not by this module, so nothing here pre-registers codes for commands that do not exist yet.
* **tests/comfy_cli/output/test_error_code_registry.py**: the registry scanner learns `error_code=` as a code kwarg beside `renderer.error`'s own `code=`. Without it, a command whose only refusal routes through these helpers would look like it registered an orphan and fail `test_every_registered_code_is_raised`.
* **tests/comfy_cli/test_interaction.py** (new): TTY and non-TTY paths, agentic-caller detection, `--yes` / explicit-option short circuits, and the envelope shape for each refusal.

**Why it lands with no caller:** the alternative is introducing it inside the build-surface change, where 8k lines of command rewrite would bury it. Landing the leaf first keeps this reviewable on its own terms, and the guardrail tests confirm it changes no existing behaviour.

**Validation**

* `uv run --locked --extra dev pytest -q`: 6008 passed, 38 skipped. The one failure (`test_build.py::test_scan_custom_nodes_records_repo_and_ref`) reproduces identically on a clean `origin/main` checkout — pre-existing, not this change.
* The registry guards (`test_every_registered_code_is_raised`, `test_every_raised_code_is_registered`) pass with the scanner change, confirming no code was silently orphaned or pre-registered.

**Contradicts:** nothing.

### DIFF
--- a/comfy_cli/interaction.py
+++ b/comfy_cli/interaction.py
@@ -1,0 +1,338 @@
+"""The three interaction tenets, in one place.
+
+Every ``comfy build`` command follows the same three rules (builder-cli-design
+lines 73-94):
+
+1. Required input missing **in an interactive context** → a good TUI prompt.
+2. Required input missing **in a non-interactive context** → fail with a
+   structured error naming the missing options, and print the command's help.
+3. All required input provided → run to completion with zero interaction, so
+   the command stays scriptable.
+
+``caller.detect_caller()`` already draws the line: ``.agentic is False`` is
+tenet 1, ``.agentic is True`` is tenet 2. Two escape hatches force tenet-3
+behaviour *even on a TTY* — ``-y`` / ``--yes`` accepts every confirmation, and
+the global ``--skip-prompt`` flag suppresses every prompt.
+
+**Output-channel rule.** Tenet 2 has two halves that land on *different*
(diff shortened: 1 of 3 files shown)


james00012 (approved):
Approving. Reads as: two refusal helpers that fail closed for a machine caller, with no caller yet.

PR #793, opened 2026-08-26
feat(build): add the comfy-build.yaml spec, path, target and diff primitives

guill:
**TL;DR:** The offline half of the `comfy-build.yaml` build spec — read/write, path resolution, target parsing, and structural diff — with no command wired to any of it yet. Three error codes land with the three call sites that raise them.

Stacked on #792 (which is stacked on #791). Review those first; this PR's diff is the 10 files below.

**What changed**

* **comfy_cli/command/build_spec.py** (new): read, write and schema-validate `comfy-build.yaml`. Serialization is deterministic — an unchanged spec round-trips byte-identically — so `status` and `pull` can diff a rewritten spec against the original without spurious churn. Raises `build_spec_invalid` and `build_spec_write_error`.
* **comfy_cli/command/build_paths.py** (new): resolve a `PATH` argument to a spec file — `<dir>/comfy-build.yaml` for a directory, the file itself for a `.yaml`/`.json` path. `BuildSpecNotFoundError` carries the exact absolute path probed in `details.path`, so an agent can retry against it without re-deriving the rules. Subclasses `ValueError`, not `FileNotFoundError`, so a broad `except OSError` around neighbouring file IO cannot swallow it. Raises `build_spec_not_found`.
* **comfy_cli/command/build_targets.py** (new): parse and normalize `os/gpu` build targets.
* **comfy_cli/command/build_diff.py** (new): structural diff between two specs — the engine `status` and `pull` report through later in the stack.
* **comfy_cli/error_codes.py**: registers exactly the three codes the modules above raise. Nothing is pre-registered for the commands still to come.
* **tests/comfy_cli/output/test_error_code_registry.py**: adds `test_every_build_code_has_a_first_call_site`. `comfy discover` publishes the registry verbatim, so a code with no call site advertises a branch to agents that nothing can ever take. This narrows the existing orphan check to the build family so a violation names the rule it broke.
* **tests** (4 new files): spec read/write and validation, serialization determinism, path resolution, and diff.

**One wording note:** two hints here are deliberately worded without naming `comfy build init` — that verb does not exist yet, and `test_no_mentions_of_nonexistent_commands` resolves every backticked command in prose against the live Typer tree. They are restored when `init` lands.

**Validation**

* `uv run --locked --extra dev pytest -q`: 6080 passed, 38 skipped. The one failure (`test_build.py::test_scan_custom_nodes_records_repo_and_ref`) reproduces identically on a clean `origin/main` checkout — pre-existing, not this change.
* Differential: the four new test modules fail on `origin/main` (the modules under test do not exist) and pass here.
* No command surface changes: `comfy build --help` is byte-identical to `origin/main`, and `comfy discover` gains only the three registered codes.

**Contradicts:** nothing.

### DIFF
--- a/comfy_cli/command/build_diff.py
+++ b/comfy_cli/command/build_diff.py
@@ -1,0 +1,321 @@
+"""Reconciling a fresh local scan against a build spec's stored ``definition``.
+
+``comfy build update`` rescans the installation and has to answer two questions
+from one pass over the same data: *what will change* — the diff it renders and
+confirms — and *what the new document is* — the ``definition`` it writes.
+
+**The rescan is merged, not substituted.** Everything the scanners produce comes
+from the scan; everything else is carried over from the stored entry. That
+distinction is load-bearing, not a nicety:
+
+- ``push`` caches ``blobId`` on an uploaded entry and may persist a resolver's
+  ``sourceUri``; ``pull`` merges server state and unknown keys back in
+  (plan decisions D-I). Substituting the definition wholesale would delete all
+  of it, so ``pull`` → ``update`` would churn the spec every single time.
+- Conversely the merge starts from the *scanned* entry, so a fact the scan no
+  longer reports — a pack that stopped being a git checkout, say — genuinely
+  disappears rather than lingering as stale metadata.
+
+**A content change invalidates the cache it backs.** ``blobId`` names bytes the
+builder already holds; once a model's ``sha256`` or a node's ``localDigest``
+moves, that blob is the wrong bytes and a persisted ``sourceUri`` is the wrong
+URL. Both are dropped here so the next ``push`` re-uploads instead of silently
+shipping what the user replaced (plan decisions D-I, "derived-field rules").
+
+Diff and merge live together because they share one notion of entry identity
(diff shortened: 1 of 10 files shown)


james00012 (approved):
Approving. Reads as: the offline half of the build spec, with nothing wired to it yet.

**Worth doing, not blocking**
* **`merge_definition` deletes `baseComfyVersion`, `pipDependencies` and `environment` when the rescan omits them** `comfy_cli/command/build_diff.py:233`. `build_definition` omits all three whenever `capture_pip_provenance` returns `None`, so a freeze timeout strips the spec rather than leaving it alone. Reproduced.
* **`.yml`, `.YAML` and `.JSON` resolve to `<file>/comfy-build.yaml`** `comfy_cli/command/build_paths.py:50`. `read_build_spec` lowercases the suffix and reads all three; `SPEC_SUFFIXES` does not, so the error names a path nobody typed (Modularity, one decision in one module).
* **`build_spec_invalid` advertises `details.path`** `comfy_cli/error_codes.py:1055`. Only `BuildSpecNotFoundError` carries `details`, and `comfy discover` publishes the promise on merge.

PR #796, opened 2026-08-26
fix(skills): carry the end-to-end findings into the bundled comfy-build

james00012:
**TL;DR:** Correct the bundled comfy-build skill against the CLI on `main`, recovering four fixes lost in the repository move and adding thirteen found by driving the file to real builds.

Linear: [BE-8971](https://linear.app/comfyorg/issue/BE-8971)

**Why:** the bundling commit copied the file while a pull request against the old repository was still open, so the most-tested content in it was dropped. Testing then found the file's output model was wrong for the agent that reads it: the CLI detects an agent, switches to JSON, and routes every advisory to stderr.

**Validation**
* **Local stack, end to end**: five staging builds reached `deployable: true` driving this content, across all three paths. The two that matter here ran against a CLI built from `origin/main`: scan (`f5d4cd28`) and workflow (`c62d1d22`), the latter carrying both models in its definition, which is the fix that the version-only recipe would have missed.
* **Local stack, end to end**: a sixth run correctly refused to cut a graph carrying an unpublishable class, which is the behaviour [BE-8426](https://linear.app/comfyorg/issue/BE-8426) asks for.
* **Not run**: Unit and Integration for the skill, which is prose with nothing to execute. This change touches no code, so the repository's suite is unaffected.

**What changed**
* **[comfy-cli] skills/comfy-build, output and environment**: says that an agent is auto-detected into JSON and that every advisory then lands on stderr, that `--builder-url` defaults to production, and that nothing in the CLI enforces the disclosure the file is built around.
* **[comfy-cli] skills/comfy-build, the workflow path**: `resolve` is in the command block, so following the commands rather than the prose no longer cuts a green build that cannot load its weights.
* **[comfy-cli] skills/comfy-build, corrections**: `release logs` answers with an object and picks a target, `release manifest` cannot confirm an allow-all seal, `comfy outdated` replaces a search per pack, `COMFY_BUILDER_TOKEN` makes `cloud whoami` a misleading pre-check, and agents branch on `error.code` rather than prose.

**Contradicts:** nothing. Two earlier drafts bent the file's own rules and were corrected rather than carried.

### DIFF
--- a/comfy_cli/skills/comfy-build/SKILL.md
+++ b/comfy_cli/skills/comfy-build/SKILL.md
@@ -11,7 +11,9 @@ The platform commands here are the `comfy build` group, from
 
 **Needs comfy-cli 1.18.0 or newer.** Earlier versions have no `build` group at
 all, so `comfy build scan` answers `No such command 'build'`. Check with
-`comfy --version`, and `pip install -U comfy-cli` if it is older.
+`comfy build --help`: a CLI without the group has no `build` command at all,
+and a CLI run from source reports a version no comparison can use. Upgrade with
+`pip install -U comfy-cli`.
 
 **A cut is not undoable and a build takes minutes**, so the user hears what is
 about to be sent, and agrees, before anything is created on the platform.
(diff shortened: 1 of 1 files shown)


guill (changes requested):
My CLI revamp is going to make this all inaccurate anyway and I really don't want to have to rebase again. IMO we should just close this PR and I'll do a big pass on the skill once my changes are in.

PR #800, opened 2026-08-27
fix(ci): raise the declared Python floor to 3.11

guill:
<details><summary><b>Stack of 6 — review bottom-up</b> (this is 800)</summary>

1. #800 — `fix(ci)`: raise the Python floor to 3.11 — base `main` &nbsp;&nbsp;**← this PR**
2. #801 — `feat(build)`: packaging, sync and Builder client primitives
3. #802 — `feat(build)!`: replace the legacy build and distribution surface **(breaking)**
4. #803 — `feat(deploy)`: the deployment control plane
5. #804 — `feat(deploy)`: the asset and job data-plane clients
6. #805 — `feat(deploy)`: `comfy deploy run`

</details>

**TL;DR:** `comfy_cli` has never been importable on Python 3.10 — `typing.assert_never` (`build_spec.py`) and `enum.StrEnum` are 3.11 additions imported at module scope — so the declared `>=3.10` floor and the 3.10 CI matrix have been describing a configuration that never worked. This raises the floor to 3.11 everywhere it is written down.

Bottom of a six-PR stack; nothing below it. Review from the bottom up.

**What changed**

* **pyproject.toml**: `requires-python`, the classifiers and ruff's `target-version` all move to 3.11.
* **workflows**: every job pinned to "follow the min version in pyproject.toml" moves with it, so the test job that runs on the floor is what keeps the floor honest.
* **lockfiles** (`uv.lock`, `pylock.toml`): re-resolved against the new floor.
* **3.10 compatibility shims**: retargeting ruff to py311 makes it see the shims the new floor obsoletes. Adopts `datetime.UTC`, and drops the `sys.version_info >= (3, 11)` guard around `-P` so `sys.path` is now hardened unconditionally.

**One deliberate omission:** UP042 stays off. Switching the nine `(str, Enum)` Typer choice types to `StrEnum` changes `str(OS.WINDOWS)` from `"OS.WINDOWS"` to `"windows"` — that is user-visible output, not a lint fix.

**Validation**

* `uv run --locked --extra dev pytest -q` at this commit: 6073 passed, 38 skipped, 1 failed.
* The one failure is `test_build.py::test_scan_custom_nodes_records_repo_and_ref`, and it is **pre-existing and environmental, not from this change**. It reproduces identically on a clean `origin/main` checkout. Cause: a local `url.git@github.com:.insteadOf https://github.com/` git rewrite on the dev machine — the fixture writes an `https://` remote and reads back the `git@` form, so the literal-URL assertion fails. CI runners carry no such rewrite.

**Contradicts:** nothing.


guill:
Closing this in favour of making the new code 3.10-compatible instead.

**The premise no longer holds.** This PR argued that `comfy_cli` could not import on 3.10 at all, so the declared `>=3.10` floor and the 3.10 CI matrix were describing a configuration that never worked. That was true at `f027859`, but #798 has since fixed it. I verified empirically against current `main` (`c1fa1f4`) on CPython 3.10.20:

* all 135 modules compile *and* import — zero failures
* no 3.11-only stdlib feature anywhere in `comfy_cli/` or `tests/`
* the full suite passes: **6096 passed, 38 skipped**

(The one failure in that run is `test_scan_custom_nodes_records_repo_and_ref`, caused by a local `url.git@github.com:.insteadOf https://github.com/` rewrite on my machine — it reproduces on a clean `main` checkout and is absent on CI runners.)

So `main` is genuinely 3.10-clean today. Raising the floor would drop a working configuration rather than stop a lie, and that is a policy decision rather than the bug fix this PR was written as.

**What happened instead.** The rest of the stack (#801–#805) has been rebased onto `c1fa1f4` and made 3.10-compatible in the commit that introduces each construct:

| Construct | Fix | Where |
|---|---|---|
| `typing.assert_never` | → `typing_extensions` (the pattern #798 established) | 8 files across all five PRs |
| `enum.StrEnum` | → `(str, Enum)`, the existing house pattern | `deploy_events.py` (#804) |
| `typing.NotRequired` + `TypedDict` | → `typing_extensions` (both, so key computation stays correct on 3.10) | `test_deploy_run.py` (#805) |
| `datetime.UTC` | → `timezone.utc` | `deploy_resolve.py` (#803) |

`typing-extensions>=4.7` was already a declared runtime dependency, so nothing new is added.

Every commit in the stack now passes the full suite on **both** 3.10 and 3.11, and `ruff check` / `ruff format --diff` are clean at CI's pinned 0.15.15 with `target-version = "py310"`.

The `stack/03b-python-floor` branch is left in place in case the team wants to revisit dropping 3.10 as a deliberate, standalone policy decision — that is a separate conversation from this stack, and #798 suggests the current preference is to keep 3.10 working.

PR #801, opened 2026-08-27
feat(build): add packaging, sync and Builder client primitives

guill:
<details><summary><b>Stack of 7 — review bottom-up</b> (this is 801)</summary>

1. #801 — `feat(build)`: packaging, sync and Builder client primitives — base `main` &nbsp;&nbsp;**← this PR**

</details>

**TL;DR:** The second offline build layer — packaging, spec-to-wire projection, and push/pull reconciliation — plus three `BuilderClient` corrections. Still no command wired to any of it.

Bottom of a five-PR stack, based directly on `main`.

**What changed**

* **build_package.py** (new): stage a custom node into an uploadable package.
* **build_validation.py** (new): project the authoring spec onto the builder wire format (D-I source precedence, authoring-only fields excluded) and validate it.
* **build_push.py / build_pull.py** (new): local-cache reconciliation and upload planning, and the identity-preserving merge that keeps local asset identities when a remote Build is fetched over the local spec.

`BuilderClient` changes in three ways:

* **Release vocabulary**: #770 moved the builder's own release endpoints server-side; the four client methods still spelled *version* now match — `cut_version` → `create_release`, `get_version` → `get_release`, and the same for the logs and manifest reads — with their call sites in `comfy_cli/command/build.py` following through.
* **Pagination**: `list_builds` and `list_releases` follow `nextCursor` to exhaustion. The builder pages both, and taking one page dropped the tail of any workspace past it, silently.
* **`update_build`** can now carry the server's existing `name` and `description`, which a local spec owns too and a sync had no way to send.

**Deliberately not here:** the non-empty targets guard on `create_release`. Today's `comfy build release create` takes no `--target` and relies on the server default, so refusing an empty list would break it. The guard arrives with `comfy build push`, which computes its own targets.

Adds no error codes — none of these modules raise one that is not already registered. Two module docstrings avoid naming `comfy build push` / `pull`, since those verbs do not exist yet and the help-string guard resolves every mention against the live Typer tree.

**Python 3.10 compatibility:** `build_validation.py` and `build_pull.py` import `assert_never` from `typing_extensions` rather than `typing`, matching the pattern #798 established in `build_spec.py`. `typing-extensions>=4.7` is already a declared runtime dependency. The declared floor stays `>=3.10`.

**Validation**

* CI (`build` job, Python 3.10) at this commit: **6185 passed, 38 skipped, 0 failed**.
* A local run shows one extra failure, `test_scan_custom_nodes_records_repo_and_ref`, which is **environmental, not from this change**: a `url.git@github.com:.insteadOf https://github.com/` rewrite makes the fixture read back a `git@` remote where it wrote an `https://` one. It reproduces on a clean `main` checkout, and CI confirms it does not occur on runners. The next PR up replaces that surface anyway.

**Contradicts:** nothing.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -38,8 +38,8 @@ history.
 
 - The builder client module is now `comfy_cli.builder_api` (was
   `comfy_cli.distribution_api`), and its methods say build and release
-  (`create_build`, `list_releases`, ...), matching the builder's public API.
-  The `distribution-definition/0` schema id is unchanged.
+  (`create_build`, `create_release`, `list_releases`, ...), matching the
+  builder's public API. The `distribution-definition/0` schema id is unchanged.
 - `comfy build --json` payloads carry the builder's vocabulary: `buildId`,
   `releaseId`, and `builds` and `releases` arrays. The retired `distributionId`,
   `versionId`, `distributions` and `versions` keys are emitted alongside them
@@ -60,6 +60,10 @@ history.
 - The shipped `build_from_snapshot.json` schema now requires the `build` key
   the builder actually serves; it still required the pre-rename `distribution`
   key, so a valid `comfy build from-snapshot --json` payload failed validation.
+- `comfy build list` and `comfy build release list` show every row again. The
+  builder pages both reads, and the client took only the first page, so a
+  workspace or a build past one page lost its tail — silently, with no error
+  and nothing in the output to say rows were missing.
 
 ## [1.16.0] - 2026-08-10
 
--- a/comfy_cli/builder_api.py
+++ b/comfy_cli/builder_api.py
@@ -22,7 +22,7 @@ from pathlib import Path
 import requests
 
 from comfy_cli import credentials
-from comfy_cli.builder_pagination import cursor_pages
+from comfy_cli.builder_pagination import PAGE_LIMIT, cursor_pages
 from comfy_cli.http import request_json
 from comfy_cli.target import Target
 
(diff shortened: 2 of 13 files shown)


skishore23 (approved):
Approving — the foundations hold up against the builder service (checked against the comfy-builder checkout of 2026-08-25: release routes, `nextCursor` on both lists, `UpdateBuildRequest` name/description, packaging traversal/symlink handling). Four non-blocking notes, all small:

1. **`build_package.py:188-199` — an archive destination inside the node root is packaged into itself.** `_package_files` collects members before `destination.open("wb+")` truncates it, so `_write_archive` reads the zip it is appending to and never hits EOF (reproduced: 2.9 GB written in 5 s). Not reachable from `push`, which uses a `TemporaryDirectory`, but `package_node` is the public API — exclude `destination` from members or refuse a destination under root. (This is CodeRabbit's open (a), materially worse than "minor".)
2. **`build_pull.py:184-192` — `_merge_entry` always lets local `blobId`/`sha256`/`sizeBytes` win and always drops the server's `sourceUri`, regardless of the local entry's `source`.** A server-side switch from private blob → public `sourceUri` is silently reverted on the next push; a local public entry pulled over a server blob loses the blob reference and hash. Reasonable for `source == "local"`, a silent clobber otherwise.
3. **`builder_pagination.py` / `builder_api.py:196,206` — the list walks send no `limit`, so `httpkit.ClampLimit` pages at 20, not the 100 `_MAX_LIST_PAGES=200` was sized for.** The walk fails loudly at 4,000 builds/releases and does 5× the round trips. `{"cursor": cursor, "limit": 100}` fixes it.
4. **`build_pull.py:238` rejects a Build with no `description`** — which is the server's default shape (`*string,omitempty`; `toBuild` only sets it when non-empty). #810 fixes this at the top of the stack, so it only matters if the stack doesn't merge together.

Also: the server has no blob dedup today (`CreateBlobUpload` always mints a fresh id; no `deduplicated` field anywhere in the Go/OpenAPI), so the `upload_url is None` branch and the docstrings describing it are dead-but-harmless.

Ran `test_build.py test_build_package.py test_builder_api.py` on the branch: 201 passed; ruff clean.

PR #804, opened 2026-08-27
feat(deploy): add the asset and job data-plane clients

guill:
<details><summary><b>Stack of 7 — review bottom-up</b> (this is 804)</summary>

4. #804 — `feat(deploy)`: the asset and job data-plane clients &nbsp;&nbsp;**← this PR**

</details>

**TL;DR:** The clients a workflow submission needs — asset dedupe/upload, job submit/poll/cancel, the event stream, and output download — with no command wired to them yet.

Stacked on #803. Review from the bottom up; this PR's diff is the 15 files below.

**What changed**

* **hashing.py** (new): blake3 content hashing. New `blake3` dependency, with the lockfiles bumped in the same change so no layer imports an undeclared package.
* **deploy_assets.py** (new): dedupe local assets by digest, upload what is missing.
* **deploy_jobs.py** (new): submit a job, poll it, cancel it.
* **deploy_events.py** (new): the job event stream.
* **deploy_download.py** (new): fetch completed outputs.
* **transfer.py**: `_stream_http_one` takes an injectable opener so the download path can reuse the shared streaming logic under its own HTTP-only opener.

**Error registry**: registers the 9 codes these clients raise. `deploy_endpoint_unknown` moves here from the control-plane group — its only call site is `deploy_download` — while `deploy_job_canceled` and `deploy_workflow_format_ui` stay out, because only `deploy run` and the workflow reader raise them.

A 422 `missing_asset` maps to `deploy_asset_missing` rather than falling through to the generic 422 row. The API lists it beside `invalid_workflow`, and without its own entry a run whose workflow is perfectly valid — but which names an asset id this account cannot mint — was reported as an invalid workflow, sending the reader off to edit a graph that has nothing wrong with it. `deploy_asset_missing`'s registry text widens to name both of its causes.

None of these modules emits a `command=` envelope label, so none depends on a Typer verb that does not exist yet.

**Python 3.10 compatibility:** `deploy_events.py`'s private `_WatchState` uses `(str, Enum)` — the existing house pattern — rather than the 3.11-only `enum.StrEnum`. It is compared only by identity in the `match` statement and never stringified, so the `str()` difference between the two does not apply. `assert_never` comes from `typing_extensions` here and in `test_deploy_assets.py`. The declared floor stays `>=3.10`.

**Validation**

* CI (`build` job, Python 3.10) at this commit: **6731 passed, 38 skipped, 0 failed**.

**Contradicts:** nothing.


skishore23 (approved):
Approving. This is the layer that had to be right, and it is — verified against the platform-gateway checkout (2026-08-25) and by adversarial testing on the branch: digest is `blake3:<64 hex>` matching the gateway's canonical regex; uploads stream in 1 MiB chunks with Content-Length equal to the streamed bytes, multipart field order the gateway requires, filename quotes/CR/LF neutralised, Bearer only to the endpoint; download origin checks are exact set membership (34 hostile URLs — subdomain suffixes, userinfo, `\\@`, `%2E`, punycode, ports, http, file:, private IPv6 — all rejected), the redirect handler refuses non-https/untrusted origins and strips auth on every hop, and the atomic part-file write, 10 GB cap, no-overwrite suffixing and symlink refusal are reused from `transfer.py`. The CodeQL hit is on the test's `in str(...)` assertion, not production. SSE parser is bounded (3 MiB lines, no terminator, bad UTF-8, deep nesting, comment floods) and the 30 s idle timeout is sound against the gateway's 15 s pings.

Three non-blocking notes:

1. **Connection-level failures escape the error mapping** — `deploy_jobs.py:164`, `deploy_assets.py:200,223,250`, `deploy_events.py:210,267` catch `(TimeoutError, URLError)`, but urllib only wraps `request()`; a socket closed before the status line raises `http.client.RemoteDisconnected` and a garbled one raises `BadStatusLine` (reproduced on loopback). On `submit_job` that's the LB-drain case the registry reserves for `deploy_job_submit_unknown` ("do not resubmit"), but #805's catch-all reports it as `deploy_server_error`, and `BadStatusLine` is caught by nothing. Same pattern in #802/#803 — one shared except tuple for `http.client.HTTPException`/`ConnectionError` would close all of them.
2. **`uv.lock` / `pylock.toml` carry cp311–cp314 blake3 wheels only, no cp310** (PyPI has them), while CI's matrix is `["3.10"]` and `requires-python >= 3.10`. `uv sync --locked` on 3.10 builds blake3 from the Rust sdist (confirmed with `--no-build-package blake3 --dry-run`); CI passed on a warm cache. `uv lock --upgrade-package blake3` + regenerate pylock.
3. `deploy_jobs._retry_after` accepts any non-negative `Retry-After` and `_sleep` is uncapped, while `deploy_events._retry_after` caps at 10 s — a `Retry-After: 86400` on a 429 parks the CLI for a day. Dormant today (gateway sends none on job 429s) but worth matching.

Minor: no cancel client lands here (it's in #805) despite the body's "submit/poll/cancel". Tests on the branch: 214 passed, 1 skipped; ruff clean.

PR #809, opened 2026-08-28
fix(cql): name every frontend widget slot in the widget order (BE-10283)

skishore23:
## Why

On every cloud env with the doc host (prod-v2 `v0.233.12`, testcloud, ephemerals) the CRDT lazy mint fails for any workflow that contains a Load Image / Load Audio / Load Video / 3D preview node:

```
createNodeMap(LoadImage): widgets_values has 2 entries but widget_order names only 1
```

and the turn silently falls back to the v0 draft path. The catalog this CLI publishes (`comfy nodes widget-catalog`, `Graph.widget_order_default`) only names schema-declared widget inputs plus `control_after_generate`, but the frontend serializes more slots. Linear: BE-10283.

## What the frontend actually writes (verified in `ComfyUI_frontend`)

| source | slot | position | value |
|---|---|---|---|
| `Comfy.UploadImage` / `Comfy.UploadAudio` (`beforeRegisterNodeDef` adds `required.upload`) | `upload` | after every declared input, optional ones included (`getOrderedInputSpecs` appends unlisted inputs last) | `"image"` on older frontends; `serialize: false` now |
| `Comfy.AudioWidget` (`required.audioUI`) on LoadAudio/SaveAudio*/PreviewAudio | `audioUI` | same, registered before `upload` | `null` / unserialized |
| `Comfy.Preview3D` / `Comfy.SaveGLB` (`required.image = ['PREVIEW_3D']`) | `image` | same | `""` |
| server-declared DOM widget inputs (`Load3D.image` is `LOAD_3D`, `LoadAudioUI.audioUI` is `AUDIO_UI`) | declared name | declared position | `""` / `null` |
| any input with `widgetType` (`FLOAT,INT` + `widgetType: "STRING"`) | declared name | declared position | the widget's value |

Captured shapes the tests pin: `LoadImage ["x.jpg","image"]`, `LoadImageMask ["m.png","red","image"]`, `LoadVideo ["x.mp4","image"]`, `LoadAudio ["x.wav",null,null]`, `SaveGLB ["mesh/ComfyUI",""]`, `Preview3D ["out/mesh.glb",""]`.

`@comfyorg/comfy-multi-player` tolerates fewer values than names, so naming every slot the frontend can write, in the position it writes it, fixes both old and new frontends.

## Change

- `engine.py`: `frontend_extra_widget_names()` (injection rules), `_FRONTEND_DOM_WIDGET_TYPES`, `PortOptions.upload_flags` / `widget_type`, `_is_link` honors DOM types and `widgetType`; wired into `widget_order`, `widget_order_default`, `_expand_widget_entries` (so `widget_order_for_node` / `set-widget` agree), `widget_defaults` (`""` for a non-trailing DOM slot, nothing for the trailing markers).
- `workflow_to_api.py`: `_is_widget_input` honors the same DOM types and `widgetType`, so the converter walk consumes `Load3D.image` in place and no longer reads `width` one slot early.
- `workflow_ops.py`: `set-widget`'s "available" listing hides the marker slots.
- Docstring / schema description / CHANGELOG.

## Real-catalog diff (`services/ingest/data/object_info.json`, 3698 classes)

60 classes change, all in the families above: the Load*/Save*/Preview* media and 3D families gain their trailing `upload` / `audioUI` / `image` or in-place `viewport_state`, `LTXVEmptyLatentAudio` gains `frame_rate`, and the "Basic data handling" math nodes gain their `widgetType`-declared value inputs (which the CLI was misreading as links, so `set-widget`/conversion on those nodes were already one slot off).

## Testing

- New `tests/comfy_cli/cql/test_frontend_widget_slots.py` (29 cases): every order surface for each family, the captured frontend shapes fit the order, converter walks for Load3D / LoadImageMask / Preview3D, `widgetType` on and off.
- Full suite: 6176 passed, 38 skipped.
- ruff 0.15.15 (CI pin) check + format clean.

## Rollout note for the cloud pin bump

The catalog hash changes and the doc host refuses to apply/project a document minted under a different pin (`catalog_mismatch`). Existing `workflow_docs` rows need `cmd/crdt-cohort -remint` (or an automatic re-mint on mismatch) when the pin reaches an env with minted documents. Tracked on BE-10283.

## Also in this PR: `comfy generate` results speak `envelope/1` (d2f0bfeb)

The submit / sync / resume *result* paths of `comfy generate` still wrote a bare partner JSON blob via `output.print_json`, so an agent parsing stdout by the documented contract got a document with no `schema`/`ok`/`error` discriminator. `list` / `schema` had already moved onto the renderer envelope.

- `_emit_result`: in JSON/NDJSON modes the partner payload is wrapped as `data.result` (verbatim) with `data.saved` listing `--download` artifacts, emitted through the renderer. Pretty mode with a tail `--json` keeps the legacy raw blob — additive for machine consumers only.
- New `comfy_cli/schemas/generate_result.json`, registered as `COMMAND_SCHEMAS["comfy generate"]` so `comfy discover` advertises it.
- New `tests/comfy_cli/command/generate/test_result_envelope.py` (5 cases: JSON + NDJSON envelope, `data.saved`, schema validation, pretty-mode guard).
- Full suite on `main` + this change: 6153 passed, 38 skipped; generate/ + output/ + this PR's widget tests on the branch: 678 passed. ruff check + format clean.

> Note: the `public-repo-hygiene` failure on this PR is `tests/comfy_cli/command/test_deprecated_nodes.py:7` (`BE-7684`), which arrived on `main` in #808 — hygiene is red on `main` itself since `929198f6`. Nothing in this PR adds an internal ticket ID.

### DIFF
--- a/comfy_cli/command/generate/app.py
+++ b/comfy_cli/command/generate/app.py
@@ -340,37 +340,14 @@ def _spinner() -> Progress:
 
 
 def _emit_result(result: poll.PollResult, *, request_id: str, download: str | None, as_json: bool) -> None:
-    if as_json:
-        # Honor --download in JSON mode too. Previously this returned before
-        # saving, so `--json --download` printed the URL but wrote no file,
-        # forcing callers to curl the URL by hand. Save first, then surface the
-        # local path alongside the response.
-        saved: list[str] = []
-        if download and result.status == "succeeded" and result.image_urls:
-            saved = [str(p) for p in output.save_urls(result.image_urls, download, request_id)]
-        renderer = get_renderer()
-        if renderer.is_json():
-            # JSON/NDJSON modes get the envelope/1 contract every other
-            # machine-readable command speaks: data.result wraps the partner
-            # payload, data.saved lists --download artifacts. Registered as
-            # COMMAND_SCHEMAS["comfy generate"] -> generate_result.json.
-            data: dict[str, Any] = {"result": result.raw}
-            if saved:
-                data["saved"] = saved
-            renderer.emit(data, ok=True, command="generate")
-            return
-        # Pretty mode with an explicit tail --json keeps the legacy raw blob.
-        if saved:
-            output.print_json({"result": result.raw, "saved": saved})
-        else:
-            output.print_json(result.raw)
-        return
+    renderer = get_renderer()
+    # One failure path for every output mode, checked FIRST: a terminally
+    # failed job is never a result. In JSON/NDJSON modes it is an ok=false
+    # envelope with a registered code (the generate_result schema promises
+    # exactly that); in pretty mode it is a red line plus the partner's raw
+    # response. Either way the exit code is 1 — a consumer that trusts ``ok``
+    # (or the exit code) must never read a failure as success.
     if result.status != "succeeded":
-        # A terminal non-succeeded job is a FAILURE, not a result, so it owes the
-        # caller an envelope even though the success paths above deliberately
-        # bypass the renderer. (The `as_json` branch returned already: that is
-        # the command-local `--json` raw-response contract, left untouched.)
-        renderer = get_renderer()
         message = f"Job {result.status}: {result.error or 'unknown error'}"
         if renderer.is_json():
             renderer.error(
@@ -386,6 +363,31 @@ def _emit_result(result: poll.PollResult, *, request_id: str, download: str | No
             rprint(f"[bold red]{sanitize_markup(message)}[/bold red]")
             output.print_json(result.raw)
         raise typer.Exit(code=1)
+    if renderer.is_json() or as_json:
+        # Honor --download in machine modes too. Previously this returned
+        # before saving, so `--json --download` printed the URL but wrote no
+        # file, forcing callers to curl the URL by hand. Save first, then
+        # surface the local path alongside the response.
+        saved: list[str] = []
+        if download and result.image_urls:
+            saved = [str(p) for p in output.save_urls(result.image_urls, download, request_id)]
+        if renderer.is_json():
+            # JSON/NDJSON modes (the global ``--output json``, a redirected
+            # stdout, or the tail ``--json``) get the envelope/1 contract every
+            # other machine-readable command speaks: data.result wraps the
+            # partner payload, data.saved lists --download artifacts.
+            # Registered as COMMAND_SCHEMAS["comfy generate"] -> generate_result.json.
+            data: dict[str, Any] = {"result": result.raw}
+            if saved:
+                data["saved"] = saved
+            renderer.emit(data, ok=True, command="generate")
+            return
+        # Pretty mode with an explicit tail --json keeps the legacy raw blob.
+        if saved:
+            output.print_json({"result": result.raw, "saved": saved})
+        else:
+            output.print_json(result.raw)
+        return
     if download and result.image_urls:
         saved = output.save_urls(result.image_urls, download, request_id)
         output.print_urls(result.image_urls, request_id=request_id)
--- a/comfy_cli/command/generate/app.py
+++ b/comfy_cli/command/generate/app.py
@@ -344,10 +344,24 @@ def _emit_result(result: poll.PollResult, *, request_id: str, download: str | No
         # Honor --download in JSON mode too. Previously this returned before
         # saving, so `--json --download` printed the URL but wrote no file,
         # forcing callers to curl the URL by hand. Save first, then surface the
-        # local path alongside the raw response.
+        # local path alongside the response.
+        saved: list[str] = []
         if download and result.status == "succeeded" and result.image_urls:
-            saved = output.save_urls(result.image_urls, download, request_id)
-            output.print_json({"result": result.raw, "saved": [str(p) for p in saved]})
+            saved = [str(p) for p in output.save_urls(result.image_urls, download, request_id)]
+        renderer = get_renderer()
+        if renderer.is_json():
+            # JSON/NDJSON modes get the envelope/1 contract every other
+            # machine-readable command speaks: data.result wraps the partner
+            # payload, data.saved lists --download artifacts. Registered as
+            # COMMAND_SCHEMAS["comfy generate"] -> generate_result.json.
+            data: dict[str, Any] = {"result": result.raw}
+            if saved:
+                data["saved"] = saved
+            renderer.emit(data, ok=True, command="generate")
+            return
+        # Pretty mode with an explicit tail --json keeps the legacy raw blob.
+        if saved:
+            output.print_json({"result": result.raw, "saved": saved})
         else:
             output.print_json(result.raw)
         return
--- a/comfy_cli/cql/engine.py
+++ b/comfy_cli/cql/engine.py
@@ -1376,8 +1376,15 @@ class Graph:
                 out[p.name] = None
             if _has_control_after_generate_slot(p):
                 out["control_after_generate"] = "fixed"
-        # Frontend-injected marker slots (``upload``, ``audioUI``) are trailing
-        # and ``serialize: false`` on current frontends: no default, no value.
+        # Frontend-injected slots: the ``upload``/``audioUI`` buttons are
+        # ``serialize: false`` on current frontends — no default, no value, so
+        # a fresh node ends before them (``_build_node`` drops trailing names
+        # with no default). The injected PREVIEW_3D ``image`` (SaveGLB /
+        # Preview3D) IS a DOM widget the frontend serializes as ``""`` — the
+        # captured shape is ``["mesh/ComfyUI", ""]`` — so it gets the same
+        # default a declared PREVIEW_3D port gets.
+        if m.id in _PREVIEW_3D_CLASSES and "image" not in out and "image" in frontend_extra_widget_names(m):
+            out["image"] = _FRONTEND_DOM_WIDGET_DEFAULTS["PREVIEW_3D"]
         return out
 
     # -- Validation --
--- a/comfy_cli/cql/engine.py
+++ b/comfy_cli/cql/engine.py
@@ -30,6 +30,42 @@ from comfy_cli.http import NoRedirectHandler, build_http_only_opener
 
 _IMPLICIT_WIDGET_TYPES = frozenset({"STRING", "INT", "FLOAT", "NUMBER", "BOOLEAN", "COMBO"})
 
+# Uppercase custom types the frontend renders as a DOM widget that SERIALIZES
+# into ``widgets_values`` (``ComponentWidgetImpl`` / ``DOMWidget``), so they
+# occupy a positional slot exactly like an INT. Every other uppercase custom
+# type is a link. Kept to types verified against saved workflows:
+# ``Load3D.image`` / ``Load3DAdvanced.viewport_state`` (``LOAD_3D``),
+# ``SaveGLB``/``Preview3D``'s injected ``image`` (``PREVIEW_3D``),
+# ``LoadAudioUI.audioUI`` (``AUDIO_UI``). ``LOAD3D_CAMERA`` is deliberately
+# absent: ``camera_info`` is ``serialize: false`` and writes no slot.
+_FRONTEND_DOM_WIDGET_TYPES = frozenset(
+    {"LOAD_3D", "LOAD_3D_ADVANCED", "PREVIEW_3D", "AUDIO_UI", "IMAGEUPLOAD", "AUDIOUPLOAD"}
+)
+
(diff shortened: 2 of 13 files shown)


comfy_cli/cql/engine.py
```diff
@@ -2385,6 +2482,8 @@ def emit(name: str, port: Port, owner: str | None, depth: int) -> None:
         if p.is_link:
             continue
         emit(p.name, p, None, 0)
+    for name in frontend_extra_widget_names(m):
+        entries.append(_WidgetEntry(name=name, port=None, owner=None))
```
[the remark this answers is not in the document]
skishore23: Fixed in 32f14bcc311954706875d4fa8a9b587584209534.

`_WidgetEntry` now carries a `frontend_injected` flag; `_expand_widget_entries` sets it for every `frontend_extra_widget_names()` entry (`upload`, `audioUI`, the `PREVIEW_3D` `image` on SaveGLB/Preview3D). Those entries stay in the positional order (they own a `widgets_values` slot), but every write surface now refuses them by name with one shared error (`frontend_injected_widget_error`): `set-widget` and `apply` replay via `_widget_index`, `set-slot`/`vary` via `_write_widget`. The "available widgets" list on both lookups is derived from schema-backed entries (`Graph.editable_widget_names`) — what `slots` advertises — rather than filtering the order by marker name, so `SaveGLB.image` is neither writable nor suggested.

`control_after_generate` is also `port=None` but carries a real serialized user value (`fixed`/`randomize`/…), so it stays writable and unadvertised, as before — pinned by tests. Tests added in `tests/comfy_cli/cql/test_frontend_widget_slots.py` cover all three surfaces, the not-found list, positional order, and the marker policy.

comfy_cli/command/generate/app.py
```diff
@@ -344,10 +344,24 @@ def _emit_result(result: poll.PollResult, *, request_id: str, download: str | No
...
+        renderer = get_renderer()
+        if renderer.is_json():
+            # JSON/NDJSON modes get the envelope/1 contract every other
+            # machine-readable command speaks: data.result wraps the partner
+            # payload, data.saved lists --download artifacts. Registered as
+            # COMMAND_SCHEMAS["comfy generate"] -> generate_result.json.
+            data: dict[str, Any] = {"result": result.raw}
+            if saved:
+                data["saved"] = saved
+            renderer.emit(data, ok=True, command="generate")
```
annehe9: **Blocking: a terminally failed job emits `ok: true` and exits 0.**

This `as_json` block returns before the `result.status != "succeeded"` check further down, so a failed job never reaches `renderer.error(code="generate_job_failed", ...)` or the `typer.Exit(code=1)`.

Reproduced by calling `_emit_result` directly with a failed `PollResult` under `--json` plus `--output json`:

```json
{"schema":"envelope/1","type":"envelope","ok":true,"command":"generate","data":{"result":{"status":"failed","error":"partner blew up"}},"error":null}
```

Exit code 0, `error: null`, `ok: true`.

`generate_result.json` promises the opposite in its own description: "Failures do not use this schema — they arrive as an ok=false envelope with a registered error.code."

Exit 0 on a failed job is pre-existing, but the bare blob carried no `ok` key, so the CLI never asserted success. It does now, and machine consumers are being told to trust that field.

Suggest hoisting the non-succeeded check above the `if as_json:` block so both output modes share one failure path.

annehe9 (commented):
Review of the widget-order and generate-envelope changes. Two blocking items on the bundled `generate` commit and two on the CQL side. The widget-order approach itself looks right and the captured shapes are well covered by the new tests.

I ran the suite locally in a worktree at `7aba70f`: `tests/comfy_cli/cql` 392 passed, `tests/comfy_cli` minus cloud/command 2904 passed / 11 skipped, `tests/comfy_cli/cloud` 197 passed, `test_result_envelope.py` 5 passed. The 6 failures I saw are all ANSI-escape-stripping assertions that fail on `main` too, so they are my terminal forcing color rather than anything here. My local ruff is 0.12.7 against the CI pin of 0.15.15, so I could not verify the lint claim.

comfy_cli/command/generate/app.py
```diff
@@ -344,10 +344,24 @@ def _emit_result(result: poll.PollResult, *, request_id: str, download: str | No
...
         # forcing callers to curl the URL by hand. Save first, then surface the
-        # local path alongside the raw response.
+        # local path alongside the response.
+        saved: list[str] = []
         if download and result.status == "succeeded" and result.image_urls:
-            saved = output.save_urls(result.image_urls, download, request_id)
-            output.print_json({"result": result.raw, "saved": [str(p) for p in saved]})
+            saved = [str(p) for p in output.save_urls(result.image_urls, download, request_id)]
+        renderer = get_renderer()
+        if renderer.is_json():
```
annehe9: **Blocking: `--output json` on its own never reaches this branch.**

`as_json` comes only from the tail `--json` meta flag (`_separate_meta_flags`, `meta_names` includes `"json"`). The global `--output json` never sets it, so `comfy --output json generate <model> ...` skips the whole `as_json` block and lands on `output.print_urls`.

Verified stdout for a succeeded result with `Renderer(OutputMode.JSON)` and `as_json=False`:

```
'\x1b[1;32mRequest:\x1b[0m req1\n\x1b[1;32mOutputs:\x1b[0m\n  \x1b[4;94mhttp://x/1.png\x1b[0m\n'
```

ANSI escapes on stdout in JSON mode.

The failure path below checks `renderer.is_json()` with no `as_json` gate, so under `--output json` a failure gets an envelope and a success gets colored text. That split is worse than either mode alone.

The CHANGELOG entry says results now speak envelope/1 "in `--output json` / `ndjson` modes". As written it only happens when the caller also passes the legacy tail `--json`. Either gate the envelope on `renderer.is_json()` alone, or narrow the CHANGELOG claim.

comfy_cli/cql/engine.py
```diff
@@ -1256,10 +1370,14 @@ def widget_defaults(self, class_name: str) -> dict[str, Any]:
...
             elif p.enum_values:
                 out[p.name] = p.enum_values[0]
+            elif p.type in _FRONTEND_DOM_WIDGET_TYPES:
+                out[p.name] = _FRONTEND_DOM_WIDGET_DEFAULTS.get(p.type)
             else:
                 out[p.name] = None
             if _has_control_after_generate_slot(p):
                 out["control_after_generate"] = "fixed"
+        # Frontend-injected marker slots (``upload``, ``audioUI``) are trailing
+        # and ``serialize: false`` on current frontends: no default, no value.
```
annehe9: **`add_node` does emit a value for these slots, contrary to this comment.**

`_build_node` in `workflow_ops.py` materializes widgets as:

```python
widgets = [defaults.get(name) for name in graph.widget_order_default(class_type)]
```

`widget_order_default` now includes the injected names, and `widget_defaults` deliberately omits them, so `.get` returns `None` for each. A fresh node carries a trailing `null`, not nothing.

Measured against a synthetic catalog:

```
LoadImage  order=['image','upload']            -> ["beach.jpg", null]
SaveGLB    order=['filename_prefix','image']   -> ["mesh/ComfyUI", null]
Preview3D  order=['model_file','image']        -> ["", null]
LoadAudio  order=['audio','audioUI','upload']  -> ["a.wav", null, null]
```

`LoadAudio` matches the captured shape in the PR description. `SaveGLB` does not: the description records `["mesh/ComfyUI", ""]`, and this writes `null`. `_FRONTEND_DOM_WIDGET_DEFAULTS["PREVIEW_3D"] = ""` is applied a few lines above for a *declared* `PREVIEW_3D` port, but the `SaveGLB`/`Preview3D` `image` arrives through `frontend_extra_widget_names` and never sees it.

The `widget_catalog.py` docstring makes the same claim ("a fresh node omits them").

Two ways out. Either have `_build_node` drop trailing names that have no default, or give the injected `image` its `""` in `widget_defaults` so `add_node` matches the shape the frontend writes.

tests/comfy_cli/cql/test_frontend_widget_slots.py
```diff
@@ -0,0 +1,495 @@
...
+
+    def test_file_upload_flag_alone_gets_no_button(self, graph: Graph):
+        # Load3D's model_file carries file_upload; the 3D loader handles its
+        # own upload and the frontend attaches no IMAGEUPLOAD button.
+        assert "upload" not in graph.widget_order("Load3D")
+
+    def test_upload_has_no_add_node_default(self, graph: Graph):
+        # Current frontends mark the button ``serialize: false``; a fresh node
+        # must not carry a phantom trailing value for it.
+        assert "upload" not in graph.widget_defaults("LoadImage")
```
annehe9, line 259: This test name and comment claim a property about `add_node`, but the assertions only inspect the `widget_defaults` dict.

`add_node` reads that dict through `[defaults.get(name) for name in widget_order_default(...)]`, so a missing key becomes a `None` in the node. A fresh `LoadImage` from `add_node` is `["beach.jpg", null]`, which is exactly the phantom trailing value the comment says must not happen.

Worth asserting on the built node's `widgets_values` instead, which is the layer the name is about.

comfy_cli/workflow_ops.py
```diff
@@ -2165,11 +2166,20 @@ def _widget_index(graph, class_type: str, widget: str, widgets_values=None) -> i
     # selected key (from ``widgets_values``), not the schema's first key, so the
     # index stays aligned to ``widgets_values`` for the node's real selection.
     order = graph.widget_order_for_node(class_type, widgets_values)
+    # A frontend-injected slot (``upload``/``audioUI``/``PREVIEW_3D`` ``image``)
+    # sits in ``order`` — it owns a position — but has no schema port to
+    # validate against and ``slots`` never advertises it; refuse it by name so
+    # it can't become a ghost target. ``control_after_generate`` is also
+    # unadvertised but stays writable (it carries a real user value).
+    if widget in graph.frontend_injected_widget_names(class_type):
```
annehe9, line 2182: **`capture_recipe` still accepts an injected slot as a `--param` target.**

Good fix here, but the up-front lift validation in `capture_recipe` checks against `widget_order_default`, which now contains the injected names:

```python
if widget not in graph.widget_order_default(node.get("type", "")):
    raise RecipeError(...)
```

So `--param <id>.upload` passes capture and only blows up later at apply, from this function. Confirmed: `capture_recipe(..., lift={(1, "upload"): "up"})` returns a recipe, and `_widget_index` then raises the frontend-injected error at apply time.

The comment above that loop says "Validate lift targets up front — no silently-ignored typos", so this looks like it wants the same treatment. `graph.editable_widget_names(...)` that this PR adds is a one-line swap.

comfy_cli/command/generate/app.py
skishore23: Fixed in d7b8522d: the non-succeeded check runs first for every output mode (ok=false envelope + exit 1 in JSON; red line + raw + exit 1 in pretty); pinned by test_failed_job_is_an_ok_false_envelope_and_exit_1_in_json_mode.

comfy_cli/command/generate/app.py
skishore23: Fixed in d7b8522d: the envelope is gated on renderer.is_json() alone (global --output json / redirected stdout), the tail --json only adds the legacy raw blob in pretty mode; pinned by test_output_json_alone_gets_the_envelope_not_colored_text. The two pretty-mode tests in test_app.py that relied on non-TTY-without-flag now pass --no-json like the rest of that file.

comfy_cli/cql/engine.py
skishore23: Fixed in d7b8522d: _build_node drops trailing names with no default (upload/audioUI → nothing, as the frontend writes), and the injected PREVIEW_3D image gets "" so SaveGLB/Preview3D match the captured shapes; comment corrected. Pinned by test_fresh_node_carries_no_phantom_value_for_the_upload_button and test_fresh_3d_nodes_carry_the_frontends_empty_preview_slot (asserted on the built node).

tests/comfy_cli/cql/test_frontend_widget_slots.py
skishore23: Fixed in d7b8522d: the test now builds the node with add_node and asserts widgets_values (LoadImage → ["beach.jpg"], LoadAudio → ["a.wav"]).

comfy_cli/workflow_ops.py
skishore23: Fixed in d7b8522d: capture_recipe validates lift targets against graph.editable_widget_names, so  is refused up front; pinned by test_capture_recipe_refuses_an_injected_slot_as_a_param_up_front.

annehe9 (approved):

PR #812, opened 2026-08-28
fix(cql): grow nested auto-grow slots under dynamic combos; stop counting link-only subs as widgets

skishore23:
> **Stacked on #809** (`kishore/upload-companion-widget`) — both change the same `engine.py` widget-order functions. Retarget to `main` once #809 lands.

## Why

The in-app agent cannot wire a Load Image / Load Video / Load Audio into any node whose reference inputs are `COMFY_AUTOGROW_V3` groups nested under a `COMFY_DYNAMICCOMBO_V3` option — Nano Banana 2, Seedance 2.5 r2v, MiniMax H3 r2v, and 24 other partner nodes. Reproduced on the pin the agent ships (`50fd0a43`) with the cloud `object_info.json`:

- agent-built `GeminiNanoBanana2V2`: `add_node` writes `inputs: []`; `connect A.IMAGE B.model.images.image_1` and `…B.model.images` both fail `input not found`;
- UI-built `api_minimax_h3_r2v`: only the one free slot the frontend pre-created (`image_2`) wires; `image_3`, `reference_videos.video_2` fail;
- agent-built dynamic-combo nodes carried **phantom widget slots** for the groups (`model.images`, `model.files`), shifting every later widget: converting Nano Banana 2 gave `response_modalities: null, system_prompt: 42, temperature: "fixed", top_p: "IMAGE"`. The published widget catalog named those phantom slots too, so through the CRDT doc host a UI-built MiniMax H3 node's `seed` position mapped onto `model.reference_images` ("filename in the seed widget").

## Root cause

**A.** `workflow_ops._resolve_input_target` only recognised a group when the node's `inputs` already held a `COMFY_AUTOGROW*`-typed base entry — true only for *top-level* groups on *CLI-built* nodes. Nested groups never get one; UI-built nodes never carry one (the frontend writes the grown `images.imageN` slots, never the base). So even a UI-built top-level `BatchImagesNode` couldn't be base-addressed or grown past its free slot.

**B.** `widget_order_default` / `widget_defaults` expanded a dynamic combo's sub-inputs with their own helper (`_dynamic_sub_widget_defaults`) that included link-only sub-inputs, while `widget_order_for_node` (`_expand_widget_entries`) skipped them. add-node, the catalog and set-widget disagreed with each other.

## Change

- `engine.py`: one walk. `_expand_widget_entries(m, values, first_key=…, sub_links=…)` drives `widget_order_default`, `widget_defaults` (new `_widget_default`) *and* `widget_order_for_node`; the duplicate sub-widget helpers are removed. New `Graph.autogrow_groups(class, widgets_values)` lists every group the node's **current selection** exposes, dotted as the frontend names them (`model.reference_images`). New `Port.autogrow_element_type` (the template's declared input type) and `Port.autogrow_limits` (`(min, max)` with `max = len(names)` else `template.max`, as `applyAutogrow` derives them). `autogrow_slot_example` uses the schema names. `morphism_to_dict` (`nodes show`) adds `dynamic_options: [{key, inputs:[…]}]` per option and, on any autogrow input, `element_type`, `slots {names|prefix, min, max}`, `wire_as`.
- `workflow_ops.py`: `_resolve_schema_autogrow` mirrors `dynamicWidgets.ts` — base address reuses the lowest **free** existing slot, else appends the next schema-named slot; a dotted key or bare element name must be that exact next slot (gap → rejected with the next free key); source type must match the group's element type (`type mismatch`); never grows past `max`. `_autogrow_group_port` resolves top-level *and* nested groups from the node's own `widgets_values`. Grow/apply splits a slot's group on the **last** dot (`_autogrow_base`) so `model.images.image_1` → `model.images`, not `model`.
- Fixture `tests/comfy_cli/fixtures/object_info_nested_autogrow.json`: the production catalog's entries for `LoadImage/LoadVideo/LoadAudio/SaveImage/BatchImagesNode/GeminiNanoBanana2V2/MinimaxHailuo03ReferenceNode/GrokImageEditNodeV2`, tooltips stripped, declaration order preserved.

## Red → green

26 new tests, all written and watched fail first (`11 == 8`, `AttributeError: autogrow_groups`, `input 'model.images.image_1' not found … inputs: []`):

- `tests/comfy_cli/cql/test_nested_autogrow.py` (10): default order = the captured frontend shape (MiniMax 8 values, `seed` at 5); add-node layout round-trips through set-widget indexing; `autogrow_groups` follows the node selection; template/limits from the schema; `nodes show` payload names group, element type, slots.
- `tests/comfy_cli/command/test_workflow_edit_nested_autogrow.py` (16): grow by name / by base / explicit next key; gap rejected with the next free key; bare element name; selection-dependent groups; UI-built free-slot reuse then growth; Load Video + Load Audio into their groups; type mismatch refused; `max` enforced; UI-built top-level `BatchImagesNode`; set-widget lands in the frontend position; agent-built node converts with `model.images.image_1/2` links and unshifted `seed`/`temperature`/`top_p` and validates; nested grow op replay idempotent; CLI `add-node` + `connect` envelopes.

## Verification

- Full suite: **6207 passed, 38 skipped**; ruff check + format clean.
- Cloud catalog sweep (3,699 classes): classes whose fresh add-node layout ≠ set-widget indexing **41 → 0**; 85 auto-grow groups (45 nested, 40 top-level) resolvable at the default selection.
- Report's repro commands with the real catalog: Nano Banana 2 `image_1` / base / `image_3` ✓, `image_9` → "use the next free key `model.images.image_4`"; MiniMax template `image_2` / `image_3` / base ✓, `LoadVideo → reference_videos.video_1` ✓, `LoadVideo → reference_images` → `type mismatch … its slots take IMAGE`; widgets stay `[…, 42, 'randomize', False]`; widget-catalog MiniMax = 8 names, NB2 = 11.

## Not in this PR (other children of the umbrella)

- Agent tool surface (`services/agent … digest.go`, `list_slots`, `connect` description) — the CLI half is done: `nodes show` now carries the names; the digest must keep `dynamic_options` / `element_type` / `wire_as`.
- comfy-knowledge canon rows "agent tooling cannot wire a Load Image to a slot that is not there yet" (nano-banana.yaml, seedream.yaml) — retract.
- In-repo follow-ups: nested groups with `min > 0` (Grok / Qwen / HappyHorse) aren't checked by `_check_autogrow_required`; grown slots carry no `label` / `shape` (the frontend recomputes them on load).

## Rollout note

Same as #809: the catalog hash changes (41 classes), so minted `workflow_docs` need a re-mint when the pin reaches an env with documents.

### DIFF
--- a/comfy_cli/workflow_ops.py
+++ b/comfy_cli/workflow_ops.py
@@ -401,23 +401,35 @@ def _lww_commit(workflow: dict, op: dict) -> None:
     workflow.setdefault("_widget_stamps", {})[json.dumps(_write_target(op), default=str)] = _stamp_key(op)
 
 
-def _autogrow_template(graph, node_type: str, base: str) -> dict | None:
-    """The schema-declared element-naming template for the ``base`` autogrow
-    input on ``node_type`` — looked up from the same object_info-derived
-    ``graph`` the connect path already resolves node schemas from (see
-    ``cql.engine.Port.autogrow_template``). None when ``graph`` is unavailable,
-    ``node_type`` isn't in the catalog (offline edit), or the catalog entry
-    carries no template — callers then fall back to the historical
-    pluralization heuristic in :func:`_autogrow_elem_name`."""
+def _autogrow_group_port(graph, node: dict, base: str):
+    """The schema :class:`~comfy_cli.cql.engine.Port` for autogrow group
+    ``base`` on ``node`` — a top-level group (``BatchImagesNode.images``) or
+    one nested under the option ``node`` currently selects at a dynamic combo
+    (``MinimaxHailuo03ReferenceNode`` → ``model.reference_images``). The
+    selection is read from the node's own ``widgets_values`` the same way
+    ``set-widget`` indexes them, so a node switched to an option without the
+    group stops resolving it. None when ``graph`` is unavailable, the class
+    isn't in the catalog (offline edit), or no such group exists."""
     if graph is None:
         return None
-    m = graph.node(node_type)
-    if m is None:
-        return None
-    for p in m.inputs:
-        if p.name == base and p.is_autogrow:
-            return p.autogrow_template
+    from comfy_cli.cql import engine as _engine
+
+    node_type = node.get("type", "")
+    values = _engine._widgets_as_positional(node.get("widgets_values"), graph, node_type)
+    for port in graph.autogrow_groups(node_type, values):
+        if port.name == base:
+            return port
     return None
+
+
+def _autogrow_template(graph, node: dict, base: str) -> dict | None:
+    """The schema-declared element-naming template for the ``base`` autogrow
+    group on ``node`` (see :func:`_autogrow_group_port` and
+    ``cql.engine.Port.autogrow_template``). None when the schema is
+    unavailable or carries no template — callers then fall back to the
+    historical pluralization heuristic in :func:`_autogrow_elem_name`."""
+    port = _autogrow_group_port(graph, node, base)
+    return None if port is None else port.autogrow_template
 
 
 def _autogrow_elem_name(base: str, n: int, template: dict | None) -> str:
@@ -467,9 +479,17 @@ def _next_autogrow_name(ins: list, requested: str, template: dict | None = None)
     taken = {i.get("name") for i in ins}
     if requested not in taken:
         return requested
-    base = requested.split(".", 1)[0]
+    base = _autogrow_base(requested)
     n = _first_free_autogrow_index(taken, base, template)
     return f"{base}.{_autogrow_elem_name(base, n, template)}"
+
+
+def _autogrow_base(slot_name: str) -> str:
+    """The group a grown slot name belongs to: everything before the LAST
+    dot. Element names never contain a dot, but a group nested under a
+    dynamic combo does (``model.images.image_1`` → ``model.images``), so
+    splitting on the first dot would name the combo, not the group."""
+    return slot_name.rsplit(".", 1)[0]
 
 
 def _next_inputcount_name(ins: list, requested: str) -> str:
@@ -1785,8 +1805,8 @@ def _apply_connect(workflow: dict, op: dict, graph) -> None:
                 # base.elemN fallback — that name is meaningless for this family.
                 name = _next_inputcount_name(ins, grow["name"])
             else:
-                base = str(grow["name"]).split(".", 1)[0]
-                template = None if grow.get("widget") else _autogrow_template(graph, dst.get("type", ""), base)
+                base = _autogrow_base(str(grow["name"]))
+                template = None if grow.get("widget") else _autogrow_template(graph, dst, base)
                 name = _next_autogrow_name(ins, grow["name"], template)
             entry = {
                 "name": name,
@@ -2373,13 +2393,12 @@ def _resolve_input_target(node: dict, graph, slot: Any, elem_type: str | None) -
       the API converter reads the link and skips the widget by name.
     """
     ins = node.get("inputs") or []
-    node_type = node.get("type", "")
     # Concrete slot (index or exact name) that is NOT an autogrow base.
     try:
         idx = _resolve_input_slot(node, None, slot)
         if str(ins[idx].get("type", "")).startswith("COMFY_AUTOGROW"):
             base = ins[idx].get("name")
-            template = _autogrow_template(graph, node_type, base)
+            template = _autogrow_template(graph, node, base)
             return None, _plan_autogrow(ins, base, elem_type, template)
         return idx, None
     except ValueError:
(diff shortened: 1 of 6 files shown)


skishore23:
@coderabbitai review

comfy_cli/workflow_ops.py
```diff
@@ -467,11 +479,19 @@ def _next_autogrow_name(ins: list, requested: str, template: dict | None = None)
...
     n = _first_free_autogrow_index(taken, base, template)
     return f"{base}.{_autogrow_elem_name(base, n, template)}"
 
 
+def _autogrow_base(slot_name: str) -> str:
+    """The group a grown slot name belongs to: everything before the LAST
+    dot. Element names never contain a dot, but a group nested under a
+    dynamic combo does (``model.images.image_1`` → ``model.images``), so
+    splitting on the first dot would name the combo, not the group."""
+    return slot_name.rsplit(".", 1)[0]
```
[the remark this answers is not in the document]
skishore23: Fixed in a7654f57: the grow conflict target is keyed on `_autogrow_base` (the group before the last dot); pinned by test_grows_on_different_nested_groups_are_different_conflict_targets.

skishore23:
@coderabbitai review

skishore23:
@coderabbitai review

annehe9 (commented):
Reviewed the diff and ran it locally at 51d0737: the 27 new tests pass, all 402 `tests/comfy_cli/cql` tests pass, and 288 targeted workflow/nodes/connect tests pass. Ruff is clean relative to the base (the only delta is 3 `UP038` hits, the same rule and the same `isinstance(x, (int, float))` style as 21 pre-existing ones on `main`).

The design is right. Folding `widget_order_default`, `widget_defaults` and `widget_order_for_node` into one `_expand_widget_entries` walk is the correct fix for root cause B, because the three can no longer disagree when they are the same code path. Root cause A is real and resolving the group from the schema rather than from the node's `inputs` is the right answer.

One blocking issue, inline. Three non-blocking notes I left out of the review and can add if useful: the new `_resolve_schema_autogrow` call at line 2428 sits above the widget-backed branch at 2433, so the bare-element arm now outranks a real widget name on any node carrying an autogrow group (no collision in the fixture, but the precedence did flip); the "not a valid autogrow slot" raise at 2539 runs before the "is full" raise at 2546, so a gapped key on a full group is told to `use the next free key 'model.images.image_1414'`, which the next call refuses; and one connect can run `_expand_widget_entries` three or four times with nothing cached, which only matters on bulk apply-ops and the catalog sweep.

comfy_cli/workflow_ops.py
```diff
@@ -2383,7 +2405,7 @@ def _resolve_input_target(node: dict, graph, slot: Any, elem_type: str | None) -
             (i for i in ins if i.get("name") == base and str(i.get("type", "")).startswith("COMFY_AUTOGROW")), None
         )
         if ag is not None:
-            template = _autogrow_template(graph, node_type, base)
+            template = _autogrow_template(graph, node, base)
             grow = _plan_autogrow(ins, base, elem_type, template)  # canonical next sequential slot
```
annehe9, line 2426: **Blocking: the new type and `max` guards only fire on one of the two branches that grow a slot.**

`_resolve_input_target` has three paths. Paths 1 and 2 (line 2397 and this line) run when the node's `inputs` already carry the `COMFY_AUTOGROW` base entry, and both call bare `_plan_autogrow`, which checks neither the declared element type nor `max`. Only path 3 (`_resolve_schema_autogrow`, line 2428) has those checks.

`_build_node` at line 2166 writes that base entry for every top-level group (`p.is_link` is true for an autogrow port), so this CLI's own `add-node` output is exactly the shape that skips the guards. Same node class, opposite behaviour:

```
add_node mints inputs: [{"name": "images", "type": "COMFY_AUTOGROW_V3", "link": null}]
connect AUDIO -> images: (None, {'name': 'images.image0', 'type': 'AUDIO'})   # CLI-built: accepted
                              # UI-built (no base entry): "type mismatch: AUDIO output
                              #   cannot connect to autogrow input 'images' … its slots take IMAGE"
```

`max` splits the same way. With all 50 slots filled, a CLI-built `BatchImagesNode` mints `images.image50`; a UI-built one raises `is full: the schema allows at most 50 slots`.

Reproduced on this branch against the PR's own fixture:

```python
import json
from comfy_cli.cql.engine import Graph
from comfy_cli import workflow_ops as W

g = Graph.from_object_info(json.load(open("tests/comfy_cli/fixtures/object_info_nested_autogrow.json")))
n = W._build_node(1, "BatchImagesNode", g.node("BatchImagesNode"), g, [0, 0], [200, 60])
W._resolve_input_target(n, g, "images", "AUDIO")
# -> (None, {'name': 'images.image0', 'type': 'AUDIO'})
```

This is not a regression, paths 1 and 2 were always this loose. But it means the guard the PR adds misses the node shape the CLI itself produces, which is the shape the agent hits most often. Suggest putting the type and `max` check where all three paths route, or having paths 1 and 2 defer to `_resolve_schema_autogrow` whenever the schema knows the group.

skishore23: Fixed in 1df17589: both base-entry paths now defer to _resolve_schema_autogrow whenever the schema knows the group, so the CLI-built shape gets the same element-type and max checks as the UI-built one — pinned by test_cli_built_top_level_group_refuses_the_wrong_element_type / _enforces_max (your repro). Your two notes too: the widget→input conversion is checked before the bare-element arm (test_a_real_widget_name_outranks_the_bare_element_guess), and a full group reports full before any next-key hint (test_full_group_reports_full_before_the_next_key_hint). Left the repeated _expand_widget_entries walks as-is for now.

annehe9 (approved):

skishore23:
GitHub closed this when #809 merged and deleted its base branch (the retarget to main did not take). Reopened as #825 on the same branch rebased onto main — same commits, same review-fix content.

PR #813, opened 2026-08-28
feat(knowledge): fetch the bundle from the cloud knowledge channel by default (BE-8952)

annehe9:
## Summary

- With `COMFY_KNOWLEDGE_URL` and `COMFY_KNOWLEDGE_FILE` both unset, the loader now fetches `knowledge.json` from `/api/knowledge/knowledge.json` under the cloud base URL, so a stock signed-in install loads a bundle with no configuration.
- The URL is derived from `get_base_url()` on every call and never stored, because `_http_get` only attaches credentials to URLs under that base and the base resolves per invocation from `COMFY_CLOUD_BASE_URL`, the persisted `cloud_base_url`, then `cloud.comfy.org`.
- An explicit `COMFY_KNOWLEDGE_URL` still wins. `COMFY_KNOWLEDGE_FILE` still skips the fetch entirely.
- `REASON_NO_URL` was unreachable and is replaced by `REASON_SIGNED_OUT`, which names the usual cause of a failed default fetch. `comfy knowledge status` reports it together with the effective URL.
- Cache, TTL and validation are unchanged. Env-only, no persisted config key.

## Login warm

`comfy cloud login` now calls `knowledge.refresh_if_stale()` after the session is saved. The attach path loads cache-only, so without this a fresh install had no knowledge block until the user ran a knowledge command. The warm skips when the cache is fresh or `COMFY_KNOWLEDGE_FILE` is set, and swallows every failure, so login cannot break because of it.

## Tests

- New: default URL follows the cloud base URL, explicit env var overrides it, `COMFY_KNOWLEDGE_FILE` bypasses the fetch, signed-out 401 degrades to `loaded=false` with the new reason and no exception.
- Login: warm runs after `save_cloud_session`, and never runs when login fails.
- Knowledge suites: `266 passed`. `ruff check .`: `All checks passed!`. `ruff format --check`: clean on touched files.

## First-run cost

With an empty cache against production: 0.69s wall, one `GET /api/knowledge/knowledge.json` request. The manifest fetch is skipped when the first request fails.

## Production check

Blocked on deploy. The endpoint returns 404 because cloud#7674 is merged but not yet deployed. Output was `loaded: false` with the signed-out reason, no exception, exit 0.

Closes BE-8952.

https://claude.ai/code/session_01MbcMMxBQyYjRk32iKtqnMX

### DIFF
--- a/README.md
+++ b/README.md
@@ -631,10 +631,11 @@ custom_nodes:
 
 Discovery commands can attach a `knowledge` block to their `--json` output: which
 model a name refers to, whether it is deprecated, ranked picks per capability,
-and known pitfalls. It is **off unless you point the CLI at a bundle**, so a
-default install never emits the block.
+and known pitfalls. A signed-in install (`comfy cloud login`) fetches the bundle
+from the cloud knowledge channel on its own. Signed out, the fetch fails quietly
+and no block is emitted.
 
-Turn it on with either of:
+Point the CLI at a different bundle with either of:
 
 ```
 export COMFY_KNOWLEDGE_URL=https://.../knowledge.json   # fetched and cached
@@ -655,8 +656,8 @@ Turn it off again:
 export COMFY_KNOWLEDGE_DISABLE=1
 ```
 
-Clearing `COMFY_KNOWLEDGE_URL` is *not* an off switch. Once a bundle is cached it
-keeps being served, stale or not. `COMFY_KNOWLEDGE_DISABLE` suppresses envelope
+Signing out is *not* an off switch. Once a bundle is cached it keeps being
+served, stale or not. `COMFY_KNOWLEDGE_DISABLE` suppresses envelope
 enrichment outright; the `comfy knowledge` verbs keep working under it, since
 those are you asking for the bundle directly.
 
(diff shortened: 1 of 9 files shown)


skishore23 (approved):
Approving. Verified the things that matter for a default-on fetch:

- Signed-out is quiet: `resolve_cloud_credential` → `None` → anonymous GET → 401 `HTTPError` → caught in `_fetch` → `REASON_SIGNED_OUT`. No prompt, no `SystemExit`, nothing on stderr.
- No new hot-path network: `attach()` is still cache-only; fetches happen only in the `knowledge` verbs, `skills install`, and the login warm.
- Login can't break: `refresh_if_stale` swallows `Exception` and nothing in the path raises `SystemExit`.
- Tests can't reach prod: autouse `_http_get` fake in `test_knowledge.py`, login tests mock the warm, no unit test invokes the `skills install` command. 330 passed on the branch.

One non-blocking comment for a follow-up: the fetch resolves credentials via `resolve_target(where="cloud")`, which defaults to `allow_clear=True`. `credentials.py`'s docstring says best-effort background readers should pass `allow_clear=False` so a spurious `invalid_grant` under concurrent refreshes (agents run many `comfy` processes in parallel) can't wipe the stored session. Pre-existing, but this PR makes the fetch default-on for everyone, so threading `allow_clear=False` through for the knowledge fetch is worth doing.

skishore23:
Re-checked 750a8d82: `allow_clear=False` is threaded `_http_get` → `resolve_target` → `resolve_cloud_credential` as suggested, and the reason-specific `knowledge_unavailable` hints are a nice touch. Knowledge, login, local_address and enrichment suites on the new head: 325 passed; ruff clean. Approval stands.

PR #814, opened 2026-08-28
perf(nodes): cache parsed annotations as JSON and add COMFY_CACHE_DIR (BE-9173)

annehe9:
Every `comfy` call that touched the node catalog parsed the annotation YAML twice: once in `load_annotation_bytes` (the cache-read validators run `yaml.safe_load`) and again in `Graph.annotate`. The pair cache is now trusted as written, and the parse result is stored as JSON keyed by the YAML bytes' digest.

Warm `_try_default_annotations` on the cloud catalog (3698 classes): **27.7 ms -> 0.6 ms**.

```
Cache dir: COMFY_CACHE_DIR > $XDG_CACHE_HOME/comfy-cli > ~/.cache/comfy-cli
Files:
  object_info-<sha256(base_url)[:16]>.json              (unchanged)
  comfy-complete/annotations.json                        (unchanged, raw YAML pair)
  comfy-complete/annotations-parsed-v1-<sha256(sup\0dis)[:16]>.json   (new, ~44 KB)
Invalidation: key is the annotation bytes' sha; v<N> bumps when the parse output shape changes.
Newest-only: writing a new parsed file deletes its siblings.
COMFY_NO_CACHE=1 skips the parsed cache. Cloud cli-runner sidecar: set COMFY_CACHE_DIR to a persistent path.
Dropped the pickled Graph: see BE-9173 comment with measurements.
```

One `file_utils.cache_dir()` replaces six copies of the XDG lookup (`cql/loader`, `cql/annotations_source`, `knowledge`, `command/outdated`, `command/templates` x2). On-disk paths are unchanged when `COMFY_CACHE_DIR` is unset.

## Verification

- `ruff check .` and `ruff format --check .` clean on ruff 0.15.15.
- `pytest`: 6164 passed. Two failures are local-environment only (`build_not_signed_in` login state, one colour-detection test) and pre-date this branch.
- `grep -rn XDG_CACHE_HOME comfy_cli/` matches only `file_utils.py`.
- `COMFY_CACHE_DIR=<tmp> comfy --json nodes show KSampler` writes both cache files under `<tmp>/comfy-complete/`.

https://claude.ai/code/session_01PTUJchybqGYXCiCatNvqjR

### DIFF
--- a/comfy_cli/command/outdated.py
+++ b/comfy_cli/command/outdated.py
@@ -44,7 +44,7 @@ import semver
 from comfy_cli._safe_exec import resolve_required_binary
 from comfy_cli.command.pack_scan import iter_pack_dirs as _iter_pack_dirs
 from comfy_cli.command.pack_scan import read_pyproject as _read_pyproject
-from comfy_cli.file_utils import atomic_write_text
+from comfy_cli.file_utils import atomic_write_text, cache_dir
 from comfy_cli.registry import RegistryAPI
 
 CACHE_TTL_SECONDS = 3600  # 1 hour
@@ -67,9 +67,8 @@ _VERSION_ASSIGN_RE = re.compile(
 
 
 def _cache_path() -> Path:
-    """Where the latest-version cache lives. XDG-respecting."""
-    base = os.environ.get("XDG_CACHE_HOME") or os.path.expanduser("~/.cache")
-    return Path(base) / "comfy-cli" / "outdated.json"
+    """Where the latest-version cache lives."""
+    return cache_dir() / "outdated.json"
 
 
 def _load_cache() -> dict[str, Any]:
--- a/comfy_cli/command/templates.py
+++ b/comfy_cli/command/templates.py
@@ -31,7 +31,7 @@ from typing import Annotated, Any
 import typer
 
 from comfy_cli import knowledge, tracking, workflow_ops
-from comfy_cli.file_utils import atomic_write_bytes
+from comfy_cli.file_utils import atomic_write_bytes, cache_dir
 from comfy_cli.http import ResponseTooLarge, plain_urlopen, read_capped
 from comfy_cli.output import get_renderer, rprint
 
@@ -80,9 +80,8 @@ _REFRESH_DEBOUNCE_SECONDS = 60.0
 
 
 def _cache_path() -> Path:
-    """Where the gallery index lives on disk. XDG-respecting."""
-    base = os.environ.get("XDG_CACHE_HOME") or os.path.expanduser("~/.cache")
-    return Path(base) / "comfy-cli" / "gallery" / "index.json"
+    """Where the gallery index lives on disk."""
+    return cache_dir() / "gallery" / "index.json"
 
 
 def _fetch_gallery(url: str = GALLERY_URL, timeout: float = 15.0) -> bytes:
@@ -1214,9 +1213,8 @@ def _template_workflow_cache_path(name: str) -> Path:
     ``quote(..., safe="")`` turns any ``/`` or ``..`` segment into ``%2F``/``..``
     with no path separators, keeping the file strictly inside ``gallery/templates``.
     """
-    base = os.environ.get("XDG_CACHE_HOME") or os.path.expanduser("~/.cache")
     safe_name = urllib.parse.quote(name, safe="")
-    return Path(base) / "comfy-cli" / "gallery" / "templates" / f"{safe_name}.json"
+    return cache_dir() / "gallery" / "templates" / f"{safe_name}.json"
 
 
 def _iter_workflow_nodes(wf: dict[str, Any]):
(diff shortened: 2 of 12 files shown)


skishore23 (dismissed):
Nice perf win and the `cache_dir()` consolidation is welcome. One high-severity regression is holding approval; three non-blocking notes below.

**1. (high) The pair cache is served without validation, which reintroduces the bug `test_poisoned_cache_is_ignored_in_favour_of_bundled` guarded against.** `_read_cached_pair` (`annotations_source.py:396`) now accepts any schema-1 entry whose bodies are strings. `parse_supported_nodes`/`parse_disable_config` (`engine.py:754-785`) swallow every parse exception into `({}, {})`/`set()`, so a schema-matching `annotations.json` with non-YAML bodies (hand edit — `comfy nodes refresh` prints the path — or a pre-seeded sidecar file now that `COMFY_CACHE_DIR` makes the location explicit) silently blanks every node's pack/labels/cloud_disabled for the whole TTL, indefinitely offline, and `_write_parsed` then persists that blank triple keyed by the garbage digest. The deleted test's docstring describes exactly this failure; the docstring's remaining "never a silent blank annotation" claim (`:371`) is no longer true. Fix that keeps the perf win: validate the pair only on the parsed-cache **miss** path (once per generation), and never persist a parse that yielded blank from non-empty bytes — treat it as an unusable cache and fall through to bundled. Please restore a test at the `load_annotation_bytes` level with `GARBAGE` bodies.

Non-blocking:

**2.** `cache_dir()` uses `pathlib.Path(...).expanduser()`, which raises `RuntimeError("Could not determine home directory")` for `COMFY_CACHE_DIR=~nosuchuser/...` (or `~/...` with HOME unset and the uid absent from passwd) — reproduced. None of the consumers catch `RuntimeError`, so `comfy nodes ls` / `outdated` / `templates` traceback instead of falling through. The call sites this replaced used `os.path.expanduser`, which never raises; use that here (and `.resolve()` while at it, so a relative `COMFY_CACHE_DIR` doesn't make every cache cwd-dependent and doesn't hand a relative cwd to `templates._refresh_cwd`'s detached child).

**3.** `_write_parsed` globs and unlinks every `annotations-parsed-*.json` before `atomic_write_text`. A failed write leaves the dir empty, and a concurrent reader between the unlink and `os.replace` gets a spurious miss. Write `path` first, then unlink siblings `!= path`.

**4.** `COMFY_NO_CACHE` sounds like a whole-subsystem switch (cf. `COMFY_NO_WATCH`, `COMFY_NO_TELEMETRY`) but only skips the parsed-annotation JSON; object_info, the annotation pair, gallery, outdated and knowledge caches ignore it. An agent running `COMFY_NO_CACHE=1 comfy nodes ls` after installing a pack still gets the stale object_info dump. Either scope the name (`COMFY_NO_ANNOTATION_CACHE`) or honour it in every `cache_dir()` consumer — and document both new vars in the README next to `COMFY_KNOWLEDGE_DISABLE`.

Ran `tests/comfy_cli/cql`, `test_file_utils`, `test_knowledge` on the branch: 478 passed, ruff clean — #1 isn't caught precisely because its test was removed.

annehe9:
@skishore23 thanks for the review — fixed in ac6ecfd:

1. **(high) Pair cache trusted without validation** — `_read_cached_pair` now re-runs the YAML validators, but only the first time a given byte-pair is seen. The gate is whether the parsed-annotation cache already has a digest file for those exact bytes; that file only exists if the pair previously parsed clean, so it's proof the bytes are good. A poisoned or hand-edited `annotations.json` gets a new digest and is caught immediately, falling through to bundled — same guarantee as before this PR. Once a pair is vouched for, repeat reads stay free of `yaml.safe_load`, so the perf win holds. Restored a `test_poisoned_cache_is_ignored_in_favour_of_bundled`-equivalent test, plus one that fails if the hot path ever re-parses YAML once vouched for.

2. **`cache_dir()` `RuntimeError` / cwd-dependent relative paths** — switched from `pathlib.Path.expanduser()` to `os.path.expanduser()` (never raises) and made the result absolute via `os.path.abspath`. Also fixed the concrete bug this caused: `templates.py`'s detached background refresher now passes its own already-resolved `cache_dir()` into the child's env instead of letting the child re-resolve a possibly-relative `COMFY_CACHE_DIR` against its own (different) cwd.

3. **`_write_parsed` evicts siblings before writing** — reordered to write the new file first, then evict old siblings, so a failed write can no longer leave the cache dir empty.

4. **`COMFY_NO_CACHE` naming/scope** — left as-is for now. Renaming or widening its scope felt like a bigger call than a review-fix pass, and the docstring already documents it as narrow. Open to revisiting if you'd rather it change now.

Also picked up two related hardening issues CodeRabbit/cursor-review flagged on the same lines: leaf-type validation on the parsed-annotation cache (a corrupted-but-valid-JSON file could otherwise hand back a non-str label and crash `Graph.annotate`), and a size cap + `RecursionError` guard on both cache reads (the `annotations.json` path isn't digest-derived, so it's a fixed, plantable path).

Full suite green: 6177 passed, 1 pre-existing local-only failure (not signed in to comfy cloud on this machine) unrelated to any of this.

skishore23 (approved):

PR #816, opened 2026-08-28
fix(print): promoted widgets print at the host address; auto-grow groups print as links

skishore23:
> **Stacked on #815** (`kishore/be-10305-…`) → #812 → #809. Retarget as they land.

## Why

`comfy workflow print` is how the agent reads a canvas before editing it. It was wrong for the two node shapes the last two PRs fixed:

- **Subgraph instances**: promoted values were read positionally from the instance's serialized `inputs[]` (only what the UI happened to show) instead of per widget-backed declared input the way the frontend does. Captured on real gallery templates: `audio_minimax_music_3` printed the caption under `switch=` and hid caption/lyrics/seed/models; `api_seedance2_5_video_extend` printed `drop_audio="lanczos"` (shifted) and dropped `interpolation`/`padding_color`; `image_z_image_turbo` printed `text=None` with all eight promoted widgets invisible and interior lines with no editable address.
- **Auto-grow groups**: an agent-built `GeminiNanoBanana2V2`'s `model.images` group was invisible; an agent-built `BatchImagesNode` printed a phantom `images=None`.

## Change (`comfy_cli/workflow_print.py`)

- Instance line: one argument per declared subgraph input in declaration order; promoted widgets print `promoted.effective_value` (host value, else interior default) under the address `set-widget` takes (`57.width`); an outside link into a promoted input prints as the link; sockets print their link; declared-but-unbacked inputs fall back to `proxyWidgets` exactly as `slots` does.
- Definition block: interior lines keep `IN.width`; the header gains `promoted widgets: IN.text, IN.width, … — each is the instance's own value: edit 57.<name>, never 57/<id>.<name>`.
- Auto-grow: groups from `Graph.autogrow_groups` for the node's current selection; grown slots print as links; when no free slot exists and the schema allows one, the next schema name prints as an open slot (the free trailing slot the frontend itself keeps — and exactly what `connect` grows next); the line comment names the group and element type (`model.images grows IMAGE (max 14)`). Widgets are never shifted.

## Red → green

`tests/comfy_cli/test_workflow_print_promoted_autogrow.py`: 21 cases written first, 17 failed on the base (4 pin already-correct value-aware naming/selector behaviour). One golden line in `test_workflow_print.py` updated for the group annotation.

## Verification

Full suite **6259 passed, 38 skipped**; ruff 0.15.15 check + format clean.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -17,6 +17,34 @@ history.
 
 ### Fixed
 
+- Promoted subgraph widgets are edited where the frontend reads them. The
+  frontend (ADR 0009) keeps a promoted widget's value on the HOST instance
+  (`widgets_values` positional over the widget-backed subgraph inputs) and
+  runs that value over the interior default; `set-widget 57.width` used to
+  follow the legacy `proxyWidgets` route into the interior node, so the edit
+  neither ran nor showed on the canvas. `set-widget`, `set-slot` and `vary`
+  now write the host value for `<instance>.<input>`, redirect an interior
+  address that backs a promotion (`57/13.width`) to the same host value
+  (`redirected_from` on the op), follow an outside link feeding the promoted
+  input to its primitive (`PrimitiveInt`/`PrimitiveString*`/legacy
+  `PrimitiveNode`, through `Reroute`s) and refuse — naming the driver — when
+  a non-primitive node computes the value. Unpromoted interior widgets still
+  write the definition. The op carries `promoted.value_index` and the
+  materialized `host_widgets_values`.
+- `comfy workflow connect` wires an outside node to a promoted widget
+  (`PrimitiveInt.INT → 57.width`): the definition declares the input, so it
+  is materialized on the instance (with the frontend's `widget` marker) and
+  linked, type-checked against the declared input type.
+- `comfy workflow slots` advertises promoted widgets at the instance address
+  with the value the frontend runs (host value, else interior default), flags
+  widgets fed by a link (`linked_from`), keeps unpromoted interior widgets
+  reachable at `<instance>/<inner>.<widget>` (nested instances included), and
+  no longer advertises the interior address behind a promotion.
+- `convert_ui_to_api` (`comfy run`, `validate`) applies host-owned promoted
+  values onto the expanded interior nodes — a post-migration template whose
+  host prompt differs from the interior default (`audio_minimax_music_3`:
+  interior caption `''`) was submitted with the interior value. Precedence
+  matches the frontend: outside link, then host value, then interior.
 - `comfy workflow connect` can wire a Load Image / Load Video / Load Audio
   into an auto-grow group nested under a dynamic combo (`GeminiNanoBanana2V2`
   `model.images`, `MinimaxHailuo03ReferenceNode` / `ByteDance2ReferenceNodeV2`
--- a/comfy_cli/cql/engine.py
+++ b/comfy_cli/cql/engine.py
@@ -2638,6 +2638,23 @@ def _node_widget_slots(node: dict, prefix: str, graph: Graph) -> list[dict]:
         }
         if entry.port.enum_values:
             slot["enum"] = list(entry.port.enum_values)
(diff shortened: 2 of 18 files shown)


skishore23:
@coderabbitai review

skishore23:
@coderabbitai review

skishore23:
@coderabbitai review

annehe9 (approved):

annehe9:
#### Medium, not blocking: Perf: `effective_value` rebuilds the definition index once per promoted widget

`workflow_print.py:884` calls the public `promoted.effective_value(...)`, which re-derives `defs_by_id(workflow)` and then re-resolves the input by name via `find_promoted` → `promoted_inputs(sg, defs)`. Both are already available at the call site: `_State.promoted_defs` caches the index (`workflow_print.py:1048`), and the enclosing loop at `workflow_print.py:874` already holds the `PromotedInput` as `pi`.

Measured on a synthetic workflow of 200 `image_z_image_turbo` instances (8 promoted widgets each), against `object_info_subgraph_promoted.json`:

| | base (`kishore/be-10305-…`) | this PR |
|---|---|---|
| `render_py` wall clock | 41 ms | 251 ms |

cProfile attributes 8.79 s of a 10.4 s profiled render to `effective_value`. Call counts for a 50-instance version of the same workflow:

| function | calls | needed |
|---|---|---|
| `promoted.promoted_inputs` | 451 | 51 |
| `promoted.defs_by_id` | 401 | 1 |

The 400 extra `promoted_inputs` calls are all `find_promoted` re-locating a `pi` the caller already has.

`promoted._effective` already takes `defs`, so this is close to a one-liner — an entry point that skips the name lookup:

```python
def effective_value_for(workflow, instance, sg, pi, graph, defs):
    """`effective_value` when the caller already resolved `pi` and `defs`."""
    value = host_value(instance, pi)
    return value if value is not UNSET else source_value(workflow, sg, pi, graph, defs)
```

and at the call site:

```python
value = _promoted.effective_value_for(state.workflow, node, sg_def, pi, ctx.graph, state.promoted_defs)
```

Note this drops the `ValueError` the current `except` at `workflow_print.py:885` catches. That handler is unreachable anyway: `pi` comes from `promoted_inputs(sg_def, …)` on the same definition the caller resolved, and `_render_nodes` only reaches this path when `is_subgraph_uuid(t)` holds and `defs_by_id[t]` exists, so neither "not a promoted input" nor "not a subgraph instance" can fire.

Not a blocker. Real templates have a handful of instances, so 250 ms is invisible today. Worth doing before anything starts printing generated or batched workflows.

Rest of the PR checks out. I rendered all six gallery fixtures on base vs this head: every case in the description reproduces, no exceptions, no new warnings. The `frontend_injected` fix is a real one — `LoadVideo(file="sloth_bullfight.mp4", control_after_generate="image")` becomes `LoadVideo(file="sloth_bullfight.mp4")`.

skishore23:
Addressed in 99428ac3 — thanks for measuring it.

- `promoted.effective_value_for(workflow, instance, sg, pi, graph, defs)` added (host value if materialized, else `source_value`); `_effective` now delegates to it, so the by-name `effective_value` path is unchanged.
- The instance line uses it with the loop's own `pi` and `_State.promoted_defs`. The `ValueError` handler is gone for the reason you gave: `pi` comes from `promoted_inputs` on the definition `_render_nodes` already resolved, so neither branch could fire.
- Pinned: `test_promoted_values_resolve_once_per_instance_not_per_widget` renders 40 z-image instances with `promoted_inputs` / `defs_by_id` counted and asserts <= N+2 / <= 2 (red on the previous head: 361 / 321). Plus `test_effective_value_for_skips_the_name_lookup` in `tests/comfy_cli/cql/test_promoted_inputs.py` (never re-walks `promoted_inputs`; agrees with `effective_value` for both the interior-default and host-value cases).
- Your 200-instance synthetic: 251 ms -> 29 ms here.

Print/cql/slots/edit suites: 550 passed; ruff 0.15.15 check/format clean.

annehe9 (approved):

skishore23:

PR #817, opened 2026-08-28
feat(knowledge): carry the model row's best_for on each pick

annehe9:
## Overview

`comfy knowledge pick <capability>` now carries each pick's model-row `best_for` (head of the list, one item). The per-pick dict is built by one shared helper, `knowledge.pick_entry`, used by both the CLI command and the envelope-enrichment block, so the two paths emit the same shape.

## Motivation

`knowledge pick` is what the hosted cloud agent's `list_model_picks` tool runs. Per pick it emitted only rank, model, route, template, caveat, status, superseded_by and `fits`. A routing opinion written only in the model row was invisible at recommendation time.

Three eval cases failed on this in run `run_2a4c15d4` (2026-08-28). The flux-2 row says `local multi-reference identity, product or brand consistency (up to 8 refs)` in `best_for[0]`, but its image-edit pick caveat says `fast/cheap everyday edits`, so the agent recommended Seedream for a multi-ref ask. With this change the image-edit table shows:

```
2 flux-2   | FLUX.2 Klein - community No.2 OSS edit model, fast/cheap eve… | best_for: ['local multi-reference identity, product or brand consistency (up to 8 refs, autogrow)']
6 seedream | Multi-reference, brand/character consistency, product edit a… | best_for: ['an unspecified closed-source still - the API photoreal default']
```

## Research findings

**Where the output is built.** The task pointed at `_pick_entries` in `comfy_cli/knowledge.py`, but the CLI command does not call it. `pick_cmd` in `comfy_cli/command/knowledge.py` built its own per-pick dict inline (rank/model/route/template/caveat/status/superseded_by, plus `fits`). `_pick_entries` is the envelope-enrichment path (`knowledge.attach`, the passenger `knowledge` block on `nodes search` etc.), and it adds a `capability` key and skips `fits`. The two also disagreed on `superseded_by`: `_pick_entries` read `bundle.deprecations` first, `pick_cmd` read only the row. Both now go through `pick_entry`.

**Consumers of the pick shape.** `tests/comfy_cli/test_knowledge_attach.py::test_pick_entry_shape` asserts the exact key set on an enrichment pick (updated). `comfy_cli/schemas/knowledge.json` documents the `pick` payload without `additionalProperties: false` (field added). In the cloud repo, `services/agent/internal/loop/cli_fields.json` contracts `data.picks[].{rank,model,route,template,caveat,status,superseded_by}` by presence only, so an added key needs no change there. No docs under `docs/` describe the pick fields.

**Sizing.** Whole `--json` envelope bytes per capability, real bundle (`COMFY_KNOWLEDGE_FILE=cloud/services/agent/knowledge/knowledge.json`), 14 capabilities, 70 picks. Baseline and final rows are `comfy --json knowledge pick -- <cap> | wc -c`; the candidate rows were measured in-process through the same command with the cap patched.

| shape | avg bytes | max bytes | capabilities over 4096 |
|---|---|---|---|
| today | 1849 | 2847 (image-edit) | 0 |
| best_for[:1] **(chosen)** | 2385 | 3657 (image-edit) | 0 |
| best_for[:1] + not_for[:1] | 2764 | 4239 (image-edit) | 1 |
| best_for[:2] | 2820 | 4120 (image-edit) | 1 |
| best_for[:3] | 2964 | 4478 (image-edit) | 2 |
| best_for[:2] + not_for[:2] | 3385 | 5167 (text-to-video) | 5 |
| full best_for + not_for (cap 8) | 3681 | 5695 (text-to-video) | 5 |

Bundle stats: every pick's row has `best_for` (median 2.5 items, max 4); `not_for` median 2, absent on 10 of 70.

**Decision.** `MAX_PICK_BEST_FOR = 1`, key name `best_for`, list of one, same key as `knowledge resolve`. `not_for` is left out: even one item sends image-edit over the cap.

## Changes

- `comfy_cli/knowledge.py`: new `pick_entry(bundle, p)` builds the per-pick dict and adds `best_for` via the same `_str_list` slice `_model_entry` uses. `_pick_entries` prepends `capability` and keeps its `available_locally` annotation. New constant `MAX_PICK_BEST_FOR = 1`.
- `comfy_cli/command/knowledge.py`: `pick_cmd` calls `pick_entry` and adds `fits` as before. Side effect: its `superseded_by` now also reads the bundle's `deprecations` list, matching the enrichment block.
- `comfy_cli/schemas/knowledge.json` and `comfy_cli/schemas/knowledge_block.json`: `picks[].best_for` documented; `knowledge.json`'s `superseded_by` description now matches the block schema (deprecations list or row).
- `tests/comfy_cli/test_knowledge.py`: `test_pick_carries_the_row_best_for_capped` (a 4-item row emits 1, a row without the key emits no key, a non-list value emits no key) and `test_pick_superseded_by_reads_the_deprecations_list`.
- `tests/comfy_cli/test_knowledge_attach.py::test_pick_entry_shape`: key set includes `best_for`, lipco-3 is capped to 1, a pick with no model row has no key.

## Notes

- The cloud agent inlines `list_model_picks` results verbatim (exempt from the spill seam), but the harness admission cap is 4096 bytes and an exempt result over the cap emits an `agent.spill outcome=exempt` span on every call. The cap of 1 is the largest that keeps all 14 capabilities under 4096 with today's bundle (max 3657, 439 bytes of headroom on image-edit). A bundle edit that grows image-edit's caveats or descriptions by that much would need this revisited.
- The enrichment block's own ceiling is `MAX_BLOCK_BYTES = 8192` with `_fit` shedding whole entries, so the larger pick rows there are bounded the same way as before.
- The cloud agent must bump its pinned comfy-cli SHA in `services/agent/Dockerfile` (currently `50d095d`) to pick this up. `cli_fields.json` needs no change; adding `data.picks[].best_for` there is optional.

## Test plan

- [x] `uvx ruff@0.15.15 check .` (CI's pinned version): All checks passed
- [x] `uvx ruff@0.15.15 format --diff .`: 355 files already formatted
- [x] `NO_COLOR=1 uv run pytest tests/comfy_cli/test_knowledge.py tests/comfy_cli/test_knowledge_attach.py tests/comfy_cli/command/test_knowledge_enrichment.py -q`: 263 passed
- [x] `uv run --extra dev pytest .` (before the schema/test follow-ups): 6146 passed, 2 failed, 39 skipped. The 2 failures (`test_build.py::test_from_workflow_refuses_a_workflow_nested_past_the_parser_limit`, `test_pretty_print_sanitize.py::test_status_glyph_keeps_its_own_tags_live`) fail identically on a clean `origin/main` checkout in this environment (signed-in state and terminal styling), unrelated to this diff.
- [x] After the follow-ups: knowledge, enrichment and schema-consuming test files (`test_knowledge*.py`, `test_knowledge_enrichment.py`, `command/generate`, `test_nodes_introspect.py`, `test_templates_get.py`): 652 passed
- [x] Real-command sizing above: 14/14 capabilities under 4096 bytes

https://claude.ai/code/session_0153s9cR6S73y6m2XYxkv9Vb

### DIFF
--- a/comfy_cli/command/knowledge.py
+++ b/comfy_cli/command/knowledge.py
@@ -206,11 +206,12 @@ def pick_cmd(
     if renderer.is_pretty():
         from rich.table import Table
 
-        columns = ("rank", "model", "route", "template", "status", "caveat")
+        columns = ("rank", "model", "route", "template", "status", "caveat", "best_for")
         tbl = Table(show_header=True, header_style="bold")
         for col in columns:
             tbl.add_column(col)
         for p in picks:
-            tbl.add_row(*(sanitize_markup("" if p[c] is None else p[c]) for c in columns))
+            cells = {**p, "best_for": ", ".join(p.get("best_for") or [])}
+            tbl.add_row(*(sanitize_markup("" if cells[c] is None else cells[c]) for c in columns))
         renderer.console().print(tbl)
     renderer.emit(payload, command="knowledge pick")
--- a/comfy_cli/command/knowledge.py
+++ b/comfy_cli/command/knowledge.py
@@ -190,18 +190,7 @@ def pick_cmd(
     knowledge.log_query("knowledge pick", logged, hit_ids=[f"cap:{capability_id}"], zero_hit=False, bundle=bundle)
     picks = []
     for p in cap["picks"]:
-        model_id = p.get("model")
-        model_id = model_id if isinstance(model_id, str) else None
-        row = bundle.models.get(model_id) or {}
-        entry = {
-            "rank": knowledge.pick_rank(p),
-            "model": model_id,
-            "route": _text(p.get("route")),
-            "template": _text(p.get("template")),
-            "caveat": _text(p.get("caveat")),
-            "status": _text(row.get("status")),
-            "superseded_by": _text(row.get("superseded_by")),
-        }
+        entry = knowledge.pick_entry(bundle, p)
         if "fits" in p:
             entry["fits"] = p["fits"]
         picks.append(entry)
--- a/comfy_cli/knowledge.py
+++ b/comfy_cli/knowledge.py
@@ -47,6 +47,10 @@ MAX_MODELS = 3
 MAX_MODELS_BRIEF = 20
 MAX_PICKS = 8
 MAX_LIST_ITEMS = 8
+# best_for items per pick. 1 keeps every capability's `knowledge pick` envelope
+# under the 4096 bytes the cloud agent admits verbatim; 2, or not_for alongside,
+# sends image-edit over.
+MAX_PICK_BEST_FOR = 1
 MAX_BLOCK_BYTES = 8192
 MAX_QUERY_CHARS = 200  # CLI text is unbounded; the clip bounds the lookup key and the nudge echo
 MAX_VERSION_CHARS = 64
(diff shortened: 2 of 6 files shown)


skishore23 (approved):
Clean refactor. Verified `_text()`/`_str_list()` preserve the old `isinstance` semantics for malformed `model`/`template` values, and the one behavior change (`pick_cmd` now reads `bundle.deprecations` first for `superseded_by`, matching the enrichment block) is intentional and tested. Ran `test_knowledge`, `test_knowledge_attach`, `test_knowledge_enrichment` on the branch: 264 passed.

Issue #104, opened 2024-07-08
Add mobile support


Mike-FreeAI:
**Describe the solution you'd like**
Please add mobile support

**Describe alternatives you've considered**
Using A1111

@DevApeAI go

Mike-FreeAI:
@DevApeAI

robinjhuang:
@Mike-FreeAI Do you need mobile support for ComfyUI? Or the cli specifically?

yoland68:
This is a command line tool, there is no model support for this

ltdrdata:
It seems you have likely found this repo after seeing Comfy-Org.
This is a sub-project repo of ComfyUI, and any UX-related issues for ComfyUI should still be requested in the [ComfyUI repo](https://github.com/comfyanonymous/ComfyUI).
And also check out [mcmonkeyprojects/SwarmUI](https://github.com/mcmonkeyprojects/SwarmUI), which uses ComfyUI as its backend.

PR #818, opened 2026-08-28
fix(ops): gate promoted connect grows as one register; positional payload for PrimitiveNode writes (amendment v1.5)

skishore23:
> **Stacked on #816** → #815 → #812 → #809. Pairs with comfy-multi-player **#104** (applier side).

## Why

comfy-multi-player #104 implemented the host-write and promoted-connect op shapes #815 introduced (so the in-app doc host can apply them), and its convergence review sent two asks back to this repo:

1. **Promoted `connect` grows were not gated.** A promoted subgraph input is ONE register named by the definition, but `_apply_connect` kept the autogrow "every grow mints its own slot" carve-out: two concurrent connects onto `57.width` left the entry with whichever link arrived first, and `grow_id` stayed on the first arrival — two replicas could disagree. Now gated by `_lww_gate`/`_lww_commit` on `("input", to_node, "grow", <full name>)` exactly like a concrete input (§11.1): higher stamp wins in either order, `grow_id` follows the winner, the loser's link is retired. The register uses the **full** declared name — sg inputs such as `images.image0` contain a dot, and the old base-split would alias two inputs onto one register.
2. **`PrimitiveNode` writes had no positional payload.** A frontend-only `PrimitiveNode` has no catalog entry, so an opaque store rejected the `legacy_primitive` op. It now carries the same `promoted` payload a host write carries (`value_index: 0`, `instance_path`, `host_widgets_values`); apply treats a `promoted` payload on a non-instance node as a plain positional write.

## Red → green

`test_concurrent_promoted_connects_converge_by_stamp` (both apply orders → one `width` entry, winner's link + `grow_id`, loser's link gone; failed red with the arrival-order winner) and `test_legacy_primitive_write_carries_a_positional_payload` (exact payload; replay on a fresh copy applies it; failed red with `KeyError: 'promoted'`).

## Contract

`docs/op-vocabulary-v1.md` **Amendment v1.5** — records the `promoted` host-write payload, `redirected_from`, `grow.promoted` (introduced by #815 without an amendment), the promoted-grow register and gating, and the opaque positional write. Host-write vs interior-`path` registers stay deliberately un-unified (needs the promotion table at apply time), matching multi-player A15.

## Verification

Full suite **6261 passed, 38 skipped**; ruff 0.15.15 check + format clean; `test_op_vocabulary_contract.py` (doc == constants == dispatch) green.

### DIFF
--- a/comfy_cli/workflow_ops.py
+++ b/comfy_cli/workflow_ops.py
@@ -1944,8 +1944,14 @@ def _apply_connect(workflow: dict, op: dict, graph) -> None:
     if grow is not None:
         # Autogrow is NOT a shared register: every grow mints its own slot keyed
         # by ``grow_id``, so two concurrent grows onto one base both survive and
-        # there is nothing to gate (§1.2 / amendment v1.2's carve-out).
-        if _find_by_str(workflow, op["from_node"]) is None:
+        # there is nothing to gate (§1.2 / amendment v1.2's carve-out) — a
+        # deleted source simply means nothing to wire. A PROMOTED grow IS a
+        # register (§14.2), so it must reach the gate below even when its
+        # source is gone: bailing here would make the incumbent depend on
+        # whether the concurrent delete of this op's source had arrived yet.
+        # Delete wins over the LINK, not over the claim — the entry is left
+        # empty at the ``src is None`` check further down.
+        if not grow.get("promoted") and _find_by_str(workflow, op["from_node"]) is None:
             return
         # Autogrow: grow a concrete slot and wire it. Keyed by ``grow_id`` (the
         # link id) so replay is idempotent AND non-clobbering — a concurrent
@@ -2600,7 +2606,7 @@ def _resolve_input_target(
     # (a concrete ``to_slot`` op would find no slot, be dropped, and never
     # replay). Apply reuses the entry by name; type-checked against the
     # declared input type.
-    if workflow is not None and isinstance(slot, str):
+    if workflow is not None and isinstance(slot, (str, int)) and not isinstance(slot, bool):
         promoted_grow = _resolve_promoted_target(workflow, node, slot, elem_type)
         if promoted_grow is not None:
             return None, promoted_grow
@@ -2710,10 +2716,14 @@ def _resolve_input_target(
     raise ValueError(f"input {slot!r} not found on node {node.get('id')}; inputs: {names}")
 
 
-def _resolve_promoted_target(workflow: dict, node: dict, slot: str, elem_type: str | None) -> dict | None:
+def _resolve_promoted_target(workflow: dict, node: dict, slot: Any, elem_type: str | None) -> dict | None:
     """The input to materialize on subgraph instance ``node`` for a connect to
     its declared input ``slot`` (see ``cql.promoted``). ``None`` when ``node``
-    is not an instance or ``slot`` is not one of its definition's inputs."""
+    is not an instance or ``slot`` is not one of its definition's inputs.
+
+    An INDEX (``57.1``) that lands on an already-materialized entry for a
+    declared input is the same input as its name (``57.width``): it maps onto
+    the name so both addresses share one register (§14.2)."""
     from comfy_cli.cql import engine as _engine
     from comfy_cli.cql import promoted as _promoted
 
@@ -2721,6 +2731,13 @@ def _resolve_promoted_target(workflow: dict, node: dict, slot: str, elem_type: s
     sg = defs.get(str(node.get("type", "")))
     if sg is None:
         return None
+    if isinstance(slot, int) or (isinstance(slot, str) and slot.lstrip("-").isdigit()):
+        ins = node.get("inputs") or []
+        idx = int(slot)
+        entry = ins[idx] if 0 <= idx < len(ins) and isinstance(ins[idx], dict) else None
+        if entry is None or not entry.get("name"):
+            return None
+        slot = str(entry["name"])
     pi = _promoted.find_promoted(sg, defs, slot)
     if pi is None:
         return None
--- a/comfy_cli/workflow_ops.py
+++ b/comfy_cli/workflow_ops.py
@@ -1883,11 +1883,13 @@ def _apply_positional_write(node: dict, positional: dict, value: Any) -> None:
     values = node.get("widgets_values")
     values = list(values) if isinstance(values, list) else []
     seed = positional.get("host_widgets_values")
-    if len(values) <= idx:
-        if isinstance(seed, list) and len(seed) > len(values):
-            values.extend(seed[len(values) :])
-        while len(values) <= idx:
-            values.append(None)
+    # A receiver holding a truncated array takes the payload's tail whether or
+    # not the written index is inside it — otherwise replicas keep different
+    # opaque state for the same node.
+    if isinstance(seed, list) and len(seed) > len(values):
+        values.extend(seed[len(values) :])
+    while len(values) <= idx:
+        values.append(None)
     values[idx] = value
     node["widgets_values"] = values
 
@@ -2589,6 +2591,19 @@ def _resolve_input_target(
       the API converter reads the link and skips the widget by name.
     """
     ins = node.get("inputs") or []
+    # Promoted subgraph input (``57.width``): the definition declares it, so it
+    # is a link input on the instance even before the instance carries an
+    # ``inputs[]`` entry for it — the frontend rebuilds those from the
+    # definition on load. ALWAYS addressed as a promoted grow, even once the
+    # entry exists: the register is the declared name, so a replica that
+    # receives a later connect before the materializing one still lands it
+    # (a concrete ``to_slot`` op would find no slot, be dropped, and never
+    # replay). Apply reuses the entry by name; type-checked against the
+    # declared input type.
+    if workflow is not None and isinstance(slot, str):
+        promoted_grow = _resolve_promoted_target(workflow, node, slot, elem_type)
+        if promoted_grow is not None:
+            return None, promoted_grow
     # Concrete slot (index or exact name) that is NOT an autogrow base.
     try:
         idx = _resolve_input_slot(node, None, slot)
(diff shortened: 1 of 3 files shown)


skishore23:
@coderabbitai review

comfy_cli/workflow_ops.py
```diff
@@ -1859,6 +1866,26 @@ def _apply_set_widget(workflow: dict, op: dict, graph) -> None:
...
+    if not isinstance(idx, int) or isinstance(idx, bool) or idx < 0:
+        return
+    values = node.get("widgets_values")
+    values = list(values) if isinstance(values, list) else []
+    seed = positional.get("host_widgets_values")
+    if len(values) <= idx:
+        if isinstance(seed, list) and len(seed) > len(values):
+            values.extend(seed[len(values) :])
+        while len(values) <= idx:
+            values.append(None)
```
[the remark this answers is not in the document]
skishore23: Fixed in e6d462e3: _apply_positional_write extends from host_widgets_values whenever the payload is longer, before the index check; pinned by test_positional_replay_extends_a_partially_present_array_from_the_payload.

comfy_cli/workflow_ops.py
```diff
@@ -1921,10 +1948,23 @@ def _apply_connect(workflow: dict, op: dict, graph) -> None:
...
+            if to_idx is None and not _lww_gate(workflow, op):
+                return
+            _lww_commit(workflow, op)
+            if to_idx is None:
+                to_idx = next((k for k, i in enumerate(ins) if i.get("name") == grow["name"]), None)
+            if to_idx is not None:
+                prev = ins[to_idx].get("link")
+                if prev is not None and prev != op["link_id"]:
+                    _remove_link(workflow, prev)
+                ins[to_idx]["grow_id"] = op["link_id"]  # the register follows the winner
```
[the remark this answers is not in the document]
skishore23: Fixed in e6d462e3: every connect to a declared promoted input is now the promoted grow (apply reuses the entry by name), so the register stays the declared name; pinned by test_later_connect_to_a_materialized_promoted_input_stays_on_the_declared_register (both replay orders). §14.2 rewritten.

docs/op-vocabulary-v1.md
```diff
@@ -723,3 +723,71 @@ rebuilds the executable graph — the API prompt — not the canvas decoration.
...
+stores its widgets opaquely (positional) and these ops carry what an opaque
+store needs.
+
+### 14.1 `set_widget` host write: the `promoted` payload
+
+A write to a promoted input is a top-level `set_widget` on the INSTANCE
+(`node_id`, `widget` = the declared input name; no `path`/`inner_widget`)
+carrying
+
+```
```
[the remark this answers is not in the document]
skishore23: Fixed in e6d462e3: both fences are now ```json.

skishore23:
@coderabbitai review

skishore23:
@coderabbitai review

comfy_cli/workflow_ops.py
```diff
@@ -1925,10 +1954,23 @@ def _apply_connect(workflow: dict, op: dict, graph) -> None:
         # schema's own element names, when the catalog carries a template).
         ins = dst.setdefault("inputs", [])
         to_idx = next((k for k, i in enumerate(ins) if i.get("grow_id") == op["link_id"]), None)
-        if to_idx is None and grow.get("promoted"):
-            # A promoted subgraph input is ONE register named by the definition:
-            # a concurrent connect that already materialized it shares the entry.
-            to_idx = next((k for k, i in enumerate(ins) if i.get("name") == grow["name"]), None)
+        if grow.get("promoted"):
```
annehe9, line 1965: **Blocking.** A promoted connect whose source node was concurrently deleted returns before this gate ever runs, so it never claims the register.

Twelve lines up, the grow branch still opens with the pre-v1.5 bail:

```python
grow = op.get("grow")
if grow is not None:
    # Autogrow is NOT a shared register: ... there is nothing to gate
    if _find_by_str(workflow, op["from_node"]) is None:
        return
```

That was correct while autogrow had no register. A promoted grow has one now, so this is exactly the order dependence the concrete branch warns about next to its own `_lww_commit`: *"whether the incumbent survives would depend on whether the concurrent delete of THIS op's source had arrived yet."* Amendment 14.2 also states the claim is unconditional once the gate passes, and this path never reaches the gate.

Repro: two promoted connects into `57.width` (stamps lo and hi) plus a `delete_node` of the hi op's source, applied in all 6 permutations.

- delete applied last: one `width` entry, `link=None`, `grow_id=<hi>`
- delete applied first: one `width` entry, `link=<lo>`, `grow_id=<lo>` — the hi op is dropped whole

`canonical()` gives **2 distinct states across the 6 orders**. The same three-op interleave on a concrete input (`SaveImage.images`) converges to a single state in all 6, so this is specific to the new promoted branch rather than an accepted limitation.

The one-line change that made it converge for me:

```python
if not grow.get("promoted") and _find_by_str(workflow, op["from_node"]) is None:
    return
```

The promoted op then gates, commits, retires the loser's link, appends the entry with `link=None`, and returns at the existing `src is None` check further down — landing on the delete-applied-last state in all 6 orders. All 22 tests in `test_workflow_edit_promoted.py` stay green, and the 553 tests that import `workflow_ops` show no new failures.

Minor, while you are in here: the comment above that bail still reads "Autogrow is NOT a shared register ... there is nothing to gate", which the block below now contradicts.

annehe9 (commented):
Reviewed by running the code on `fa5f876`, not just reading it. One blocking finding, one behaviour split I think is unintended, one non-blocking note. Details inline.

The `PrimitiveNode` half of the PR looks right to me — `trace_upstream_write` resolves the primitive through `_find_top`, so the single-segment `instance_path` always holds, and `apply_op` mutating in place makes the post-write `host_widgets_values` read correct.

Note on verification: `test_ack_flag.py` and `test_login_command.py` fail for me on the unpatched head too. Terminal width / ANSI in my shell, pre-existing and unrelated to this PR.

comfy_cli/workflow_ops.py
```diff
@@ -2526,6 +2572,19 @@ def _resolve_input_target(
...
+    # is a link input on the instance even before the instance carries an
+    # ``inputs[]`` entry for it — the frontend rebuilds those from the
+    # definition on load. ALWAYS addressed as a promoted grow, even once the
+    # entry exists: the register is the declared name, so a replica that
+    # receives a later connect before the materializing one still lands it
+    # (a concrete ``to_slot`` op would find no slot, be dropped, and never
+    # replay). Apply reuses the entry by name; type-checked against the
+    # declared input type.
+    if workflow is not None and isinstance(slot, str):
+        promoted_grow = _resolve_promoted_target(workflow, node, slot, elem_type)
```
annehe9, line 2610: Moving the promoted resolution above the concrete-slot resolution splits one register into two. `57.width` and `57.1` address the same materialized input, but they no longer gate against each other:

```
base branch:  57.width -> ('input','57',1)                57.1 -> ('input','57',1)
this branch:  57.width -> ('input','57','grow','width')   57.1 -> ('input','57',1)
```

Both forms are reachable from the CLI: `_split_addr` passes whatever follows the dot straight through, and `_resolve_input_slot` accepts a numeric string as an index. Two people connecting to the same input by different addresses now clobber by arrival order.

Amendment 14.2 says "every connect to a declared promoted input is this promoted grow", which is not true for the index form. Either map an index that lands on a declared promoted input onto the name register at mint time, or scope index addressing out of 14.2 explicitly.

Related, and worth a line in the amendment either way: an op minted by a pre-v1.5 replica as a concrete `to_slot` never gates against a v1.5 promoted grow on the same input.

comfy_cli/workflow_ops.py
```diff
@@ -1863,6 +1870,28 @@ def _apply_set_widget(workflow: dict, op: dict, graph) -> None:
...
+    if not isinstance(idx, int) or isinstance(idx, bool) or idx < 0:
+        return
+    values = node.get("widgets_values")
+    values = list(values) if isinstance(values, list) else []
+    seed = positional.get("host_widgets_values")
+    # A receiver holding a truncated array takes the payload's tail whether or
+    # not the written index is inside it — otherwise replicas keep different
+    # opaque state for the same node.
+    if isinstance(seed, list) and len(seed) > len(values):
+        values.extend(seed[len(values) :])
```
annehe9, line 1890: Non-blocking, but the tail copied here is not gated. Only `value_index` goes through `_lww_gate`; the rest of `host_widgets_values` rides along with it, so two concurrent writes carrying different tails settle by apply order:

```
apply hi then lo -> [512, 'fixed']
apply lo then hi -> [512, 'randomize']
```

This is still a net improvement over the previous behaviour, which left a truncated replica permanently truncated, so I would not block on it. But 14.1's "extending a shorter stored array from the payload" reads as a converged value — suggest saying the tail is best-effort repair.

comfy_cli/workflow_ops.py
skishore23: Fixed in 5159ebd8 — exactly your one-liner (the bail now applies to autogrow only), plus the stale comment. Pinned by test_promoted_connect_claims_its_register_even_when_its_source_was_deleted: lo/hi promoted connects + delete of hi's source, all 6 permutations → one canonical state (one width entry, link=None, grow_id=hi).

comfy_cli/workflow_ops.py
skishore23: Fixed in 5159ebd8: an index that lands on a materialized entry for a declared promoted input maps onto the name at mint time, so 57.1 and 57.width share ("input","57","grow","width"); pinned by test_index_address_of_a_materialized_promoted_input_shares_the_name_register (both apply orders). §14.2 now states the index mapping and scopes pre-v1.5 concrete ops out explicitly.

comfy_cli/workflow_ops.py
skishore23: Agreed — §14.1 now says the tail extension is best-effort repair (only the written index is under the register; concurrent different tails settle by apply order) in 5159ebd8.

annehe9 (approved):

skishore23:

PR #819, opened 2026-08-28
fix(subgraph): repair legacy proxyWidgets promotions before writing them, the way the frontend does on load (BE-10305)

skishore23:
> **Stacked on #818** → #816 → #815 → #812 → #809. Closes the "legacy proxies not backed by a linked input still route to the interior" gap left open in #815.

## Why

364 of the 1,922 `proxyWidgets` entries in the gallery corpus (old templates) are not backed by a linked subgraph input. For those, #815 kept the legacy behaviour — a write into the interior node, an edit under "layer 2" that the frontend's load-time migration happens to re-seed the host from, but that no other promoted write shares and that `comfy run` never submitted as a host value.

## What

A faithful port of the frontend's forward migration (`proxyWidgetMigration.ts`, ADR 0009) into `cql.promoted`, run on write:

`cql.promoted` now carries a port of the frontend's forward migration
(`proxyWidgetMigration.ts`, ADR 0009): `plan_proxy_migration` classifies
every entry of an instance exactly as `classify` does (already linked;
`createSubgraphInput`; `PrimitiveNode` fan-out bypass, all-or-quarantine;
`$$`/preview pseudo-widget exposure; quarantine with the frontend's reason
codes), and `flush_proxy_migration` performs it: a subgraph input named with
`nextUniqueName`, a boundary link from the subgraph input node into the
widget's backing input slot (synthesized in the shape the frontend
serializes when an older save omits it), the host value read positionally
by legacy proxy order, quarantine rows `{originalEntry, reason, hostValue?,
attemptedAtVersion: 1}`, repaired entries consumed. Preview entries are left
in `proxyWidgets` so the frontend's own preview-exposure migration (and its
auto-exposure of preview nodes) produces the state it would have produced
from the untouched file.

`resolve_write` turns such an entry — addressed flat, by its legacy widget
name, or through the interior widget/primitive target it owns — into a HOST
target carrying the repair; the op records `promoted.repair = {entry, ids}`
with input/link ids derived by SHA-256 from (instance path, source node,
widget), never random, so replay on another replica is byte-identical and
concurrent repairs of one instance converge. A shared definition is forked
first, like an interior write. Entries the migration quarantines
(`control_after_generate` has no backing slot) keep the interior widget as
the live one, so those writes land there as before.

Reads present a pending repair where the frontend will show it:
`slots`/`effective_value` advertise `<instance>.<name>` with the legacy host
value or the interior source, and hide the interior address; reads never
mutate. The converter now overlays a host value onto EVERY boundary-link
target, which a repaired primitive fan-out needs.

Fixtures: three verbatim gallery templates (primitive fan-outs under user
titles, `$$canvas-image-preview` exposures next to a value widget with no
serialized input slot, and the common `seed` + `control_after_generate`
pair) with a trimmed cloud catalog. `TestSetWidgetSubgraph` is rewritten for
the migration semantics on a realistic legacy fixture, with the old
declared-but-unlinked shape pinned separately (`nextUniqueName` mints
`text_1` beside the dangling socket).

## Verification

Targeted subgraph/edit/converter suites on the rebased head: 635 passed; full suite on the stack head (#820) is reported there. ruff 0.15.15 check + format clean. Contract: §8.7 pinned text + Amendment v1.5 §14.4.

## Known follow-ups (pre-existing, not introduced here)

- A host write on an instance whose `proxyWidgets` entries are ALL already linked does not flush the legacy list, so a linked-order `widgets_values` sits beside a proxy-order-indexed legacy list. Harmless when the orders coincide — they do in every gallery template — but a hazard in principle; flushing on any host write would close it.
- Replaying a repair op requires the catalog (`graph=None` raises clearly rather than quarantining everything).

Verification on this head (`27214b8a`): full suite 6301 passed, 38 skipped.

### DIFF
--- a/comfy_cli/cql/promoted.py
+++ b/comfy_cli/cql/promoted.py
@@ -1352,15 +1352,27 @@ def flush_proxy_migration(
         source = _inner_node(sg, e.source_node)
         slot_index = e.slot_index
         if slot_index is None:
-            source.setdefault("inputs", []).append(
-                {
+            if e.source_input is not None:
+                # A nested instance backs the widget with its OWN host input
+                # for the projecting subgraph input (``getSlotFromWidget``
+                # over ``promotedInputWidget``): the slot carries that input's
+                # name, or the outer definition could never resolve the
+                # promotion through the inner one.
+                entry = {
+                    "name": e.source_input,
+                    "type": e.type,
+                    "widget": {"name": e.source_input},
+                    "link": None,
+                }
+            else:
+                entry = {
                     "localized_name": e.widget,
                     "name": e.widget,
                     "type": e.type,
                     "widget": {"name": e.widget},
                     "link": None,
                 }
-            )
+            source.setdefault("inputs", []).append(entry)
             slot_index = len(source["inputs"]) - 1
         new_input = _add_subgraph_input(sg, ids[e.key]["input"], str(e.name), str(e.type), e.label)
         _add_boundary_link(sg, new_input, ids[e.key]["links"][0], source, slot_index, str(e.type))
@@ -1382,9 +1394,12 @@ def flush_proxy_migration(
             quarantine.extend(_quarantine_entry(m, "primitiveBypassFailed") for m in members)
             continue
         targets = _primitive_targets(sg, primitive)
-        for link in [x for x in sg.get("links") or [] if str(x.get("origin_id")) == primitive_id]:
-            if link.get("origin_slot") == 0:
-                _remove_link(sg, link.get("id"))
+        for link in [
+            x
+            for x in sg.get("links") or []
+            if isinstance(x, dict) and str(x.get("origin_id")) == primitive_id and x.get("origin_slot") == 0
+        ]:
+            _remove_link(sg, link.get("id"))
         link_ids = ids[first.key]["links"]
         if len(link_ids) < len(targets):
             link_ids = repair_ids(instance_path, first.source_node, first.widget, len(targets))["links"]
--- a/comfy_cli/cql/promoted.py
+++ b/comfy_cli/cql/promoted.py
@@ -17,17 +17,24 @@ linked inputs"*, ``SubgraphNode.ts``) represents a promoted widget as a
 The interior widget is only a schema/default provider — *"the host/exterior
 value wins over the interior/source value during repair, persistence, and
 prompt serialization"*. ``properties.proxyWidgets`` is legacy load-time input
-the frontend consumes on migration; it is honored here only as a fallback for
-a name the definition does not declare as an input.
+the frontend consumes on migration. A legacy entry the definition does not
+back with a linked input is repaired the way the frontend repairs it on load
+(:func:`flush_proxy_migration`, the "forward migration" section below) before
+a write lands on it; an entry the migration cannot repair is quarantined and
+its interior widget stays the live one.
 
-Everything in this module is a pure function over workflow JSON. It never
-mutates a subgraph definition: a promoted value is edited on the instance,
-so sibling instances of a shared definition stay independent by construction.
+Everything in this module is a pure function over workflow JSON. A promoted
+value is edited on the instance — the definition is touched only by the
+forward migration, which the op layer runs on an isolated (forked) copy when
+the definition is shared, so sibling instances stay independent.
 """
 
 from __future__ import annotations
 
-from dataclasses import dataclass
+import hashlib as _hashlib
+import re as _re
+import uuid as _uuid
+from dataclasses import dataclass, field
 from typing import Any
 
 #: LiteGraph's id for the virtual subgraph *input* node inside a definition.
@@ -290,6 +297,11 @@ def effective_value(workflow: dict, instance: dict, name: str, graph) -> Any:
     when nothing outside feeds it: the host value if materialized, else the
     interior source value (:data:`UNSET` if even that cannot be read).
 
+    A legacy ``proxyWidgets`` promotion the definition does not back yet
+    reads as the frontend will show it after its own migration: the legacy
+    positional host value, else the interior source (the address is the
+    input name the repair mints — see :func:`plan_proxy_migration`).
+
     Raises ``ValueError`` for a name the definition does not declare, or one
     that is a socket-only input.
     """
@@ -297,6 +309,13 @@ def effective_value(workflow: dict, instance: dict, name: str, graph) -> Any:
     sg = defs.get(str(instance.get("type", "")))
     if sg is None:
         raise ValueError(f"node {instance.get('id')} is not a subgraph instance")
+    pi = find_promoted(sg, defs, name)
+    if pi is None or not pi.is_widget:
+        pending = next(
+            (e for e in plan_proxy_migration(workflow, instance, graph, defs) if e.repairable and e.name == name), None
+        )
+        if pending is not None:
+            return entry_effective_value(workflow, sg, pending, graph, defs)
     return _effective(workflow, instance, sg, name, graph, defs)
 
 
@@ -396,6 +415,11 @@ class WriteTarget:
     widget: str | None = None
     segments: list[str] | None = None
     redirected_from: str | None = None
+    #: A ``host`` target whose promotion is still a legacy ``proxyWidgets``
+    #: entry: the write must first run :func:`flush_proxy_migration` on the
+    #: instance. This is the entry it repairs; ``widget`` is the input name
+    #: the repair mints.
+    repair: Any = None
 
 
 def _find_top(workflow: dict, node_id: Any) -> dict | None:
@@ -442,8 +466,14 @@ def resolve_write(
       there (``redirected_from`` records the given address). An interior
       widget fed by another interior node is refused the same way.
     * any other interior widget — written in the definition, as before.
-    * a promoted name the definition does not declare as an input but the
-      legacy ``proxyWidgets`` still route — the interior, as before.
+    * a name only the legacy ``proxyWidgets`` route — a promotion the
+      frontend's load-time migration turns into a linked input. The host,
+      after the same repair (``repair`` names the entry; see
+      :func:`flush_proxy_migration`). An interior address the repair would
+      own (the entry's source widget, a primitive's fan-out target) redirects
+      there too. An entry the migration cannot repair (quarantined — e.g.
+      ``control_after_generate``, which has no backing input slot) keeps the
+      interior widget as the live one, so the write lands there as before.
     """
     defs = defs_by_id(workflow)
     if len(segments) > 1:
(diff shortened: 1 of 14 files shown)


skishore23:
@coderabbitai review

comfy_cli/cql/promoted.py
```diff
@@ -542,3 +594,863 @@ def trace_upstream_write(
...
+            source.setdefault("inputs", []).append(
+                {
+                    "localized_name": e.widget,
+                    "name": e.widget,
+                    "type": e.type,
+                    "widget": {"name": e.widget},
+                    "link": None,
+                }
+            )
+            slot_index = len(source["inputs"]) - 1
```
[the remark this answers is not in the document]
skishore23: Fixed in 2edaec36. When the source is a nested instance with no serialized slot, the synthesized entry is now named after the projecting input (`e.source_input`) — `{name, type, widget:{name}, link}`, the shape a subgraph instance serializes — so the outer definition resolves the promotion through the inner one. Pinned by `test_flush_synthesizes_a_nested_instance_slot_under_its_projecting_input` on a fixture where the projecting input (`prompt_text`) and the interior widget (`text`) differ; it was red before the fix (slot named `text`, promotion unresolved).

comfy_cli/cql/promoted.py
```diff
@@ -542,3 +594,863 @@ def trace_upstream_write(
...
+        # ``repairPrimitive`` re-validates against the graph as it is NOW: a
+        # ``createSubgraphInput`` above may have taken over one of the
+        # primitive's targets, so the fan-out is collected again.
+        if _validate_primitive_cohort(sg, primitive, members) is not None:
+            quarantine.extend(_quarantine_entry(m, "primitiveBypassFailed") for m in members)
+            continue
+        targets = _primitive_targets(sg, primitive)
+        for link in [x for x in sg.get("links") or [] if str(x.get("origin_id")) == primitive_id]:
+            if link.get("origin_slot") == 0:
+                _remove_link(sg, link.get("id"))
```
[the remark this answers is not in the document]
skishore23: Fixed in 2edaec36: the primitive-bypass link scan now screens `isinstance(x, dict)` like the other scans. `test_flush_ignores_a_malformed_link_entry` inserts a stray list into `links` and asserts the flush result is identical to the clean one (it raised `AttributeError` before).

tests/comfy_cli/cql/test_proxy_migration.py
```diff
@@ -0,0 +1,407 @@
...
+    _inner(sg, 156)["outputs"][0]["links"] = []
+    assert _plans(wf, 143, graph)[("156", "value")].reason == "unlinkedSourceWidget"
+
+
+def test_plan_uses_a_unique_name_when_the_widget_name_is_taken(graph):
+    wf = _load(FLUX)
+    sg = _def_of(wf, _instance(wf, 56))
+    sg["inputs"].append({"id": "x", "name": "seed", "type": "INT", "linkIds": []})
+    assert _plans(wf, 56, graph)[("52", "seed")].name == "seed_1"
+
```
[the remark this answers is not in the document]
skishore23: Added in 2edaec36: `test_plan_normalizes_a_node_prefixed_widget_name` (`['52','52:seed']` → widget `seed`, disambiguator `52`, input `seed`) plus `test_plan_resolves_a_nested_instance_source_by_disambiguator`, which exercises the disambiguator branch of `_resolve_source_widget` through a nested instance (`27:text` resolves; `99:text` quarantines as `missingSourceWidget`).

skishore23:
@coderabbitai review

skishore23:
@coderabbitai review

comfy_cli/cql/promoted.py
```diff
@@ -544,3 +596,878 @@ def trace_upstream_write(
...
+            continue
+        widget_name, projecting = resolved
+        if widget.startswith("$$") or _is_preview_pseudo_widget(widget_name):
+            e.plan = PLAN_PREVIEW
+            continue
+        slot = _slot_for_widget(source, widget_name, projecting, defs, graph)
+        if slot is None:
+            e.reason = "missingSubgraphInput"  # the widget has no backing input slot
+            continue
+        idx, entry, slot_type = slot
```
annehe9, line 1076: **A legacy entry whose backing slot is already fed by an interior link is repaired anyway, and the repair deletes that link.**

Nothing between `_slot_for_widget` returning and `PLAN_CREATE` being set looks at `entry["link"]`. The flush then reaches `_add_boundary_link`, which calls `_remove_link(sg, existing)` and takes the slot over. The old link disappears from `sg["links"]` and from the driver's `outputs[0].links`, silently.

What makes me think this is unintended rather than a faithful port: `resolve_write` refuses this exact shape when no legacy entry exists, with the message *"that link supplies the value, so a widget write there would be ignored; edit that node's own widgets instead"*. But the `planned_repair_for_interior` block sits above that check and returns a HOST target first, so the refusal never runs. Same node, same widget, same link, two opposite behaviours depending only on whether a `proxyWidgets` entry names it.

Reproduced on `templates_graphic_design_recomposer.json` by wiring interior node 2 into `135.choice` before the write:

```
no-legacy-entry:   refused -> 143/135.choice is fed by interior node 143/2 (PrimitiveStringMultiline) - that link supplies the value...
with-legacy-entry: wrote. old link 999001 survives? False | node2 outputs[0].links: [211, 247]
```

The cheapest root-cause fix is one guard here, since it covers the flat address, the interior address and the primitive fan-out targets at once:

```python
idx, entry, slot_type = slot
if isinstance(entry, dict) and entry.get("link") is not None:
    link = next((x for x in sg.get("links") or [] if isinstance(x, dict) and x.get("id") == entry["link"]), None)
    if link is not None and link.get("origin_id") != SUBGRAPH_INPUT_NODE_ID:
        e.reason = "sourceWidgetLinked"
        continue
```

Worth a fixture in `test_proxy_migration.py` too - no test covers a link-fed backing slot today.

If the frontend really does replace the link here, please say so and I will drop this, but then the asymmetry with `resolve_write` still deserves a comment.

comfy_cli/cql/promoted.py
```diff
@@ -544,3 +596,878 @@ def trace_upstream_write(
...
+        for idx, entry in enumerate(entries):
+            if isinstance(entry, dict) and entry.get("name") == name:
+                return idx, entry, str(entry.get("type") or "*")
+        pi = find_promoted(inner_def, defs, name)
+        return (None, None, pi.type) if pi is not None else None
+    for idx, entry in enumerate(entries):
+        marker = entry.get("widget") if isinstance(entry, dict) else None
+        if isinstance(marker, dict) and marker.get("name") == widget:
+            return idx, entry, str(entry.get("type") or "*")
+    m = graph.node(str(source.get("type", ""))) if graph is not None else None
```
annehe9, line 981: **An interior input slot serialized by name only (no `widget` marker) gets duplicated.**

The loop above matches an ordinary node's slot only on `marker["name"] == widget`. A slot carrying `name` but no `widget` marker falls through to this catalog branch and returns `(None, None, port.type)`, so the flush treats it as "not serialized" and appends a second entry with the same name. The original is left behind with `link: null`.

On the recomposer fixture, with the marker stripped from node 135's `choice` input:

```
135 inputs after: [('choice', None), ('choice', 3935858563152590)]
```

Older saves are exactly the population this PR targets, so it seems worth falling back to an `entry.get("name") == widget` match before deciding to synthesize. Cheap, and it turns a corrupted interior node into a correct one.

comfy_cli/cql/promoted.py
```diff
@@ -544,3 +596,878 @@ def trace_upstream_write(
...
+            return "target slot type incompatible"
+    return None
+
+
+def planned_repair(workflow: dict, instance: dict, graph, widget: str, defs: dict[str, dict] | None = None):
+    """The repairable legacy entry a host address ``<instance>.<widget>`` names:
+    by the input name the repair will mint, else by the legacy tuple's own
+    widget name (an alias the caller reports as a redirect)."""
+    plan = [e for e in plan_proxy_migration(workflow, instance, graph, defs) if e.repairable]
+    by_name = next((e for e in plan if e.name == widget), None)
```
annehe9, line 1146: **Minor: the legacy-widget-name alias is ambiguous and resolves silently.**

The `e.widget == widget` fallback below takes the first match. `templates_graphic_design_recomposer.json` has two entries whose widget is `value` (`['156','value']` and `['142','value']`), so `143.value` always writes `resolution` and `aspect_ratio` is unreachable through that alias.

Low impact, since `slots` advertises the minted names (`143.resolution`, `143.aspect_ratio`) and never the alias, so only someone guessing the legacy name hits it. Still, raising on more than one match would beat picking one arbitrarily.

Separate nit while I am here: `control_after_generate` quarantines with reason `missingSubgraphInput`, but nothing is missing - the widget has no backing input slot at all, which is what `_slot_for_widget` says in its own docstring. If that string is the frontend's own reason code then fidelity wins, otherwise a distinct code would read better in the quarantine JSON.

annehe9 (commented):
Reviewed the port against the three gallery fixtures. The design is sound and I verified the properties the description claims: replay is byte-identical (single and shared-definition), shared definitions fork and siblings stay untouched, the flush is idempotent, `slots` reads do not mutate the document, the converter fan-out fix puts one host value on both primitive targets, and a quarantined `control_after_generate` still writes the interior. Full suite on `db071b0`: 6336 passed, 39 skipped, 2 failed, and both failures reproduce identically on `kishore/applier-op-shapes` (rich terminal-env artifacts, neither file in this diff).

Three findings inline, one of which I think is a blocker.

comfy_cli/cql/promoted.py
skishore23: Checked against the frontend: this IS what it does. `repairCreateSubgraphInput` calls `newSubgraphInput.connect(slot, sourceNode)` without looking at the slot's link, and `SubgraphInput.connect` (SubgraphInput.ts L85-108) resolves `inputLink(subgraph, node.id, slotIndex)`, `replaceLinkTopology`s it and `_disconnectNodeInput`s the incumbent — so on load the interior feeder is dropped and the host value runs. Keeping the replacement, but you were right that it was undocumented and untested: 905070f0 pins it with `test_flush_replaces_an_interior_link_feeding_the_source_slot` (node 2 wired into 135.choice → link 999001 gone from `links` and from node 2's `outputs[0].links`, boundary link installed) and `test_interior_address_owned_by_a_legacy_entry_is_redirected_even_when_link_fed` (same slot: refused without a legacy entry, host target with it). The `resolve_write` asymmetry now carries a comment explaining exactly that — the redirect targets the value the frontend will run after its own migration, whereas the unowned slot keeps its feeder and is refused.

comfy_cli/cql/promoted.py
skishore23: Fixed in 905070f0: `_slot_for_widget` now falls back to `entry.get('name') == widget` for a marker-less slot, and the flush restores the `widget: {name}` marker on it (frontend key order) instead of appending a duplicate. Pinned by `test_flush_reuses_a_slot_serialized_by_name_only` — marker stripped from 135.choice, red before (two `choice` entries), green after (one entry, marker back, label preserved on the minted input).

comfy_cli/cql/promoted.py
skishore23: Fixed in 905070f0: `planned_repair` raises when more than one legacy entry shares the widget name, listing the minted addresses (`143.value is ambiguous: ... 143.resolution, 143.aspect_ratio`); a single match still redirects. Pinned by `test_planned_repair_refuses_an_ambiguous_legacy_alias` and `test_ambiguous_legacy_alias_is_refused_with_the_minted_names` (document untouched on refusal). On the nit: `missingSubgraphInput` is the frontend's own code for this case — `repairCreateSubgraphInput`: "source widget has no backing input slot; quarantining" → `reason: 'missingSubgraphInput'` — so fidelity wins; the classification now says so in a comment.

annehe9 (approved):

skishore23:

PR #822, opened 2026-08-28
chore(hygiene): drop the internal ticket id from test_deprecated_nodes

skishore23:
The public-repo-hygiene check flags ticket-shaped ids. This one arrived in #808 and has kept `main` (and every open PR, whose merge ref includes it) red since. The sentence loses nothing without it.


skishore23:
Closing — the same one-line fix ships at the base of the open stack (#809, commit 7aba70fc), so no separate PR is needed.

PR #824, opened 2026-08-28
fix(cli): unbreak the interactive pickers and Go timestamp parsing on 3.10

guill:
<details><summary><b>Stack of 7 — review bottom-up</b> (this is the top)</summary>

7. #824 — `fix(cli)`: unbreak the interactive pickers and Go timestamp parsing on 3.10  &nbsp;&nbsp;**← this PR**

Each PR is based on the branch below it, so its own diff is only its commits. Merge in order.
</details>

**TL;DR:** Two independent defects found while exercising the stack by hand, neither caught by the suite. Every interactive picker in the CLI raised `ValueError` before it could draw, and roughly one Go-issued timestamp in ten was unparsable on Python 3.10 — the version CI actually runs. Stacked on #810.

---

## 1. `ce93092` — every interactive picker raised before drawing

`prompt_select` and `prompt_autocomplete` defaulted `default` to `""` and passed it straight to questionary. questionary validates that `default` is one of `choices`, and `""` never is:

```
ValueError: Invalid `default` value passed. The value (``) does not exist in
the set of choices. Please make sure the default value is one of the available choices.
```

This fires on construction, before any prompt renders, so **no caller with a real choice list worked**. That is every picker in the CLI:

| Call site | Picker |
|---|---|
| `build.py:2029` | the Build picker `comfy build pull` opens when given no `--id` |
| `deploy_compute.py:26` | GPU class |
| `deploy_compute.py:43` | Region |
| `uv.py:227` | Python version |
| `cmdline.py:1853`, `:1858` | feedback scores |

The fix is the correct sentinel: `default: ChoiceType | None = None`. questionary treats `None` as "no default" and selects the first choice, which is what the `""` was reaching for.

Worth noting for #810's reviewer: the broken picker is the same one the adopt flow depends on — `comfy build pull` with no `--id` lists the workspace's builds and asks you to choose one.

## 2. `f40f519` — Go's trimmed fractional seconds are unparsable on Python 3.10

Our Go services marshal `time.Time`, which encodes as RFC 3339 **Nano** — and that trims trailing zeros from the fractional part. A timestamp holding `437450` microseconds ships as `...:09.43745Z`: five fractional digits.

`datetime.fromisoformat` on Python 3.10 — `requires-python = ">=3.10"`, and the CI matrix runs 3.10 only — accepts exactly 3 or 6 fractional digits and rejects a bare `Z`. So that value raises. Trailing-zero trimming makes this roughly a one-in-ten event, and since a `createdAt` never changes, an affected row was **permanently** unreadable rather than intermittently so.

New `utils.parse_rfc3339` parses any sub-second precision and always returns an aware datetime: it pads so `.1` is a tenth rather than a microsecond, truncates so Go's nanosecond precision degrades instead of failing, and reads a missing zone as UTC because callers order these against each other and comparing naive to aware raises `TypeError`.

Three call sites move onto it, and each had its own failure shape:

* **`deploy_resolve.deployment_selection_key`** — raised `deploy_server_error` on an affected deployment. The status and timestamp checks are also split apart now: one `except (KeyError, ValueError)` covered both and reported `status=...`, so a valid `provisioning` was printed as the evidence while the timestamp that actually failed went unmentioned. Each is now reported separately, naming the value it rejected.
* **`build._release_order`** — compared `createdAt` **as text**, so a whole-second `...:16Z` sorted above the strictly later `...:16.5Z` (`.` precedes `Z` in ASCII), and `comfy build release` picked the wrong newest release. Now ordered on the parsed instant, with an `_UNDATED` floor so a badly dated release loses the tiebreak instead of winning it.
* **`jobs._parse_epoch`** — swallowed the failure and returned `0.0`, sorting an affected row below every dated one, precisely where `jobs ls` slices `[:limit]` and drops a fresh completion.

**Validation**

* `test_utils.py`, `test_deploy_resolve.py`, `test_jobs.py`, `test_build_release.py`: **314 passed**.
* Full command + output suites on this branch: **3749 passed, 3 skipped**. `ruff check .` / `ruff format --check .` clean at CI's pin (0.15.15).
* Seven new tests, including property-based coverage that a release-order key stays comparable for every `createdAt` shape and that a dated terminal row never sinks to the epoch-zero floor.
* The picker fix is verified by construction against the installed questionary: `default=""` raises, `default=None` constructs.

### DIFF
--- a/comfy_cli/ui.py
+++ b/comfy_cli/ui.py
@@ -43,7 +43,7 @@ def show_progress(iterable, total, description="Downloading..."):
 
 
 def prompt_autocomplete(
-    question: str, choices: list[ChoiceType], default: ChoiceType = "", force_prompting: bool = False
+    question: str, choices: list[ChoiceType], default: ChoiceType | None = None, force_prompting: bool = False
 ) -> ChoiceType | None:
     """
     Asks a single select question using questionary and returns the selected response.
@@ -65,7 +65,7 @@ def prompt_autocomplete(
 
 
 def prompt_select(
-    question: str, choices: list[ChoiceType], default: ChoiceType = "", force_prompting: bool = False
+    question: str, choices: list[ChoiceType], default: ChoiceType | None = None, force_prompting: bool = False
 ) -> ChoiceType | None:
     """
     Asks a single select question using questionary and returns the selected response.
--- a/comfy_cli/command/build.py
+++ b/comfy_cli/command/build.py
@@ -36,8 +36,9 @@ import time
 import urllib.error
 from collections.abc import Callable, Iterator, Sequence
 from dataclasses import dataclass, replace
+from datetime import datetime, timezone
 from pathlib import Path
-from typing import TYPE_CHECKING, Annotated, NoReturn
+from typing import TYPE_CHECKING, Annotated, Final, NoReturn
 from urllib.parse import urlsplit
 
 import requests
@@ -96,6 +97,7 @@ from comfy_cli.constants import SUPPORTED_PT_EXTENSIONS
 from comfy_cli.interaction import confirm, require_option
 from comfy_cli.output import get_renderer
 from comfy_cli.registry.api import sanitize_error_body
+from comfy_cli.utils import parse_rfc3339
 
 if TYPE_CHECKING:
     from comfy_cli.builder_api import BuilderClient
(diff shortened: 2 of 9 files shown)


comfy_cli/ui.py
```diff
@@ -43,7 +43,7 @@ def show_progress(iterable, total, description="Downloading..."):
 
 
 def prompt_autocomplete(
-    question: str, choices: list[ChoiceType], default: ChoiceType = "", force_prompting: bool = False
+    question: str, choices: list[ChoiceType], default: ChoiceType | None = None, force_prompting: bool = False
```
skishore23, line 46: Nit, non-blocking: autocomplete's default is a prefill string, not a choice, so `default=None` raises TypeError inside questionary if anyone ever calls this without a default (no callers today). Either keep this one as `default: str = ""` or coalesce None to "" before passing it down.

skishore23 (approved):
Reviewed at 5c21a05. Both fixes address the real root cause.

- Picker: questionary 2.1.x validates select's default against the choice list, so the old `default=""` crashed every prompt_select caller that did not pass one (build picker, GPU class and region pickers, feedback). `default=None` is the documented no-default; verified against the locked questionary.
- Timestamps: `parse_rfc3339` normalizes the fraction and offset before fromisoformat, so 3.10 parses what Go emits. All three wire-parsing sites are converted, and the remaining direct fromisoformat sites only read locally written stamps that 3.10 accepts. CI runs 3.10, which is where the fix differentiates.

One inline nit on prompt_autocomplete, non-blocking (no callers today). The mix_stderr and click.Group asserts in tests remain open, inert under the locked click 8.1.8 / typer 0.21.1 - worth one typer-compat follow-up across the stack rather than fixing here.

comfy_cli/ui.py
guill: Are you sure that's true? I specifically encountered issues with `= ""` and from my research it looks like `None` is used to specify "no default". If you pass an empty string, it looks for a valid entry that is an empty string and then errors if there is none.

PR #828, opened 2026-08-31
feat(knowledge): carry the bundle's compile date in every envelope (BE-9893)

annehe9:
## What changed

Every knowledge payload now carries `compiled_at`, the bundle's content date, read from `manifest.json`.

**Merge after Comfy-Org/comfy-knowledge#33**, which is what writes the key. Until a bundle carrying it is published, this ships `compiled_at: null` everywhere and looks inert.

## Why

A consumer reading an enrichment envelope had no way to tell how old the advice was. Two fields look like the answer and neither is one.

`stale` is true only on the `stale-cache` path: the TTL expired and no fetch replaced it. A caller that sets `COMFY_KNOWLEDGE_FILE` takes the `env` branch, where it is hardcoded `False` no matter how old the file is. The cloud agent always sets that variable, so `stale` is structurally always `False` there.

`as_of` is the loaded file's mtime. On the cloud agent that file is written by a GCS warm fetch at pod start, so it reports how long the pod has been up and says nothing about the content. Before the warm fetch shipped it was at least the image build time.

Found triaging eval run `run_3e91922d97784821be2c96bf9a0debcf`: every response carried `stale: false` on canon that was 21 days behind, and nothing in the envelope contradicted it.

## What this does not change

Additive, per the bundle schemas' additive-only rule. `as_of` and `stale` keep their meanings and gain comments saying what those meanings are, so nothing that reads them today changes behaviour. A rename would also have put the code out of step with the design doc, which uses `stale` with exactly this meaning in two places.

`compiled_at` is `None` for a bundle compiled before the key existed, including the 0.1.4 bundle vendored into the cloud agent images, so old bundles keep loading.

## Where it ships

`status`, `resolve`, `pick`, the pick-capabilities miss, the `attach()` block and its miss marker, and the `knowledge_query` telemetry event. The miss log is the curation inbox, so a gap found against an old bundle should not rank beside one found against a current bundle.

Declared in `comfy_cli/schemas/knowledge.json` and `knowledge_block.json` as `["string", "null"]` and documented as opaque. The producer validates the date; a consumer must not, which is the tolerant-reader rule the bundle schemas already state.

An over-long value from a hostile manifest is dropped rather than truncated. Cutting a timestamp to length yields a date that parses to the wrong instant, and no date beats a wrong one. The `version` field beside it tolerates truncation because nothing parses it.

## Testing

`7289 passed, 39 skipped`. ruff check and format clean on every file touched.

`TestCompiledAtReachesEveryPayload` covers each emit site. Confirmed by deleting each `"compiled_at": bundle.compiled_at,` line and checking a test fails: without these, all six sites could be removed with the suite still green.

End to end against a bundle compiled from comfy-knowledge#33:

```
manifest: {'version': '0.1.4', 'compiled_at': '2026-08-31T17:28:37Z'}
envelope: {'bundle_version': '0.1.4', 'compiled_at': '2026-08-31T17:28:37Z',
           'stale': False, 'as_of': '2026-08-27'}
```

`as_of` there is the capability row's curator date, a different field from `Bundle.as_of`. The point of the change is that these are now distinguishable.

## Follow-ups, both blocked on a pin bump

Cloud can stamp `compiled_at` as a Langfuse span attribute beside `bundle_version` in `services/agent/internal/loop/knowledge_span.go`, and contract `data.compiled_at` in `services/agent/internal/loop/cli_fields.json`. Neither can land until this merges and the agent's comfy-cli pin moves.

### DIFF
--- a/comfy_cli/command/knowledge.py
+++ b/comfy_cli/command/knowledge.py
@@ -84,6 +84,7 @@ def status_cmd(
             "stale": bundle.stale,
             "version": bundle.version,
             "schema_version": knowledge.SCHEMA_VERSION,
+            "compiled_at": bundle.compiled_at,
             "as_of": bundle.as_of,
             "path": bundle.path,
             **_env_context(),
@@ -137,6 +138,7 @@ def resolve_cmd(
         "model": row,
         "deprecation": bundle.deprecations.get(model_id),
         "bundle_version": bundle.version,
+        "compiled_at": bundle.compiled_at,
         "stale": bundle.stale,
     }
     if renderer.is_pretty():
@@ -164,6 +166,7 @@ def _emit_capabilities(renderer, bundle: knowledge.Bundle, *, query: str | None 
         "capabilities": caps,
         "zero_hit": query is not None,
         "bundle_version": bundle.version,
+        "compiled_at": bundle.compiled_at,
         "stale": bundle.stale,
     }
     if query is not None:
@@ -212,6 +215,7 @@ def pick_cmd(
         "as_of": _text(cap.get("as_of")),
         "picks": picks,
         "bundle_version": bundle.version,
+        "compiled_at": bundle.compiled_at,
         "stale": bundle.stale,
     }
     if renderer.is_pretty():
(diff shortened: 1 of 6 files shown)


skishore23 (approved):
Approving. Additive and tolerant-reader throughout: every emit site (`status`, `resolve`, `pick`, pick-miss, `attach` block, miss marker, `knowledge_query` telemetry) has its own pinning test, the over-long-value guard drops rather than truncates (right call for a timestamp), `None` for pre-key bundles keeps 0.1.4 loading, and both schemas document the three date-ish fields against each other. The `stale`/`as_of` clarifying comments are worth having independent of the new field. Ran knowledge/attach/enrichment suites on the branch: 277 passed. Sequencing note acknowledged: inert (`compiled_at: null`) until comfy-knowledge#33 publishes a bundle carrying the key.

PR #829, opened 2026-08-31
perf(workspace): resolve the workspace on first read (BE-10535)

annehe9:
Linear: [BE-10535](https://linear.app/comfyorg/issue/BE-10535). Follow-up to BE-9907.

## Problem

`entry()` in `comfy_cli/cmdline.py:301` calls `setup_workspace_manager` for every subcommand before dispatch. That eagerly ran `get_workspace_path()`, which calls `check_comfy_repo(os.getcwd())`, which imports GitPython and opens the repo. Only `--version` and `--help-json` short-circuit before it. Every command that never reads the workspace (`knowledge`, `cloud`, `skills`, ...) paid for a value it discards.

## Change

Option 2 from the ticket. `setup_workspace_manager` now only stores the flags. `workspace_path` and `workspace_type` are properties that resolve on first read and cache the pair.

Confirmed nothing reads the path eagerly, so no fallback to option 1 was needed. Every reader is inside a command handler: `cmdline.py:749/816/1687/1697`, `command/build.py:933+`, `command/launch.py:415/432/848`, `command/custom_nodes/*`, `command/models/models.py:55`. `_maybe_nudge_setup` and `tracking.prompt_tracking_consent` do not touch it. `command/install.py:273` calls `setup_workspace_manager` a second time after creating the repo; that discards the cached pair.

The property setters and deleters exist because roughly 100 test call sites do `patch.object(manager, "workspace_path", ...)`. Patching a property needs a setter, and unpatching goes through the deleter.

## Measurements

Median of runs on a quiet machine, dependencies from `uv sync --locked`, two alternating rounds against `origin/main`.

| | main | branch |
|---|---|---|
| `comfy --json` (bare), min of 20 | 0.13 s / 0.17 s | 0.09 s / 0.09 s |
| `git` in `-X importtime` | 58 ms | absent |

`comfy --json knowledge status` shows no measurable change (0.86-0.94 s either way). The two sequential authed GETs dominate that command; BE-10534 is the ticket for those.

## Tests

`uv run --locked --extra dev pytest`, matching the CI workflow:

- `origin/main`: 7281 passed, 39 skipped
- this branch: 7289 passed, 39 skipped

The 8 new tests are in `tests/comfy_cli/test_workspace_manager.py`. Five of them fail against `origin/main`, so they pin the change rather than restating it:

- `setup_workspace_manager` does not probe (`check_comfy_repo` patched to raise)
- first read resolves exactly once, and caches
- a second `setup_workspace_manager` discards the cached path
- `knowledge status` and `cloud status` run with `check_comfy_repo` patched to raise
- a subprocess asserts GitPython never reaches `sys.modules`
- `which` still resolves the workspace, and the path is None before setup

`uvx ruff@0.15.15 check` and `format --diff` both clean, matching `.github/workflows/ruff_check.yml`.

## Behaviour note

`get_workspace_path` can raise `typer.Exit(1)` when `--recent` is passed with no recent workspace set. That now fires on first read rather than in the callback, so `comfy --recent knowledge status` succeeds instead of aborting. That is the point of the change: a workspace-free command should not be gated on workspace resolution. Every workspace-consuming command still aborts exactly as before.

### DIFF
--- a/comfy_cli/workspace_manager.py
+++ b/comfy_cli/workspace_manager.py
@@ -161,9 +161,9 @@ class WorkspaceManager:
         self.specified_workspace = None
         self.use_here = None
         self.use_recent = None
-        self.workspace_path = None
-        self.workspace_type = None
         self.skip_prompting = None
+        self._flags_set = False
+        self._resolved: tuple[str | None, WorkspaceType | None] | None = None
 
     def setup_workspace_manager(
         self,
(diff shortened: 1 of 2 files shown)


skishore23 (approved):
Approving. Checked the two ways this could bite and both are clean:

- **`typer.Exit` moving from the callback into first-read sites:** swept every `workspace_manager.workspace_path` reader on main for enclosing try/except. The only broad `except Exception` around a first-read site is the shell-completion helper in `custom_nodes/command.py` (~499), where returning `[]` on any failure is exactly right; `get_installed_packages` catches only `subprocess.CalledProcessError`, so an Exit still propagates. The disclosed behaviour change (`--recent` with no recent workspace now succeeds for workspace-free commands, still aborts for consumers) is the point of the ticket.
- **The property/patch compatibility story:** setter/deleter keep `patch.object` working, the second-`setup` discard covers `comfy install`'s re-setup, and the subprocess test pinning GitPython out of `sys.modules` is the strongest possible guard against regression.

Ran workspace_manager + login + knowledge suites on the branch: 166 passed; CI fully green. Nice measurement discipline in the description.

PR #830, opened 2026-08-31
perf(file_utils): import httpx lazily (BE-10537)

annehe9:
Linear: [BE-10537](https://linear.app/comfyorg/issue/BE-10537). Follow-up to BE-9907.

## Problem

`comfy_cli/file_utils.py` imported httpx at module top, and two module-level constants built httpx objects at import time (`_DOWNLOAD_TIMEOUT = httpx.Timeout(...)` and the transient/retriable exception tuples). Only `_download_file_httpx`, `download_file` and `_friendly_network_error` need httpx. Everything else that imports this module wants `atomic_write_text` or `atomic_write_bytes` alone: `knowledge.py`, `build.py`, `build_spec.py`, `outdated.py`, `project.py`, `workflow.py`, `cql/loader.py`, `jobs_state.py`, `skills/__init__.py`.

## Change

`import httpx` moved into those three functions. In `download_file` it sits after the aria2 early return, so the aria2 path never imports it.

The two constants became three `functools.cache` helpers built on first download: `_download_timeout`, `_transient_exceptions`, `_retriable_exceptions`. Same classes in the same order, and the cache means `except _retriable_exceptions() as exc:` compares against the same tuple object every time, so timeout and retry behaviour is unchanged. Nothing outside the module referenced the removed names.

`atomic_write_text` / `atomic_write_bytes` were not split into a new module, per the ticket.

## Measurements

`python -X importtime -c "import comfy_cli.knowledge"` on a quiet machine, dependencies from `uv sync --locked`, min of 11, two alternating rounds against `origin/main`.

| | main | branch |
|---|---|---|
| `comfy_cli.knowledge` | 79.0 ms / 79.5 ms | 70.0 ms / 70.9 ms |
| `httpx` line | 11.3 ms | absent |

The ticket estimated 0.13 s. The real figure on a warm machine is 11 ms, because `rich` is already loaded by the time httpx pulls `httpx._main`.

## Tests

`uv run --locked --extra dev pytest`, matching the CI workflow:

- `origin/main`: 7281 passed, 39 skipped
- this branch: 7282 passed, 39 skipped

The new test is `test_knowledge_import_does_not_pull_httpx` in `tests/test_file_utils_network.py`. It runs `-X importtime` in a subprocess and asserts httpx is absent from the import list. Verified it fails against `origin/main`.

Download behaviour is covered by the existing suites, which pass unchanged: `tests/test_file_utils_network.py` and `tests/comfy_cli/command/test_model_download_background.py`, 301 tests including the timeout and retry paths.

`uvx ruff@0.15.15 check` and `format --diff` both clean, matching `.github/workflows/ruff_check.yml`.

### DIFF
--- a/comfy_cli/file_utils.py
+++ b/comfy_cli/file_utils.py
@@ -1,3 +1,4 @@
+import functools
 import json
 import logging
 import os
@@ -10,7 +11,6 @@ import zipfile
 from collections.abc import Callable
 from http import HTTPStatus
 
-import httpx
 from pathspec import PathSpec
 
 from comfy_cli import constants, ui
@@ -471,13 +471,6 @@ _VALID_DOWNLOADERS = {"httpx", "aria2"}
 
 _DOWNLOAD_MAX_RETRIES = 3
 _DOWNLOAD_RETRY_BACKOFF = 2  # seconds multiplier
-_DOWNLOAD_TIMEOUT = httpx.Timeout(10.0, read=300.0)
-_TRANSIENT_EXCEPTIONS = (
-    httpx.TimeoutException,
-    httpx.NetworkError,
-    httpx.ProtocolError,
-    httpx.ProxyError,
-)
 # HTTP statuses that typically indicate a transient server-side or rate-limit
 # problem worth retrying with backoff. Auth/not-found/redirect statuses stay
 # out of this set so they fail fast.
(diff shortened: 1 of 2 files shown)


skishore23 (approved):
Approving. Verified the three things that matter for a lazy-import move: nothing outside `file_utils.py` references the removed names (`_DOWNLOAD_TIMEOUT`/`_TRANSIENT_EXCEPTIONS`/`_RETRIABLE_EXCEPTIONS` — the only grep hit is `transfer.py`'s unrelated `_DOWNLOAD_TIMEOUT_S`); the `functools.cache` helpers return the same tuple object every call so except-clause behaviour is identical; and the aria2 early-return really does precede the `import httpx` in `download_file`. The `-X importtime` subprocess test pins the win. Ran `test_file_utils_network.py` + `test_file_utils.py` on the branch: 129 passed; CI fully green.

PR #838, opened 2026-09-02
fix(generate): route --emit-workflow through workflow_ops CRDT primitives (BE-11131)

christian-byrne:
`comfy generate <model> --emit-workflow` / `comfy workflow compose` build API-format workflow dicts (`workflow["1"] = {"class_type": ..., "inputs": ...}`, sequential string ids) by hand-writing dict entries. `comfy workflow` slot-editing commands (`list_slots`, `set_widget`, `connect`, …) gate on `_is_frontend_format`, which requires `nodes` to be a list — API format keys `nodes` as a dict, so every one of those commands rejected `--emit-workflow` output as `workflow_not_frontend_format` (Linear [BE-11131](https://linear.app/comfyorg/issue/BE-11131), 48 staging refusals).

`emit.py::build_workflow` now constructs the partner node, `LoadImage` loaders, `ImageBatch` chains, and the `SaveImage`/`SaveVideo` sink via `workflow_ops.add_node`/`connect`/`set_widget` — the same CRDT-ready op primitives `comfy workflow` slot-editing already requires — instead of a local sequential-id counter and raw dict writes. Output is frontend-format (`nodes[]`/`links[]`) plus the replayable op list those primitives emit, so a later `--emit-ops` flag (tracked separately) can hand it straight to `apply_ops`.

`--emit-workflow` has no server/API-key dependency (works with no ComfyUI instance running), so the node classes it can address — the 5 partner nodes in `MODEL_NODE_MAP` plus `LoadImage`/`ImageBatch`/`SaveImage`/`SaveVideo` — are resolved against a bundled, offline `object_info` snapshot (`comfy_cli/command/generate/data/emit_object_info.json`) rather than a live catalog fetch. This is the same recorded-snapshot approach `test_emit.py`'s completeness-contract test already used; it's now the single source both the module and the test read, declared as package data in `pyproject.toml`.

`ImageBatch` is marked `deprecated` in the recorded snapshot (superseded upstream by a node this emitter doesn't yet target), and the `flux-2` partner node (`Flux2ProImageNode`) is too — every `add_node` call in this module passes `allow_deprecated=True` since `MODEL_NODE_MAP` intentionally still targets these classes.

This is PR A of a 3-PR split (see the linked plan): PR B does the same rewrite for `fragments.py::Pipeline` (`comfy workflow compose`), PR C adds the opt-in `--emit-ops` flag that writes the op batch alongside the workflow JSON.

## Evidence

```
$ uv run pytest tests/comfy_cli/command/generate/test_emit.py -q
31 passed in 4.04s

$ uv run ruff check comfy_cli/command/generate/emit.py comfy_cli/command/generate/data/ tests/comfy_cli/command/generate/test_emit.py
All checks passed!

$ uv run ruff format --check comfy_cli/command/generate/emit.py tests/comfy_cli/command/generate/test_emit.py
2 files already formatted
```

Confirmed the rest of `tests/comfy_cli/command/generate/*` (e.g. `test_app.py`, `test_list_schema_envelope.py`) times out identically on a clean `origin/main` checkout in this sandbox (no network) — pre-existing, not a regression from this change.

Plan doc (root cause + PR split, written before this PR): `research/architecture/ecw-52-emit-ops-implementation-plan-2026-09-01.md` in `christian-byrne/in-app-agent-program`.


skishore23 (changes requested):
Reviewed against current main, which now has #840 merged. Two blockers, both verified by running things rather than reading.

## 1. This merges clean and then breaks `--emit-ops`

`git merge origin/main` reports no conflicts. That is the bad case here, because the incompatibility is semantic and git cannot see it.

`build_workflow` changes shape twice in this PR. It returns `(workflow, ops)` instead of a dict, and the workflow is frontend format instead of API format. Main now has a second caller that still expects the old contract:

```
emit.py:411  workflow, _ops = build_workflow(...)   # this PR, fine
emit.py:518  api = build_workflow(...)              # from main, gets a tuple
             specs = ops_from_api_workflow(api, graph)
```

After the clean merge:

```
TypeError: tuple indices must be integers or slices, not list
10 failed, 332 passed    # tests/comfy_cli/command/generate/
```

All 10 failures are `test_emit_ops.py`. CI will catch it, so nothing reaches main broken, but the PR needs a rebase and a decision before it can land.

## 2. Same bug #840 had: the document never goes through `strip_internal`

`build_workflow` builds with `workflow_ops.add_node` / `set_widget` / `connect`, which leave apply-only bookkeeping on the dict. `write_workflow` then writes it straight to disk:

```
top-level keys on disk: ['_applied_ops', '_widget_stamps', 'last_link_id', 'last_node_id', 'links', 'nodes']
```

Two consequences. First, `_widget_stamps` seeds the last writer wins register, so the next edit against the emitted file gets dropped. 20 trials of emit then `set-widget` with default stamps:

```
EDITED: 7    DROPPED: 13
```

Every dropped one exits 0 and reports `ok: true`.

Second, `version` is missing, because `strip_internal` is also what calls `complete_save_format`. The frontend validator requires it, so the canvas rejects the file. That undercuts the point of the PR.

Fix is one line in `write_workflow`, same as #840:

```python
workflow_ops.strip_internal(workflow)
path.write_text(json.dumps(workflow, indent=2) + "\n", encoding="utf-8")
```

Worth a test that emits and then edits, since the current tests only check the shape and never edit the file afterwards.

## 3. Which contract do we want for `--emit-workflow`

Worth settling before the rebase, and it is a real design question, not a nit. #840 shipped `--emit-workflow` unchanged as API format and put frontend format behind `--emit-ops`. This PR changes `--emit-workflow` itself to emit frontend. Both solve the refusals. The difference is that this one changes the output format of an existing flag, so anything consuming API format from it breaks.

Building ops from the primitives directly, the way this PR does, is the nicer approach. `ops_from_api_workflow` on main is a conversion step that exists only because the builder produced API format. If we take this PR's direction then that function should go away, rather than both living in the file. That is a bigger change than a rebase, which is why it needs a call first.

## 4. Bundled snapshot has no provenance and no drift check

`emit_object_info.json` is 9 node classes with no recorded source version or capture date, no refresh script, and nothing in CI comparing it to a live catalog. Only `emit.py` reads it.

This matters more than usual because frontend format is positional. `widgets_values` is an array paired back to input names by `input_order`. One new required input upstream shifts everything. Simulated a single added input on `GeminiImageNode`, converting with main's fixed converter:

```
quality = 'MY PROMPT'
prompt  = 'gemini-2.5-flash-image'
model   = 42
```

Silent, no error. API format is immune to this because it names its inputs. #840 avoids it a different way, by resolving the live catalog at emit time instead of bundling one, though that costs the offline behavior this PR wants. If we keep the snapshot, it needs provenance in the file and a test that fails when it drifts.

## 5. Smaller

`allow_deprecated=True` is passed for every `add_node`, but only `Flux2ProImageNode` and `ImageBatch` are actually deprecated in the snapshot. A node deprecated later would be emitted with no signal. Scope it to the two classes, or read the snapshot's own flag.

Also note the branch predates the `input_order` fix in `convert_ui_to_api` that went in with #840, so files emitted from this branch today convert with the widget values shifted. I checked, rebasing fixes it, nothing to do here beyond the rebase.

## Verdict

Requesting changes for 1 and 2. Your own evidence reproduces exactly as written on the branch, 31 passed and ruff clean, so this is about the collision with what landed, not the work itself. Happy to take 3 either way, you have thought about this longer than I have. If we go your direction I would rather delete `ops_from_api_workflow` in the same PR than carry both.

PR #840, opened 2026-09-03
feat(generate): --emit-ops writes a frontend-format workflow plus a stamped op batch (BE-11131)

skishore23:
## What

`--emit-workflow` writes API format, which the ComfyUI canvas and every cloud-agent edit tool refuse (`workflow_not_frontend_format` — 48 refusals in one staging day), and which a shared-document (CRDT) consumer cannot attribute.

`--emit-ops` (requires `--emit-workflow`) re-expresses the **same** graph — `build_workflow` stays the single source of the model→node mapping — as `add_node`/`set_widget`/`connect` specs and materializes it through `workflow_ops.apply_specs`: the machinery every hand edit already uses, so widget order, autogrow growth and position assignment keep one answer. The written file becomes frontend format (canvas-editable), and the envelope carries a stamped `replace_ops` batch exactly like `templates fetch --emit-ops` (`--actor`/`--base-version` accepted the same way).

Also fixes a latent converter bug the round-trip contract exposed: `convert_ui_to_api` paired widgets positionally from the input **dict** order and ignored `input_order`, silently swapping neighboring widget values on any re-serialized catalog (GeminiImageNode's prompt/model traded places on an alphabetized fixture). It now honors `input_order` the way the cql engine's `_ordered_names` does.

## Testing

- `test_emit_ops.py` (11 tests, TDD): ops shape, apply-to-frontend, and the **round-trip contract** — lowering the materialized frontend workflow back to API format reproduces `build_workflow`'s semantics (image-edit, video, no-image, multi-image/ImageBatch cases); `write_frontend_workflow` file+batch behavior incl. the delete half over a previous graph; CLI end-to-end via `COMFY_OBJECT_INFO_FILE`; the converter `input_order` regression.
- Full suite: 424 passed in the touched areas; the only failures are 4 pre-existing on clean main in this sandbox (TTY/network-dependent spend-gate + stderr tests — verified by stashing the change).
- ruff check/format clean.

Consumer wiring (cloud agent `generate_workflow` passes the flags + pin bump) follows in Comfy-Org/cloud.

https://claude.ai/code/session_01QX1YteLA2BYbfjup13xNg1


skishore23:
**Coordination note:** this overlaps with #838 (opened ~12h earlier from the same ticket, BE-11131) — #838 is PR A of a 3-PR plan that rebuilds `build_workflow` on the op primitives directly, with B (compose) and C (`--emit-ops` flag) to follow. Converting this to draft to avoid a collision.

Proposed resolution: #838 lands first; this PR then rebases into its PR-C slot, contributing the pieces #838's plan hasn't covered yet: the `--emit-ops --actor --base-version` flag surface with the stamped `replace_ops` envelope batch (including the delete half over a previous canvas, matching `templates fetch --emit-ops`), the frontend↔API round-trip conformance test, and the consumer wiring in Comfy-Org/cloud#8274.

One piece here is independent of the collision and worth extracting either way: the `convert_ui_to_api` `input_order` fix (widgets were paired from input **dict** order, silently swapping neighboring values on any re-serialized catalog).

annehe9:
`write_frontend_workflow` skips `workflow_ops.strip_internal`, so the emitted file carries `_widget_stamps` to disk and the first edit against it is silently discarded.

`emit.py:492-494` writes the applied workflow directly:

```python
path.parent.mkdir(parents=True, exist_ok=True)
path.write_text(json.dumps(workflow, indent=2) + "\n", encoding="utf-8")
return workflow, ops
```

Every other write path in the repo goes through `strip_internal` first (`workflow_edit.py:99`, `:395`, `:792`, `:927`), whose docstring states the contract: *"Called at every point the document leaves this process, so the two guarantees — no internal bookkeeping, no missing save-format keys — hold for file writes, `--stdout` and batch output alike."*

Because it is skipped here, `_applied_ops` and `_widget_stamps` land in the file. `_widget_stamps` is the last-writer-wins register read by `_lww_gate` (`workflow_ops.py:384`), so a later op stamped below the emit run's `--base-version` is dropped. `set-widget` still reports `ok: true`.

## Repro

```
comfy generate nano-banana --prompt ORIGINAL --image cat.png \
  --emit-workflow x.json --emit-ops --actor agent --base-version 7

comfy workflow set-widget x.json "<node_id>.prompt" EDITED   # ok: true
```

Result, against the same graph with the bookkeeping keys removed as a control:

```
A: file as emitted            -> ['ORIGINAL']   # edit discarded, exit 0
B: same graph, stamps stripped -> ['EDITED']
```

The stamp that wins:

```json
{"[\"widget\", \"3240488523872070\", \"prompt\"]": [7, "agent", "ef21759ae1074916b06fa8002deb026c"]}
```

This gets worse the higher `--base-version` climbs, which is the direction the cloud-agent consumer will drive it. It also undercuts the PR's own premise: the file is accepted by `comfy workflow` (I confirmed `workflow slots` reads it, and rejects the plain `--emit-workflow` artifact as before), and then the first edit against it goes nowhere.

## Fix

One line, verified:

```python
workflow_ops.strip_internal(workflow)
path.parent.mkdir(parents=True, exist_ok=True)
path.write_text(json.dumps(workflow, indent=2) + "\n", encoding="utf-8")
```

After it the file's top-level keys are `['last_link_id', 'last_node_id', 'links', 'nodes', 'version']` and the `set-widget` above applies. `strip_internal` mutates in place, so `assert on_disk == wf` in `test_write_frontend_workflow_writes_frontend_and_returns_replace_batch` still holds.

Worth a regression test that emits with a non-zero `--base-version`, edits, and asserts the new value, since the existing tests only check the file is frontend-shaped and never edit it afterwards.

This is independent of the #838 sequencing — whichever PR ends up owning the write, the emitted document has to leave through `strip_internal`.

annehe9:
Correcting my comment above: I understated this. It is not limited to a high `--base-version`, it hits the plain invocation with no flags at all.

`_stamp_key` (`workflow_ops.py:376`) is `[base_version, actor, op_id]`. With defaults on both sides, the emit run and the later edit both stamp `[0, "cli", ...]`, so the comparison falls through to the random `op_id` hex. Whether your edit survives is a coin flip.

Twenty trials, no `--actor` and no `--base-version` anywhere:

```
comfy generate nano-banana --prompt ORIGINAL --image cat.png --emit-workflow t.json --emit-ops
comfy workflow set-widget t.json "<node_id>.prompt" EDITED

as-is                                  → 10/20 silent drops
with strip_internal before the write   →  0/20
```

Every drop returns `ok: true`.

One refinement to the fix. Put the call before `replace_ops` rather than just before `path.write_text`, so the returned `workflow` and the emitted batch describe the same document:

```python
workflow_ops.strip_internal(workflow)
try:
    ops = workflow_ops.replace_ops(previous, workflow, actor=actor, base_version=base_version)
```

`test_emit_ops.py` still passes 11/11 with that in place.

Also withdrawing one thing I raised elsewhere: the API-format-previous case does not diverge from `templates fetch --emit-ops`. That path reads `old.get("nodes")` the same way, so neither emits the delete half. No issue there.

skishore23:
pushed c4239b6 and 63f2feb.

**strip_internal** (the two comments above): confirmed, `write_frontend_workflow` was writing the `apply_specs` product straight to disk with `_widget_stamps` and `_applied_ops` in it. Now calls `workflow_ops.strip_internal(workflow)` before `replace_ops`, so the returned workflow, the file and the batch are the same document. Two regression tests: one asserts the bookkeeping keys are gone from disk and from the returned dict, the other does the real repro end to end (emit at `--base-version 7`, then a default `workflow set-widget` on the partner node's prompt, then lowers the file and asserts the new value survived). Both were red before the one line change, green after.

**coderabbit, `--emit-ops` without `--emit-workflow`**: now goes through `_bail` with a `SchemaError` like the other bad-arg branches, so `generate:start` gets its `generate:error`. Added a lifecycle test for it.

**coderabbit, declare the flags with Typer**: not doing that in this PR. `generate` runs with `allow_extra_args` + `ignore_unknown_options` because everything after the model alias is schema driven per model, and all eleven run-level flags (`--download`, `--async`, `--json`, `--timeout`, `--api-key`, `--emit-workflow`, `--output-prefix`, `--yes` plus the three new ones) go through `_separate_meta_flags`. Pulling three of them out into Typer options would leave two parsers for one surface. The valid part was discoverability, so the hand-rolled `generate --help` now shows the `--emit-ops [--actor ID] [--base-version N]` form. Moving all eleven at once can be its own PR.

the three inline coderabbit comments from the first pass (base-version via `_bail`, `_get_graph` exit tracked, ticket id out of the docstring) were already done in c5a50ec.

annehe9 (approved):

PR #846, opened 2026-09-03
fix(workflow): reshape dynamic combo values on set-widget

christian-byrne:
Seedance copies now keep the selected model schema.

- Dynamic-combo edits rebuild their positional widget roster.
- Seed and watermark stay aligned; copied 2.0 values survive API lowering.
- Selector/sub-widget races are surfaced as merge conflicts.

## Root cause

The cloud agent already emits name-keyed `set_widget` operations and the frontend materializes the resulting positional array. The top-level comfy-cli `set_widget` apply path duplicated a scalar index write instead of using comfy-cli's existing schema-aware CQL writer. Changing `model` therefore changed only the selector while leaving the previous option's variable-width sub-widget span in place.

The apply path now delegates to that schema-aware writer, which replaces the old dynamic option roster with the new option's defaults and preserves values after the roster. The edit receipt also carries the existing `dynamic_combo_roster_rebuilt` warning. Conflict detection now treats a selector and one of its dotted sub-widgets as overlapping writes while leaving trailing fields such as `seed` independent.

## Regression coverage

The production `ByteDance2ReferenceNodeV2` object-info fixture and Seedance 2.5 gallery workflow cover:

- Seedance 2.5 → 2.0 roster shrink.
- Seedance 2.0 → 2.5 roster growth.
- operation replay fidelity.
- agent-style `add_node` plus name-keyed writes copying customized Seedance 2.0 values.
- API lowering with numeric `seed=42` and no stale 2.5-only `task_type`/`output_format` fields.
- selector/sub-widget conflict detection and selector/seed commutativity.

## Evidence

- `.venv/bin/pytest -q tests/comfy_cli/command/test_workflow_edit_dynamic_combo.py tests/comfy_cli/cql/test_engine.py tests/comfy_cli/command/test_workflow_slots.py tests/comfy_cli/command/test_workflow_edit.py` — 389 passed.
- `.venv/bin/pytest -q tests/comfy_cli/command/test_workflow_edit_dynamic_combo.py` — 3 passed after formatting.
- `.venv/bin/ruff check .` — passed.
- `.venv/bin/ruff format --check .` — 447 files formatted.
- `.venv/bin/pytest -q` — 7,379 passed, 38 skipped, 5 failed. All five failures are in untouched environment-sensitive tests (`test_node_deps`, rich ANSI rendering, two file-mode/default-ACL assertions, platform CA roots); the same five fail when invoked from the unmodified `origin/main` test tree on this worker.

## Sources

- [Linear BE-10306](https://linear.app/comfyorg/issue/BE-10306/agent-cant-wire-asset-load-nodes-load-image-video-audio-into-nodes)
- [Jo's Seedance reproduction](https://comfy-organization.slack.com/archives/C0BGH348Z0C/p1788460446804899)

</details>

### DIFF
--- a/comfy_cli/workflow_ops.py
+++ b/comfy_cli/workflow_ops.py
@@ -777,6 +777,15 @@ def _set_widget_impl(
     value, norm_note = _normalize_combo(graph, class_type, widget, value)
     old = widgets[idx] if idx < len(widgets) else None
     warnings = _validate_widget(graph, class_type, widget, value)  # raises on shape mismatch
+    morphism = graph.node(class_type)
+    port = next((p for p in morphism.inputs if p.name == widget), None) if morphism is not None else None
+    if port is not None and port.dynamic_options:
+        # Preview through the same value-aware writer apply/replay uses. A
+        # selector change owns a variable-width positional span, so its roster
+        # rebuild warning cannot be derived by scalar validation alone.
+        for warning in _engine._write_widget(copy.deepcopy(node), widget, value, graph, extend=True):
+            if warning not in warnings:
+                warnings.append(warning)
     if norm_note:
         warnings = [norm_note, *warnings]
     op = _new_op(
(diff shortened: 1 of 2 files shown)


skishore23 (approved):
Approving. Re-verified against the BE-10306 repro on top of main `3fddc3e1` with the Sep 2 cloud `object_info`.

**Reproduced on main, fixed here**

- `set-widget <id>.model "Seedance 2.0"` on a fresh `ByteDance2TextToVideoNode` left the 2.5 roster in place, so validate reported `seed: expected INT, got str 'mp4'` and `watermark: expected BOOLEAN, got int`. With this branch the array is rebuilt to `['Seedance 2.0', '', '480p', '16:9', 7, True, 0, 'fixed', False]`, the receipt carries `dynamic_combo_roster_rebuilt`, and validate is clean. Switching back to 2.5 restores the wider roster with the trailing seed preserved.
- Same on the frontend-authored `api_seedance2_5_r2v` template (`ByteDance2ReferenceNodeV2`): reference links on `model.reference_*` survive the switch and a subsequent `seed` write lands at the right index.
- The `workflow apply --ops` path now reshapes too (it went through the same scalar write before).
- Bonus: main accepted an unknown option (`Seedance 9.9`) silently and corrupted the node; this branch rejects it with the valid list.

**Behaviour check against the canvas**: resetting sub-widgets to schema defaults while carrying links across matches `dynamicWidgets.ts::updateWidgets`, which removes and recreates every `<combo>.*` widget on a selection change.

Targeted suites pass under a typer 0.27 / click 8.5 venv (389 tests). Merging.

PR #850, opened 2026-09-04
fix(validate): hard-error a type-mismatched link on a server-reachable node

mattmillerai:
## ELI-5

If you wire a LATENT output into an IMAGE input, ComfyUI refuses to run the graph. `comfy workflow validate` used to shrug and say "looks fine, just a warning" — so an agent read the graph as runnable, submitted it, and watched the run die on the server. Now the validator says no for the same reason the server does, but only for the nodes the server actually looks at.

## What changed

`Graph.validate_workflow` reported a type-mismatched link as a `edge_type_mismatch` **warning**, so the envelope came back `ok:true, valid:true` for a graph that cannot run. The server hard-rejects it in `validate_prompt` — `return_type_mismatch`, "Return type mismatch between linked nodes", in `execution.py`'s `validate_inputs`. A downstream consumer computing readiness from errors alone therefore treats this class of graph as runnable, submits, and burns a submit and a run slot on a prompt that never starts.

The finding is now a hard **error**, gated on output-reachability exactly like `required_input_missing` / `below_min` / `above_max`: ComfyUI's `validate_prompt` only validates output nodes and their transitive input ancestors, so a mismatch on a node no output depends on stays an advisory warning rather than refusing a prompt the server would happily run.

Everything on the permissive side stays permissive — that is the deliberate mitigation for the over-rejection risk below, and it is what keeps this from trading false negatives for false positives:

- wildcards on either end (`*`, `COMFY_MATCHTYPE_*`) — no finding at all, unchanged;
- comma-separated union overlap (`FLOAT` into `INT,FLOAT`) — no finding at all, unchanged;
- blank / unrecognised types — no finding at all, unchanged;
- a wired `COMFY_DYNAMICCOMBO_V3` selector — keeps the advisory warning it has always emitted. Its option, and the sub-inputs that option contributes, resolve at execution time (which is why the expansion checks already skip a wired one), so its declared type is not what the server ends up comparing. New carve-out, called out below.

## Known over-rejection risk (stated at the call site as well)

The server **skips** its type check entirely when the destination node's `VALIDATE_INPUTS` declares an `input_types` argument (`'input_types' not in validate_function_inputs`), and `object_info` does not expose that signature — we cannot tell such a node apart from any other. Sized against a local ComfyUI checkout at `c68789b0`: the whole tree declares `VALIDATE_INPUTS` **8** times, exactly **1** of which takes `input_types`, and that one is a test-only node under `tests/execution/testing_nodes/`. **Zero shipping core nodes** use the escape hatch. Third-party custom node packs are not measurable from here — see Residual.

The compatibility predicate itself is the other half of the mitigation: `_edge_types_compatible` is strictly **more** permissive than the server's `validate_node_input`. Both do set intersection over comma-separated unions; ours additionally treats `COMFY_MATCHTYPE_*` as a wildcard and compares case-insensitively. So every edge this promotes is one the server also rejects, unless the `input_types` escape hatch above applies.

## Verification against the reported repro

Run through the real CLI against the captured sd15 catalog, `SaveImage.images` fed from `KSampler[0]`:

| | before | after |
|---|---|---|
| mis-wire on an output-reachable node | `ok:true valid:true`, exit 0, one warning | `ok:false valid:false`, exit 1, one `edge_type_mismatch` error |
| same mis-wire on a node reaching no output | `ok:true`, warning | `ok:true`, warning (unchanged) |

## Tests

New `TestEdgeTypeMismatchIsAHardError` covers the reported repro against the real sd15 catalog, its correctly-wired counter-experiment, the reachable/pruned pair on one fixture, union overlap, a wildcard on each end, and the dynamic-combo carve-out. Existing edge tests were updated rather than relaxed:

- `TestValidateWorkflow::test_edge_type_mismatch` now asserts the error, and gained a sibling for the pruned case;
- `TestMultiTypeEdge::_mismatches` and the new `_edge_findings` helper scan **errors and warnings**. This mattered: the old helper read `warnings` only, so every `== []` assertion in that class would have passed vacuously the moment a finding moved buckets;
- two `TestValidateEmptyCombo` fixtures wired a MODEL/VAE loader straight into `SaveImage.images` purely to make it output-reachable. That is a genuine mismatch, so the class gained a wildcard `PreviewAny` sink and those two cases now wire through it — the fixture was fixed, not the assertion weakened.

## Judgment calls

- **Which compatibility predicate.** The plan said to reuse the connect gate's predicate (`workflow_ops._types_compatible`). I kept the validator's own `_edge_types_compatible` instead. Both are set intersection over comma-separated unions with wildcards and unknown types permissive, but the validator's also recognises `COMFY_MATCHTYPE_*` as a wildcard and is case-insensitive. Swapping in the connect gate's version would have *removed* the match-type handling that was added precisely to kill a run of false `edge_type_mismatch` warnings, and re-introduced them as hard errors. The more permissive of the two is the correct one to promote on.
- **The dynamic-combo carve-out** is not in the plan; it came out of an existing test asserting that a wired selector validates clean because it resolves at execution time. It only ever *demotes*, so it preserves existing behaviour rather than changing it.
- **The error-code registration criterion is UNMET as literally written** — see Residual.

## Residual

- **`edge_type_mismatch` was not added to any error-code registry.** Neither of the repo's two registries applies, and adding it to either breaks a deliberate pin: `comfy_cli/error_codes.py` is the registry for envelope `error.code` values, and its enforcement test explicitly excludes `cql/engine.py`'s internal validation codes *and* fails any registered code that is never raised through `renderer.error`; `cql.engine.FATAL_FINDING_CODES` is pinned by `test_fatal_set_is_exactly_the_four_run_time_rejections` to exactly the four widget-**value** rejections the edit path refuses, and an edge finding never reaches that path. There is today no registry of `validate_workflow` finding codes — the prior promotion of `below_min`/`above_max` to hard errors added the code to none either. Worth its own change: either add a real registry for validation finding codes, or record explicitly that these codes are registry-exempt.
- **Third-party custom node packs were not swept for the `VALIDATE_INPUTS(input_types=…)` escape hatch.** The `8 definitions / 1 with input_types / 0 shipping` count above covers the ComfyUI core tree only. A custom node that uses it and is wired across a type boundary will now be hard-rejected here while the server would have run it. No pack corpus was available to this change; if one is, that sweep is the number that decides whether this needs an escape valve (e.g. demoting to a warning for classes outside core).
- **Downstream pin bump not performed.** The consuming agent picks this up only after its own image pin moves; that is a change in a different repository and outside this PR. No code change is needed there — the error flows through the envelope chain untouched and readiness flips false on its own.
- **Unexercised artifacts.** The parent tracking issue and the originating investigation's findings comment live in an internal tracker whose contents were not available here; only the plan reproduced into this work was read. Everything this PR claims about server behaviour was re-derived directly from a local ComfyUI checkout rather than taken from that plan.
- **Behaviour change, by design.** `comfy run` preflight now exits 1 on this class of graph instead of printing a warning and submitting. That is the point of the change, but it is a user-visible refusal of something that previously proceeded.

## Provenance

- **Authored by:** agent-work loop
- **Verified:** `pytest -q` (fresh 3.12 venv, `.[dev]`): 7388 passed, 38 skipped, 2 failed — `test_node_deps.py::test_non_pep440_installed_version_is_unknown_not_a_crash` and `test_http.py::test_an_unloadable_supplement_falls_through_to_the_platform_roots`, both confirmed failing identically on the unmodified base commit (environment-dependent, unrelated to this diff). `pytest -q tests/comfy_cli/cql/ tests/comfy_cli/test_edit_findings_contract.py tests/comfy_cli/command/test_run.py`: 685 passed. `ruff check .`: clean. `ruff format --check .`: 446 files formatted. Plus the before/after CLI runs in the table above.
- **Deviations:** the error-code registration in the plan is unmet (no applicable registry exists — first Residual bullet); the compatibility predicate chosen differs from the one the plan named, in the more-permissive direction (first Judgment call); one carve-out not in the plan (wired dynamic-combo selectors stay advisory).

### DIFF
--- a/comfy_cli/cql/engine.py
+++ b/comfy_cli/cql/engine.py
@@ -1505,11 +1505,11 @@ class Graph:
         # Server parity: ComfyUI's validate_prompt only validates output nodes
         # and their transitive input ancestors — any node not reachable from an
         # output is pruned and never validated (execution.py). Restrict the
-        # promoted hard checks (required-input presence, autogrow-required, and
-        # below_min/above_max ranges) to that reachable set, so a
-        # disconnected/incomplete node the server would silently drop isn't
-        # hard-rejected here. Edge/shape/enum checks are left as-is (pre-existing
-        # behavior, out of scope for this change).
+        # promoted hard checks (required-input presence, autogrow-required,
+        # below_min/above_max ranges, and edge type mismatches) to that
+        # reachable set, so a disconnected/incomplete node the server would
+        # silently drop isn't hard-rejected here. Shape/enum checks are left
+        # as-is (pre-existing behavior, out of scope).
         reachable = _output_reachable_node_ids(workflow, self)
 
         for node_id, node_data in workflow.items():
@@ -1662,9 +1662,38 @@ class Graph:
                         )
                         continue
 
-                    # (iii) type compatibility — advisory only.
-                    # ComfyUI allows cross-type wiring via reroutes, converters,
-                    # and wildcard ports; the server is the authoritative validator.
+                    # (iii) type compatibility — a hard error on a node the
+                    # server will actually run, advisory on one it prunes.
+                    #
+                    # The server hard-rejects a type-mismatched link
+                    # (`return_type_mismatch`, "Return type mismatch between
+                    # linked nodes" — execution.py's validate_inputs), so
+                    # reporting it as a warning produced an `ok:true,
+                    # valid:true` envelope for a graph that cannot run: a
+                    # downstream agent reads it as runnable, submits, and dies
+                    # server-side, wasting a submit and a run slot.
+                    #
+                    # Gated on output-reachability exactly like
+                    # required_input_missing / below_min / above_max: the server
+                    # prunes nodes not reachable from an output and never
+                    # validates them, so on a pruned node the mismatch stays an
+                    # advisory warning rather than hard-rejecting a prompt the
+                    # server would happily run.
+                    #
+                    # KNOWN over-rejection risk, accepted deliberately: the
+                    # server SKIPS this check entirely when the destination
+                    # node's VALIDATE_INPUTS declares an `input_types` argument
+                    # (`'input_types' not in validate_function_inputs`), and
+                    # object_info does not expose that signature — we cannot
+                    # tell such a node apart from any other. Mitigation is to
+                    # keep the permissive side permissive:
+                    # `_edge_types_compatible` is strictly MORE permissive than
+                    # the server's `validate_node_input` (same comma-separated
+                    # union set-intersection, plus COMFY_MATCHTYPE wildcards and
+                    # a case-insensitive compare), so only an exact
+                    # known-type/empty-intersection mismatch is ever promoted —
+                    # a wildcard, a union overlap, or a blank/unknown type on
+                    # either end still yields no finding at all.
                     port = port_by_name.get(input_name)
                     if port is not None:
                         src_type = src_m.outputs[out_idx].type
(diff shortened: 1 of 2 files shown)


skishore23 (approved):
Approving. Verified on a merge tree of this branch over current main (`05b32cfb`, which includes #846) with the Sep 2 cloud `object_info`.

**Behaviour**

| case | main | this branch |
|---|---|---|
| `SaveImage.images <- KSampler[0]` on an output-reachable node | `valid:true`, warning | `valid:false`, `edge_type_mismatch` error |
| same mis-wire on a node no output reaches | warning + `node_not_reachable_from_output` | unchanged |
| correct wiring through `VAEDecode` | clean | clean |
| the BE-10306 repro graphs (Nano Banana 2 / MiniMax H3 / Seedance r2v with grown `model.*` slots) | clean | clean |

`tests/comfy_cli/cql/`, `test_edit_findings_contract.py`, `test_run.py`, `test_workflow_validate.py` and #846's suite pass on the merged tree under a typer 0.27 / click 8.5 venv (696 tests).

**On the panel's High finding (`VALIDATE_INPUTS(input_types)` over-rejection)**: I sized it against a local install with ~40 custom packs. Five classes declare `input_types` (kaytool ToInt/DisplayAny, basic_data_handling comparison nodes, VNCCS model manager, Bjornulf AnythingToFloat) and every one of them types the affected input `*` / `IO.ANY`, which stays permissive here, so none would be promoted. That matches why nodes reach for the hatch in the first place: they already declare a wildcard and want to validate the runtime type themselves. Combined with the zero core users, I'm comfortable shipping without an escape valve.

**On the `UNKNOWN` sentinel finding**: measured zero `UNKNOWN`-typed inputs or outputs across the cloud catalog (3,699 classes, 19,609 inputs) and a 1,330-class local catalog, so there is no exposure today. Treating that sentinel as permissive in `_edge_types_compatible` would still be a cheap hardening for a follow-up.

**On partial execution**: a correct `SaveImage` plus a mis-wired `PreviewImage` now fails preflight where the server would run the good branch. That is the same trade the existing promoted checks (`required_input_missing`, `below_min`/`above_max`) already make, so it is consistent rather than new.

**Follow-ups worth their own tickets (none block this)**

- `workflow_to_api.py` bypass tracer: when a bypassed node has no type-compatible linked input, the `fallback_link` branch (~line 810) wires the first linked input regardless of type, where the frontend's `_getBypassSlotIndex` drops the link. Today that mis-typed edge reaches the server and is rejected there; with this PR it is refused at preflight instead. Same outcome, earlier, but the tracer should drop the link like the frontend does.
- Type check skips dotted keys: a mismatched link into an autogrow slot (`images.image0`) or a dynamic-combo sub-input never hits `port_by_name`, so it still validates clean while the server rejects it. Pre-existing gap, now the notable remaining false negative of this class.
- `connect` gates on canvas slot types (`*` on Reroute outputs and widget-converted inputs), so a mis-wire through a wildcard slot is accepted at edit time and only refused at validate/run. Consulting the catalog type when the document slot is a wildcard would give the agent the error at the right step.

Merging.

PR #851, opened 2026-09-04
fix(generate): drop 7 retired partner aliases and re-vendor the openapi spec

mattmillerai:
## ELI-5

`comfy generate list` was advertising seven models you cannot actually generate with. Five Stability ones were deleted from the API a while back — the route is simply not there any more, so asking for one could only ever fail. Two Flux ones (`flux-canny`, `flux-depth`) are gone at the provider itself: BFL's own public API no longer declares either endpoint, while their siblings still do. This deletes all seven from the menu, refreshes the copy of the API description the CLI ships (it was months stale), and adds two tests so the menu and the API description can't quietly drift apart again.

## What this changes

1. **Seven aliases and their allowlist tuples are gone** from `comfy_cli/command/generate/spec.py`: `stability-ultra`, `stability-sd3`, `stability-upscale`, `stability-upscale-creative`, `stability-upscale-fast`, `flux-canny`, `flux-depth`. `comfy generate list` drops from 52 models to 45. The full `/proxy/...` escape hatch for those ids goes with them, which is intended — `get_endpoint` resolves against the allowlist, and there is nothing behind these paths to reach.
2. **`comfy_cli/command/generate/spec/openapi.yml` is re-vendored.** It had not been refreshed since it was first added and carried 193 `/proxy/` paths against the 279 the API serves today. The new copy is the response body of `https://api.comfy.org/openapi` plus the single trailing newline `end-of-file-fixer` requires — that endpoint is public and needs no token, so refreshing it is a reproducible two-command download (documented in the `spec.py` module docstring, together with the newline-aware `cmp` that compares the two) rather than a hand-edit nobody can re-derive. Two consecutive fetches were byte-identical. The body is minified JSON under a `.yml` name: JSON is a subset of YAML 1.2, `_YamlLoader` already reads either (its point-less-exponent float resolver exists specifically for the JSON form), and the on-disk cache `comfy generate refresh` writes was already this same JSON.
3. **Allowlist drift becomes loud in tests, not at runtime.** `tests/.../test_spec.py` gains `test_every_allowlisted_endpoint_exists_in_vendored_spec` (every allowlisted id must be a `/proxy/` path the **bundled** spec declares — `_USER_CACHE` is pointed at a non-existent file so a developer's own cache cannot decide the result) and `test_every_alias_targets_an_allowlisted_endpoint`. `_registry()` keeps its `continue` on a missing node, so a stale user cache still cannot crash `generate`; only the shipped spec is held to the stricter rule.
4. **Three existing tests moved with the refreshed spec**, because upstream moved and not because the tests were wrong: `bfl/flux-pro-1.1/generate` now requires only `prompt` (`width`/`height` gained server-side defaults), its `output_format` enum gained `webp`, and the byteplus model enum gained the `dreamina-seedance-*` naming scheme alongside `seedance-*`. The three missing-required tests were passing `--prompt` and relying on `width`/`height` being required; they now omit `--prompt` instead, which is the durable version of the same assertion. Without that fix those three tests stopped exercising argument validation and started making real network calls.
5. Prose that named a retired partner is updated: `_HELP`, two `README.md` partner lists, and the `save_binary_response` docstring in `output.py` (which cited Stability as the example of a partner answering `image/*` inline) no longer name Stability; the `category` example list in `comfy_cli/schemas/generate_list.json` no longer names `controlnet`, since after this change no endpoint can emit that category.

## Evidence each alias is actually dead

Everything below is read-only. Probing a `/proxy/` route would dispatch upstream and bill credits, so no request was sent to any partner endpoint.

**The five Stability ids — the route is gone.** The public spec at `api.comfy.org/openapi` declares none of the five paths (`0` paths matching `stability` in the refreshed document). That means `_registry()` was already skipping all five on its `continue`, and `comfy generate list` was advertising five models that could not resolve against the API the CLI talks to. This is the direction the new test catches: re-add any of these five tuples and `test_every_allowlisted_endpoint_exists_in_vendored_spec` fails by name (verified by injecting `stability/v2beta/stable-image/generate/ultra` back and watching it go red, then reverting).

**The two BFL ids — the proxy route still exists; the capability behind it does not.** `/proxy/bfl/flux-pro-1.0-canny/generate` and `/proxy/bfl/flux-pro-1.0-depth/generate` are still declared in the public spec, so the proxy layer alone cannot tell you they are dead. The falsifying check is the provider's own published API: `https://api.bfl.ai/openapi.json` declares **no** `flux-pro-1.0-canny` and **no** `flux-pro-1.0-depth` path, while the siblings this PR keeps — `/v1/flux-pro-1.0-fill` and `/v1/flux-pro-1.0-expand` — are both present and alive in the same document. So it is not a credentials, connectivity or family-wide problem: those two model paths specifically no longer exist upstream, and BFL's endpoint index documents no `canny`/`depth` successor (their reference page for the canny endpoint now 404s). That is consistent with the prod status-code table recorded in the API service's own de-list change, which shows both paths answering 404 in the window their siblings answered 422. A user typing `comfy generate flux-canny` today reaches a route that resolves to a model the provider deleted.

## Judgment calls

**A `exclude_paths:` entry was added to `.github/workflows/public-repo-hygiene.yml`, and a reviewer should decide whether to keep it.** The refreshed spec body contains eight tokens the hygiene checker flags: six ticket-shaped ids written into upstream `description` prose and two hits on an IETF language-tag standard. I ran the checker locally at the pinned commit and confirmed the failure, then confirmed the exclusion clears it. The exclusion names that one exact file, not the `spec/` subtree, and the reasoning is written into the workflow next to it so the entry does not become folklore. The argument for it: the file is not source, it is an upstream document stored as served (plus the one required newline) so that `curl … | cmp` is how you verify it — redacting tokens would silently fork it from upstream and destroy that property; none of the flagged ids sit under a `/proxy/` path this CLI surfaces; and `api.comfy.org` already serves every one of them to any unauthenticated caller, so vendoring them publishes nothing new. The argument against it: that caller file previously said "no `exclude_paths:` here, deliberately", with the bar set at "unfixable both here and upstream" — and this **is** fixable upstream, by quoting those descriptions without ticket ids. I could not make that upstream fix from here. If a reviewer prefers the upstream fix, the alternative is to hold this PR until the ids are out of the served spec.

**The `.yml` file is now one 1.2 MB line of JSON.** GitHub will not render a useful diff for it. The alternative — pretty-printing — buys a diff nobody will read either, and costs the reproducible-`curl` property. Stated so it is not mistaken for an accident.

## Scope not covered, measured

The refreshed spec declares **279** `/proxy/` paths. This PR leaves the curated allowlist at **45** of them, so **234 declared proxy paths remain un-offered** by `comfy generate`. That is deliberate — this change retires dead entries and does not add new models — but it is the number, so nobody has to guess. In the other direction the sweep is now clean: **45 of 45** remaining allowlisted ids exist in the refreshed spec, and **0 of 45** aliases dangle, which is what the two new tests pin. What that sweep cannot measure is the `flux-canny`/`flux-depth` class of failure — an id the spec still declares while the model behind it is gone. Only the provider's own API document shows that, and nothing in this repo checks it.

## Verification

| Check | Result |
| -- | -- |
| `pytest tests/comfy_cli/command/generate` | 327 passed, 2 failed — both fail identically on current `main` (`0343f505`), see Residual |
| `ruff check .` / `ruff format --diff .` (0.15.15, the version CI pins) | clean |
| public-repo hygiene checker, run locally at the pinned commit | no internal-only references |
| `comfy generate list --json` (clean `HOME`, so the bundled spec, not a user cache) | 45 models, none of the seven retired names |
| allowlist vs refreshed spec | 45/45 allowlisted ids declared; 0 dangling aliases; 0 `stability` paths in the spec |
| new tests fail on re-added drift | injected a retired Stability tuple and a dangling alias; both new tests went red naming the offender, then reverted |
| BFL provider API, read-only | `api.bfl.ai/openapi.json` declares `flux-pro-1.0-fill` and `-expand`, and no `-canny` / `-depth` |

Note on ruff: run it as the pinned **0.15.15**. A newer ruff (0.16.0) reformats Python code blocks inside `docs/DESIGN-uv-compile.md` and, in doing so, corrupts an illustrative snippet (`] = None,` → `] = (None,)`). That churn is not in this diff.

## Residual

**The waiver cleanup on the API service side is not done here and cannot be.** That repo's partner-catalog waiver list carries R1 waivers for all seven ids, written against exactly the drift this PR removes. The moment that service bumps its pinned `comfy-cli` to a build containing this change, those waivers go stale and its partner-catalog check fails on them — so the pin bump has to delete the seven entries in the same change. This is cross-repo and private; nothing in this PR touches it, and it will not happen by itself.

**No test guards the two BFL removals.** The refreshed spec still declares both `/proxy/` paths (the de-list on the API-service side is a separate, unmerged change), so `test_every_allowlisted_endpoint_exists_in_vendored_spec` would **not** fail if someone re-added the two BFL tuples — the new test protects the five Stability ids only. The evidence for the BFL half is the provider's own published API document rather than anything this repo can assert, and I did not reproduce the upstream 404 with a request, because probing a `/proxy/` route dispatches and bills. A reviewer who wants the BFL half held back can drop those two hunks without affecting anything else in this PR.

**A de-listed alias can still go unnoticed for a week.** The contract check that would have caught this class of drift runs on `pull_request` only and never on the default branch, so an alias orphaned by a merge stays invisible until the next PR that happens to touch one of its trigger paths. This PR does not change that; it is worth its own fix and belongs to the workflow, not to this allowlist.

**Pre-existing failures I did not fix.** `tests/comfy_cli/command/generate/test_spend_gate.py::test_interactive_prompt_accept_proceeds` and `::test_interactive_prompt_goes_to_stderr_when_stdout_is_machine` fail in this environment on current `main` (`0343f505`) exactly as they do here — interactive-prompt tests, unrelated to this change. Out of scope, unfixed. The wider suite also carries pre-existing failures and collection errors from missing optional dependencies (e.g. `blake3`), identical on base.

**Unexercised artifacts.** The API service's openapi document, its waiver file and its open de-list PR are in a private repo — nothing in this PR verifies against them at build time, and any edit made to them after this PR was written is invisible to it. The BFL proxy routes themselves were never called, for the billing reason above.

## Provenance

- **Authored by:** agent-work loop
- **Verified:** `pytest tests/comfy_cli/command/generate`: 327 passed, 2 failed (both fail identically on current `main` `0343f505`); `ruff check .` and `ruff format --diff .` at the CI-pinned 0.15.15: clean; public-repo hygiene checker at the pinned commit: no findings; `comfy generate list --json` under a clean `HOME`: 45 models, none retired; allowlist-vs-spec sweep: 45/45 declared, 0 dangling aliases; both new tests verified red on injected drift and green after revert; `api.bfl.ai/openapi.json` read read-only to falsify the two BFL removals
- **Deviations:** the `exclude_paths:` entry on the hygiene workflow is an addition the ticket did not ask for and is required to land the re-vendor — see Judgment calls; the two BFL aliases are removed on the provider's published API rather than on their proxy route being absent, so no test guards them — see Residual


mattmillerai:
Addressed the review finding on the verbatim/byte-for-byte claim in `spec.py` (line 12) and the matching rationale in `.github/workflows/public-repo-hygiene.yml`: the docstring now says the vendored file is the served body **plus the one trailing newline `end-of-file-fixer` requires**, and carries the newline-aware comparison (`curl … | { cat; printf '\n'; } | cmp - …`) instead of implying a raw `cmp` works. The workflow comment is corrected the same way. The newline itself is kept — the pre-commit hook re-adds it — so documenting the normalization is the fix rather than dropping the byte.

Same commit also finishes the prose sweep the alias removal started: `README.md` named Stability in two partner lists and `output.py`'s `save_binary_response` docstring cited it as the example partner returning `image/*` inline.

Separately, the two BFL removals now rest on first-hand evidence rather than a second-hand status table: `https://api.bfl.ai/openapi.json` declares no `flux-pro-1.0-canny` and no `flux-pro-1.0-depth` path, while `/v1/flux-pro-1.0-fill` and `/v1/flux-pro-1.0-expand` — the siblings this PR keeps — are both present in the same document. PR body updated.

PR #863, opened 2026-09-11
V1.5: emit insert_workflow ops from the CLI

christian-byrne:
Draft. Adds `comfy workflow insert-workflow <file> <template>`, which emits one stamped `insert_workflow` op without validating, remapping, applying, or writing the workflow locally. Python gates green. Blocked on cmp publish: [comfy-multi-player draft PR adding the insert_workflow op](https://github.com/Comfy-Org/comfy-multi-player/pull/183).

Design: [Notion TDD — agent subgraph lifecycle, V1.5 row](https://www.notion.so/3d16d73d365081719b85ed515af98276). Linear: [agent editing subgraph — human tech review needed](https://linear.app/comfyorg/issue/BE-10305), [frontend agent subgraph follower](https://linear.app/comfyorg/issue/FE-2033).

## Why

The CLI emits the cmp `insert_workflow` operation while cmp remains the sole owner of validation, ID remapping, application, replay, and projection.

## What changed

- `comfy_cli/command/workflow.py`: new standalone `insert-workflow` command.
- `comfy_cli/command/workflow_edit.py`, `workflow_ops.py`: preserve the input payload verbatim and emit it without mutating or writing the local workflow.
- `comfy_cli/command/error_codes.py`: adds cmp rejection codes.
- `docs/op-vocabulary-v1.md`: documents the emit-only ownership boundary.
- `tests/comfy_cli/test_insert_workflow_op.py`: covers exact payload preservation, semantic pass-through, no local mutation, envelope shape, and standalone-only behavior.

## Verification

- `ruff check .` and `ruff format --check .`: pass.
- Focused tests: 12 passed.
- Full `pytest`: 7,407 passed, 32 skipped, 47 unrelated baseline failures.
- `git diff --check`: clean.
- Not run: round trip against a live dochost. Needs the cmp publish and pin bump tracked on the cloud PR.

## Decisions to veto

1. New command name `insert-workflow` rather than a flag on the existing `load` command.
2. The command refuses to emit more than one op.

## Ship order

cmp → cloud → comfy-cli (this PR) → frontend follower. Sibling PRs: [cloud agent tool draft PR](https://github.com/Comfy-Org/cloud/pull/8981).

</details>

### DIFF
--- a/comfy_cli/command/workflow.py
+++ b/comfy_cli/command/workflow.py
@@ -1934,6 +1934,9 @@ app.add_typer(_wfrag.fragment_app, name="fragment")
 
 from comfy_cli.command import workflow_edit as _wedit  # noqa: E402
 
+app.command("insert-workflow", help="Insert a complete workflow; emits one insert_workflow op.")(
+    _wedit.insert_workflow_cmd
+)
 app.command("add-node", help="Add a node to the graph; emits an add_node op.")(_wedit.add_node_cmd)
 app.command("connect", help="Wire an output slot to an input slot; emits a connect op.")(_wedit.connect_cmd)
 app.command("set-widget", help="Set a widget by name (`<id>.<widget>`); emits a set_widget op.")(_wedit.set_widget_cmd)
--- a/comfy_cli/command/workflow_edit.py
+++ b/comfy_cli/command/workflow_edit.py
@@ -148,7 +148,7 @@ def _graph_or_exit(input_path, host, port, renderer, where=None):
 @tracking.track_command("workflow")
 def insert_workflow_cmd(
     file: Annotated[str, typer.Argument(help="Frontend-format workflow JSON to update.")],
-    template: Annotated[str, typer.Argument(help="Frontend-format workflow JSON to insert.")],
+    template: Annotated[str, typer.Argument(help="Frontend-format workflow JSON to insert, or '-' for stdin.")],
     actor: ActorOpt = "cli",
     base_version: BaseVersionOpt = 0,
 ):
@@ -157,8 +157,13 @@ def insert_workflow_cmd(
     renderer.command = "workflow insert-workflow"
     p, workflow = _load_workflow_or_fail(renderer, file)
     try:
-        template_path = Path(template).expanduser()
-        inserted = json.loads(template_path.read_text(encoding="utf-8"))
+        if template == "-":
+            import sys
+
+            raw = sys.stdin.read()
+        else:
+            raw = Path(template).expanduser().read_text(encoding="utf-8")
+        inserted = json.loads(raw)
         if not isinstance(inserted, dict):
             raise ValueError("template must be a JSON object")
         _, op = workflow_ops.insert_workflow(workflow, inserted, actor=actor, base_version=base_version)
--- a/comfy_cli/command/workflow_edit.py
+++ b/comfy_cli/command/workflow_edit.py
@@ -151,7 +151,6 @@ def insert_workflow_cmd(
     template: Annotated[str, typer.Argument(help="Frontend-format workflow JSON to insert.")],
     actor: ActorOpt = "cli",
     base_version: BaseVersionOpt = 0,
-    stdout: StdoutOpt = False,
 ):
     """Insert a workflow template and emit one atomic ``insert_workflow`` op."""
     renderer = get_renderer()
@@ -166,14 +165,7 @@ def insert_workflow_cmd(
     except (OSError, json.JSONDecodeError, UnicodeDecodeError, ValueError) as e:
         _emit_edit_error(renderer, e, hint="provide a frontend-format workflow template JSON file")
         raise typer.Exit(code=1) from e
-    try:
-        _emit_op(renderer, p, op, base_version, "workflow insert-workflow")
-    except ValueError as e:
-        # The outbound edit transport may return cmp's rejection. Preserve its
-        # message exactly: it is already the most actionable explanation and
-        # callers must not have to unwrap a second client-side paraphrase.
-        renderer.error(code="workflow_edit_invalid", message=str(e))
-        raise typer.Exit(code=1) from e
+    _emit_op(renderer, p, op, base_version, "workflow insert-workflow")
 
 
 @tracking.track_command("workflow")
--- a/comfy_cli/command/workflow_edit.py
+++ b/comfy_cli/command/workflow_edit.py
@@ -128,6 +128,14 @@ def _finish(renderer, p, workflow: dict, op: dict, base_version: int, stdout: bo
     renderer.emit(payload, command=command, changed=not stdout)
 
 
+def _emit_op(renderer, p: Path, op: dict, base_version: int, command: str) -> None:
+    """Emit an op without applying it or writing the source workflow."""
+    payload = {"workflow": str(p), "op": op, "base_version": base_version, "wrote": None}
+    if renderer.is_pretty():
+        rprint(f"[bold green]✓[/bold green] {op['op']} emitted for [dim]{p}[/dim]")
+    renderer.emit(payload, command=command, changed=False)
+
+
 def _graph_or_exit(input_path, host, port, renderer, where=None):
     return _get_graph(input_path, host, port, where=where)
 
@@ -154,11 +162,11 @@ def insert_workflow_cmd(
         inserted = json.loads(template_path.read_text(encoding="utf-8"))
         if not isinstance(inserted, dict):
             raise ValueError("template must be a JSON object")
-        workflow, op = workflow_ops.insert_workflow(workflow, inserted, actor=actor, base_version=base_version)
+        _, op = workflow_ops.insert_workflow(workflow, inserted, actor=actor, base_version=base_version)
     except (OSError, json.JSONDecodeError, UnicodeDecodeError, ValueError) as e:
         _emit_edit_error(renderer, e, hint="provide a frontend-format workflow template JSON file")
         raise typer.Exit(code=1) from e
-    _finish(renderer, p, workflow, op, base_version, stdout, "workflow insert-workflow")
+    _emit_op(renderer, p, op, base_version, "workflow insert-workflow")
 
 
 @tracking.track_command("workflow")
--- a/comfy_cli/command/workflow_edit.py
+++ b/comfy_cli/command/workflow_edit.py
@@ -166,7 +166,14 @@ def insert_workflow_cmd(
     except (OSError, json.JSONDecodeError, UnicodeDecodeError, ValueError) as e:
         _emit_edit_error(renderer, e, hint="provide a frontend-format workflow template JSON file")
         raise typer.Exit(code=1) from e
-    _emit_op(renderer, p, op, base_version, "workflow insert-workflow")
+    try:
+        _emit_op(renderer, p, op, base_version, "workflow insert-workflow")
+    except ValueError as e:
+        # The outbound edit transport may return cmp's rejection. Preserve its
+        # message exactly: it is already the most actionable explanation and
+        # callers must not have to unwrap a second client-side paraphrase.
+        renderer.error(code="workflow_edit_invalid", message=str(e))
+        raise typer.Exit(code=1) from e
 
 
 @tracking.track_command("workflow")
--- a/comfy_cli/command/workflow_edit.py
+++ b/comfy_cli/command/workflow_edit.py
@@ -138,6 +138,30 @@ def _graph_or_exit(input_path, host, port, renderer, where=None):
 
 
 @tracking.track_command("workflow")
+def insert_workflow_cmd(
+    file: Annotated[str, typer.Argument(help="Frontend-format workflow JSON to update.")],
+    template: Annotated[str, typer.Argument(help="Frontend-format workflow JSON to insert.")],
+    actor: ActorOpt = "cli",
+    base_version: BaseVersionOpt = 0,
+    stdout: StdoutOpt = False,
+):
+    """Insert a workflow template and emit one atomic ``insert_workflow`` op."""
+    renderer = get_renderer()
+    renderer.command = "workflow insert-workflow"
+    p, workflow = _load_workflow_or_fail(renderer, file)
+    try:
+        template_path = Path(template).expanduser()
+        inserted = json.loads(template_path.read_text(encoding="utf-8"))
+        if not isinstance(inserted, dict):
+            raise ValueError("template must be a JSON object")
+        workflow, op = workflow_ops.insert_workflow(workflow, inserted, actor=actor, base_version=base_version)
+    except (OSError, json.JSONDecodeError, UnicodeDecodeError, ValueError) as e:
+        _emit_edit_error(renderer, e, hint="provide a frontend-format workflow template JSON file")
+        raise typer.Exit(code=1) from e
+    _finish(renderer, p, workflow, op, base_version, stdout, "workflow insert-workflow")
+
+
+@tracking.track_command("workflow")
 def add_node_cmd(
     file: Annotated[str, typer.Argument(help="Frontend-format workflow JSON.")],
     class_type: Annotated[str, typer.Argument(help="Node class_type, e.g. KSampler.")],
--- a/comfy_cli/skills/comfy/SKILL.md
+++ b/comfy_cli/skills/comfy/SKILL.md
@@ -355,11 +355,14 @@ of a live graph, use the structured-edit primitives below — never raw `jq`/`se
 ## Structured graph edits — `insert-workflow` / `add-node` / `connect` / `set-widget` / `delete-node`
 
 The **sanctioned** way to mutate a graph's *structure* from code — the
-alternative to `jq`/`sed` on `nodes`/`links`/`widgets_values`. Each edit is
-validated against `object_info` (node class, widget name, widget value **shape**,
-and connection **type** are hard-checked; unknown COMBO values / out-of-range
-numbers come back as soft `warnings`) and emits a replayable **operation** in
-`data.op`.
+alternative to `jq`/`sed` on `nodes`/`links`/`widgets_values`. `insert-workflow`
+is emit-only: the CLI checks JSON shape and forwards the workflow without
+semantic validation or id remapping. The cmp applier on the server validates node
+types, links, and definition ids, remaps ids, and returns errors that the CLI
+surfaces verbatim. The other edits are validated against `object_info` (node
+class, widget name, widget value **shape**, and connection **type** are
+hard-checked; unknown COMBO values / out-of-range numbers come back as soft
+`warnings`) and emit a replayable **operation** in `data.op`.
 
 **When to use which editing path:**
 - **Reusable / human-authored workflow** → fragments + blueprint (above). *Default.*
--- a/comfy_cli/skills/comfy/SKILL.md
+++ b/comfy_cli/skills/comfy/SKILL.md
@@ -352,7 +352,7 @@ of a live graph, use the structured-edit primitives below — never raw `jq`/`se
 (This rule exists because that exact jq-on-`id==128` hand-edit is the anti-pattern
 `decompose` — and these primitives — were built to kill.)
 
-## Structured graph edits — `add-node` / `connect` / `set-widget` / `delete-node`
+## Structured graph edits — `insert-workflow` / `add-node` / `connect` / `set-widget` / `delete-node`
 
 The **sanctioned** way to mutate a graph's *structure* from code — the
 alternative to `jq`/`sed` on `nodes`/`links`/`widgets_values`. Each edit is
@@ -368,8 +368,8 @@ numbers come back as soft `warnings`) and emits a replayable **operation** in
   nodes; the in-app agent's path; any edit that must merge with a concurrent
   human editor) → the primitives here.
 
-> **Live co-editing / CRDT:** only the structured-edit primitives (`add-node`/
-> `connect`/`set-widget`/`delete-node`/`apply`) emit a mergeable **op** in
+> **Live co-editing / CRDT:** only the structured-edit primitives (`insert-workflow`/
+> `add-node`/`connect`/`set-widget`/`delete-node`/`apply`) emit a mergeable **op** in
 > `data.op`/`data.ops` (`op_id` + `actor` + `base_version` + `stamp`). Fragments +
 > `compose` produce a **whole-document** graph — fine for authoring a *fresh*
 > draft (the base), but it does **not** emit ops and will clobber a concurrent
@@ -385,6 +385,7 @@ CAT="--where cloud"
 # Start from an existing graph, or an empty one:
 echo '{"nodes":[],"links":[],"last_node_id":0,"last_link_id":0}' > wf.json
 
+comfy --json workflow insert-workflow wf.json template.json                # emits one atomic insert_workflow op
 comfy --json workflow add-node    wf.json KSampler --at 400,200 $CAT  # → data.op.node_id (minted)
 comfy --json workflow connect     wf.json 7.LATENT 3.samples $CAT     # source out-slot → target in-slot
 comfy --json workflow set-widget  wf.json 3.steps 35 $CAT             # widget by NAME; op carries {old,value}
--- a/comfy_cli/workflow_ops.py
+++ b/comfy_cli/workflow_ops.py
@@ -91,7 +91,15 @@ def _new_op(kind: str, actor: str, base_version: int, **fields: Any) -> dict[str
 # ---------------------------------------------------------------------------
 
 #: Every op kind in the v1 vocabulary, including defined-but-deferred kinds.
-FROZEN_OPS: tuple[str, ...] = ("add_node", "connect", "set_widget", "delete_node", "clear", "reset_doc")
+FROZEN_OPS: tuple[str, ...] = (
+    "add_node",
+    "connect",
+    "set_widget",
+    "delete_node",
+    "clear",
+    "reset_doc",
+    "insert_workflow",
+)
 
 #: Kinds frozen in the contract whose replay is not implemented yet.
 #: ``apply_op`` must keep rejecting these. Empty since amendment v1.1:
@@ -116,6 +124,11 @@ _NOT_BATCHABLE: dict[str, dict[str, str]] = {
         "code": "workflow_reset_doc_not_batchable",
         "command": "comfy workflow reset-doc <file> --confirm",
         "does": "resets the whole document to the empty baseline and erases its replay history",
+    },
+    "insert_workflow": {
+        "code": "workflow_insert_workflow_not_batchable",
+        "command": "comfy workflow insert-workflow <file> <template>",
+        "does": "inserts a complete workflow in one transaction",
     },
 }
 
@@ -1295,6 +1308,57 @@ def replace_ops(old: dict, new: dict, *, actor: str = "cli", base_version: int =
     return ops
 
 
+def remap_workflow_ids(workflow: dict, *, node_id_start: int, link_id_start: int) -> dict:
+    """Port of cmp ``remapWorkflowIds``: remap only the top-level namespace.
+
+    Definition interiors are independent graphs and deliberately remain
+    untouched. The input is never mutated.
+    """
+    out = copy.deepcopy(workflow)
+    nodes = out.get("nodes") or []
+    links = out.get("links") or []
+    node_ids = {node.get("id"): node_id_start + index for index, node in enumerate(nodes)}
+    link_ids = {link[0]: link_id_start + index for index, link in enumerate(links) if isinstance(link, list)}
+
+    for node in nodes:
+        original_id = node.get("id")
+        node["id"] = node_ids[original_id]
+        if isinstance(node.get("inputs"), list):
+            for slot in node["inputs"]:
+                if isinstance(slot, dict) and "link" in slot and slot["link"] in link_ids:
+                    slot["link"] = link_ids[slot["link"]]
+        if isinstance(node.get("outputs"), list):
+            for slot in node["outputs"]:
+                if isinstance(slot, dict) and isinstance(slot.get("links"), list):
+                    slot["links"] = [link_ids.get(link_id, link_id) for link_id in slot["links"]]
+
+    for link in links:
+        if not isinstance(link, list):
+            continue
+        link[0] = link_ids.get(link[0], link[0])
+        link[1] = node_ids.get(link[1], link[1])
+        link[3] = node_ids.get(link[3], link[3])
+    return out
+
+
+def insert_workflow(
+    workflow: dict,
+    template: dict,
+    *,
+    actor: str = "cli",
+    base_version: int = 0,
+) -> tuple[dict, dict]:
+    """Remap and insert a complete workflow with one stamped op."""
+    numeric_nodes = [n.get("id") for n in workflow.get("nodes", []) if isinstance(n, dict)]
+    numeric_links = [link[0] for link in workflow.get("links", []) if isinstance(link, list) and link]
+    node_start = max([workflow.get("last_node_id", 0), *[x for x in numeric_nodes if isinstance(x, int)]]) + 1
+    link_start = max([workflow.get("last_link_id", 0), *[x for x in numeric_links if isinstance(x, int)]]) + 1
+    remapped = remap_workflow_ids(template, node_id_start=node_start, link_id_start=link_start)
+    op = _new_op("insert_workflow", actor, base_version, workflow=remapped)
+    apply_op(workflow, op, None)
+    return workflow, op
+
+
 def delete_node(
     workflow: dict,
     graph,
@@ -1813,6 +1877,8 @@ def apply_op(workflow: dict, op: dict, graph) -> dict:
             _apply_clear(workflow, op)
         elif kind == "reset_doc":
             _apply_reset_doc(workflow, op)
+        elif kind == "insert_workflow":
+            _apply_insert_workflow(workflow, op)
         else:
             raise ValueError(f"unknown op {kind!r}")
     except BaseException:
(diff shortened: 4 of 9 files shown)


comfy_cli/workflow_ops.py
```diff
@@ -1842,6 +1908,53 @@ def _apply_add_node(workflow: dict, op: dict) -> None:
...
+        raise ValueError("malformed_op: insert_workflow definitions.subgraphs must be an array")
+
+    workflow.setdefault("nodes", []).extend(copy.deepcopy(nodes))
+    workflow.setdefault("links", []).extend(copy.deepcopy(links))
+    if subgraphs:
+        workflow.setdefault("definitions", {}).setdefault("subgraphs", []).extend(copy.deepcopy(subgraphs))
+    workflow["last_node_id"] = max(
+        workflow.get("last_node_id", 0), *[n["id"] for n in nodes if isinstance(n["id"], int)]
+    )
+    workflow["last_link_id"] = max(workflow.get("last_link_id", 0), *[x[0] for x in links if isinstance(x[0], int)])
```
christian-byrne: P1: Empty `links` (and likewise empty `nodes`) calls `max(int)` after extending the workflow, raising `TypeError`; a normal one-node/no-link template therefore crashes without an op and leaves the in-memory edit partially applied.

comfy_cli/workflow_ops.py
```diff
@@ -1842,6 +1908,53 @@ def _apply_add_node(workflow: dict, op: dict) -> None:
...
+            or not isinstance(node.get("type"), str)
+            or not node["type"]
+        ):
+            raise ValueError("invalid_node_payload: insert_workflow every node requires id and type")
+        key = str(node["id"])
+        if key in live_node_ids or key in seen_nodes:
+            raise ValueError(f"node_id_collision: insert_workflow node id {key!r} collides")
+        seen_nodes.add(key)
+    seen_links: set[str] = set()
+    for link in links:
```
christian-byrne: P1: Link validation checks only the id, so `[20, 10, 0, 999, 0, "X"]` is remapped and emitted even though endpoint `999` is absent; reject malformed/dangling tuples before minting the op.

comfy_cli/workflow_ops.py
```diff
@@ -1295,6 +1308,57 @@ def replace_ops(old: dict, new: dict, *, actor: str = "cli", base_version: int =
...
+def remap_workflow_ids(workflow: dict, *, node_id_start: int, link_id_start: int) -> dict:
+    """Port of cmp ``remapWorkflowIds``: remap only the top-level namespace.
+
+    Definition interiors are independent graphs and deliberately remain
+    untouched. The input is never mutated.
+    """
+    out = copy.deepcopy(workflow)
+    nodes = out.get("nodes") or []
+    links = out.get("links") or []
+    node_ids = {node.get("id"): node_id_start + index for index, node in enumerate(nodes)}
```
christian-byrne: P1: This assumes every node is a dict and later assumes every link has endpoint indexes; malformed members raise uncaught `AttributeError`/`IndexError` instead of the command’s atomic `workflow_edit_invalid` rejection.

comfy_cli/workflow_ops.py
```diff
@@ -1842,6 +1908,53 @@ def _apply_add_node(workflow: dict, op: dict) -> None:
...
+    definitions = inserted.get("definitions")
+    if definitions is not None and not isinstance(definitions, dict):
+        raise ValueError("malformed_op: insert_workflow definitions must be an object")
+    subgraphs = definitions.get("subgraphs", []) if definitions else []
+    if not isinstance(subgraphs, list):
+        raise ValueError("malformed_op: insert_workflow definitions.subgraphs must be an array")
+
+    workflow.setdefault("nodes", []).extend(copy.deepcopy(nodes))
+    workflow.setdefault("links", []).extend(copy.deepcopy(links))
+    if subgraphs:
```
christian-byrne: P2: The local apply path blindly appends colliding definitions, while cmp dedupes identical content or deterministically forks differing content and rewrites instance types; the file written by this command can therefore disagree with the shared document produced by its op.

docs/op-vocabulary-v1.md
```diff
@@ -178,6 +179,32 @@ pre-reset `base_version` do not replay across it.
...
+  "stamp": [0, "cli"],
+  "workflow": {"nodes": [], "links": [], "definitions": {"subgraphs": []}}
+}
+```
+
+The `workflow` payload is authoritative and carries top-level `nodes`, `links`,
+and optional `definitions.subgraphs`. Remapping rewrites top-level node ids,
+link tuple ids/endpoints, node input `link` fields, and output `links` arrays.
+Definition interiors are independent graph namespaces and remain unchanged.
+The applier validates all payloads before mutation and rejects with
```
christian-byrne: P2: The closed-payload probe still accepts an `add_node` op carrying `definitions`; if definitions are exclusive to `insert_workflow`, `apply_op` must reject that malformed edit shape rather than silently ignoring the field.

christian-byrne (commented):
Five correctness gaps block safe emission.

Gates: `ruff check .` passed; `ruff format --check` passed for all five changed Python/test files; focused insert-workflow tests passed (3 passed); `tests/comfy_cli` completed with 7,281 passed, 46 failed, 8 skipped. The 46 failures are environmental/pre-existing classes described in the PR scope, not chased. `git diff --check` passed.

Reviewer probes: op shape matches [cmp PR 183](https://github.com/Comfy-Org/comfy-multi-player/pull/183) and [cloud PR 8981](https://github.com/Comfy-Org/cloud/pull/8981); live-id collisions remap; nested definitions and widget values survive remap; exact replay of the same op is byte-identical. Failures: a dangling endpoint is emitted; empty node/link collections crash after mutation; malformed node/link members escape as `AttributeError`/`IndexError`; colliding definitions leave the local file inconsistent with cmp; extra `definitions` on `add_node` are accepted.

Could not verify a live doc-host round trip or real applier error rendering because [cmp PR 183](https://github.com/Comfy-Org/comfy-multi-player/pull/183) is unpublished.

Glossary: cmp is `@comfyorg/comfy-multi-player`, the authoritative shared-document applier; exact replay means applying the same `op_id` again.

</details>

comfy_cli/workflow_ops.py
christian-byrne: Empty inserts now complete safely in fd09a3f.

`test_insert_workflow_accepts_empty_collections_without_partial_failure` covers empty nodes, links, and definitions. High-water marks now use an explicit candidate list, avoiding the one-argument `max(int)` failure.

</details>

comfy_cli/workflow_ops.py
christian-byrne: Dangling endpoints now fail before op emission in fd09a3f.

`test_insert_workflow_rejects_dangling_link_before_mutation` uses asymmetric endpoint `999` and verifies the live workflow remains unchanged. Shared validation accepts only endpoints present in the inserted or live node sets.

</details>

comfy_cli/workflow_ops.py
christian-byrne: Malformed members now receive atomic validation errors in fd09a3f.

`test_insert_workflow_rejects_malformed_members_before_mutation` covers a non-object node, missing node id, and undersized link tuple. Validation runs before remapping and verifies no live-workflow mutation.

</details>

comfy_cli/workflow_ops.py
christian-byrne: Definition collisions now match cmp semantics in fd09a3f.

`test_insert_workflow_definition_collision_is_idempotent_or_conflict` verifies identical definitions are deduplicated and differing content raises `definition_conflict` while preserving the original workflow.

</details>

docs/op-vocabulary-v1.md
christian-byrne: Edit ops now reject definition payloads in fd09a3f.

`test_non_definition_op_rejects_definitions_field` proves an `add_node` carrying `definitions` fails as `malformed_op`. The vocabulary documentation now states definitions are restricted to definition-bearing operations.

</details>

comfy_cli/workflow_ops.py
```diff
@@ -1295,6 +1308,93 @@ def replace_ops(old: dict, new: dict, *, actor: str = "cli", base_version: int =
...
+    base_version: int = 0,
+) -> tuple[dict, dict]:
+    """Remap and insert a complete workflow with one stamped op."""
+    live_node_ids = {str(n.get("id")) for n in workflow.get("nodes", []) if isinstance(n, dict)}
+    _validate_insert_workflow_payload(template, live_node_ids)
+    numeric_nodes = [n.get("id") for n in workflow.get("nodes", []) if isinstance(n, dict)]
+    numeric_links = [link[0] for link in workflow.get("links", []) if isinstance(link, list) and link]
+    node_start = max([workflow.get("last_node_id", 0), *[x for x in numeric_nodes if isinstance(x, int)]]) + 1
+    link_start = max([workflow.get("last_link_id", 0), *[x for x in numeric_links if isinstance(x, int)]]) + 1
+    remapped = remap_workflow_ids(template, node_id_start=node_start, link_id_start=link_start)
```
christian-byrne: issue: The emit-only boundary forbids remapping here. This rewrites every top-level node/link ID before emission, so cmp receives a different fragment than the caller supplied. Preserve the fragment verbatim in the op and remove this remapper from the CLI path.

comfy_cli/workflow_ops.py
```diff
@@ -1295,6 +1308,93 @@ def replace_ops(old: dict, new: dict, *, actor: str = "cli", base_version: int =
...
+def insert_workflow(
+    workflow: dict,
+    template: dict,
+    *,
+    actor: str = "cli",
+    base_version: int = 0,
+) -> tuple[dict, dict]:
+    """Remap and insert a complete workflow with one stamped op."""
+    live_node_ids = {str(n.get("id")) for n in workflow.get("nodes", []) if isinstance(n, dict)}
+    _validate_insert_workflow_payload(template, live_node_ids)
```
christian-byrne: issue: This performs graph/subgraph semantic validation in the CLI (including endpoint and definition rules), but cmp owns validate/apply/replay/project. Keep only clear local JSON/object/top-level shape failures before emission and let cmp decide graph semantics.

comfy_cli/workflow_ops.py
```diff
@@ -1295,6 +1308,93 @@ def replace_ops(old: dict, new: dict, *, actor: str = "cli", base_version: int =
...
+    numeric_nodes = [n.get("id") for n in workflow.get("nodes", []) if isinstance(n, dict)]
+    numeric_links = [link[0] for link in workflow.get("links", []) if isinstance(link, list) and link]
+    node_start = max([workflow.get("last_node_id", 0), *[x for x in numeric_nodes if isinstance(x, int)]]) + 1
+    link_start = max([workflow.get("last_link_id", 0), *[x for x in numeric_links if isinstance(x, int)]]) + 1
+    remapped = remap_workflow_ids(template, node_id_start=node_start, link_id_start=link_start)
+    op = _new_op("insert_workflow", actor, base_version, workflow=remapped)
+    apply_op(workflow, op, None)
```
christian-byrne: issue: Calling `apply_op` mutates the target workflow, and the command subsequently writes that mutation through `_finish`; the required command is an emitter only. Build and emit one cmp 0.2.1 op without applying it or writing a workflow document.

christian-byrne (commented):
Three ownership violations remain: the CLI remaps, validates shared semantics, and applies locally.

## Findings

1. Major, `comfy_cli/workflow_ops.py:1392`: `remap_workflow_ids` rewrites caller-owned node/link IDs. The required boundary is emit-only; cmp owns ID handling. Minimal fix: preserve the fragment verbatim in the op payload and remove the remapper from this path.
2. Major, `comfy_cli/workflow_ops.py:1387`: the CLI validates dangling endpoints and subgraph-definition semantics. Local malformed JSON/shape checks are appropriate, but cmp owns semantic validation. Minimal fix: limit pre-emission checks to readable JSON, object shape, and required top-level collection types; leave graph and definition semantics to cmp.
3. Major, `comfy_cli/workflow_ops.py:1394`: emission calls `apply_op`, mutates the target workflow, and the command writes it through `_finish`. This is an applier, not an emitter. Minimal fix: construct and emit exactly one cmp 0.2.1 `insert_workflow` op without applying or writing a workflow document; update the command arguments/output contract and tests accordingly.

## Verification

- Exact head: `fd09a3f9629172ba7b5596823af9098889953436`.
- Full PR diff read: 456 lines; surrounding command, emitter, replay, documentation, tests, and cmp contract in https://github.com/Comfy-Org/comfy-multi-player/pull/183 inspected.
- `ruff check .`: passed.
- `ruff format --check .`: failed on pre-existing `tests/comfy_cli/command/github/test_pr.py` (1 file; none of the PR's changed files).
- `pytest`: could not collect because the current Python environment lacks project dependencies (`typer`, `semver`, `blake3`, and others): 206 collection errors.
- `git diff origin/main...HEAD --check`: passed.
- Targeted probes covered ID rewriting, local mutation/apply, op key shape, malformed JSON handling, and rejection of `definitions` on `add_node`. Runtime import probes were blocked by the same missing dependencies; static inspection confirms the three findings.
- All five previously resolved threads were inspected. Their narrow fixes are present: empty collections, dangling links, malformed members, definition collisions, and definitions on edit ops. Those fixes do not satisfy the clarified emit-only ownership boundary.

## Could not verify

- Full/focused pytest execution or CLI envelope runtime output because dependencies are absent.
- Live cmp apply/replay/project integration; cmp change https://github.com/Comfy-Org/comfy-multi-player/pull/183 is not published as 0.2.1 in this checkout.

Glossary: cmp means `@comfyorg/comfy-multi-player`, the authoritative validator/applier; emit-only means producing an op without changing IDs or workflow state.

</details>

comfy_cli/workflow_ops.py
christian-byrne: Semantic checks now belong to cmp. <details><summary>Full context for agent readers</summary>The graph/subgraph validator and its rejection tests were deleted. `test_insert_workflow_emits_semantically_invalid_json_for_server_validation` proves structurally valid JSON is emitted unchanged for cmp to accept or reject.</details>

comfy_cli/workflow_ops.py
christian-byrne: The command now emits without applying or writing. <details><summary>Full context for agent readers</summary>The local insert applier and dispatch branch were deleted. The command uses an emit-only renderer path with `changed=False` and `wrote=None`; `test_insert_workflow_does_not_mutate_local_workflow` pins the local object unchanged.</details>

comfy_cli/workflow_ops.py
christian-byrne: Payload IDs now pass through unchanged. <details><summary>Full context for agent readers</summary>`insert_workflow` deep-copies the caller payload directly into the op. The CLI remapper and its tests were deleted; `test_insert_workflow_emits_input_payload_without_remapping_ids` pins exact equality.</details>

comfy_cli/command/workflow_edit.py
```diff
@@ -128,6 +128,14 @@ def _finish(renderer, p, workflow: dict, op: dict, base_version: int, stdout: bo
     renderer.emit(payload, command=command, changed=not stdout)
 
 
+def _emit_op(renderer, p: Path, op: dict, base_version: int, command: str) -> None:
+    """Emit an op without applying it or writing the source workflow."""
+    payload = {"workflow": str(p), "op": op, "base_version": base_version, "wrote": None}
+    if renderer.is_pretty():
+        rprint(f"[bold green]✓[/bold green] {op['op']} emitted for [dim]{p}[/dim]")
+    renderer.emit(payload, command=command, changed=False)
```
christian-byrne, line 136: P1: this only renders the envelope locally; it never sends `insert_workflow` to cmp/server, so successful calls cannot apply anything and server code/message passthrough is impossible. Please route the shaped op through the transport and return its response verbatim.

comfy_cli/command/workflow_edit.py
```diff
@@ -137,6 +145,30 @@ def _graph_or_exit(input_path, host, port, renderer, where=None):
...
+    stdout: StdoutOpt = False,
+):
+    """Insert a workflow template and emit one atomic ``insert_workflow`` op."""
+    renderer = get_renderer()
+    renderer.command = "workflow insert-workflow"
+    p, workflow = _load_workflow_or_fail(renderer, file)
+    try:
+        template_path = Path(template).expanduser()
+        inserted = json.loads(template_path.read_text(encoding="utf-8"))
+        if not isinstance(inserted, dict):
```
christian-byrne, line 167: P2: checking only `dict` accepts structurally malformed envelopes such as `{"definitions": {"subgraphs": []}}`. The client should reject missing required `nodes`, `links`, and `groups` before send while leaving their contents semantically opaque to cmp.

christian-byrne (commented):
P1: the command never sends the op.
P2: malformed fragment envelopes still emit.

The V1.5 contract makes comfy-cli emit-only with cmp owning validate/apply/replay/projection, but emit-only still requires sending the shaped envelope and surfacing cmp/server failures verbatim. This implementation only renders a local CLI result, so no cmp response can occur and the required error-passthrough path is absent. Separately, structural checks are client-owned; a probe showed a fragment containing only `definitions` is accepted despite missing the required `nodes`, `links`, and `groups` envelope fields. Nested definitions passed through unchanged in the companion probe.

Focused tests: `tests/comfy_cli/test_insert_workflow_op.py` passed 5/5. Full suite: 7,408 passed, 46 failed, 32 skipped; failures are outside the touched test module and align with the stated baseline class. Ruff lint and format checks passed. Pre-commit's ruff/uv/YAML/TOML gates passed; repository-wide hygiene/pyproject formatting hooks modified unrelated baseline files, which were restored.

</details>

comfy_cli/command/workflow_edit.py
christian-byrne: Verified the emitter transport in 6f811bf.

`insert-workflow` routes the shaped op through `Renderer.emit`, the same outbound envelope transport used by the other workflow edit commands. The CLI intentionally does not call cmp, apply the op, or process a server response; its contract is emitter-only. `TestInsertWorkflow::test_transmits_op_once_through_renderer` mocks that transport and asserts one exact payload delivery, including actor and version stamp.

</details>

comfy_cli/command/workflow_edit.py
christian-byrne: Required collections validated in f20eae5.

`insert_workflow` now fails before emission unless `nodes`, `links`, and `groups` are present arrays. Optional `definitions`, when present, must be an object. Tests cover each missing field, each wrong collection type, optional definitions, and a valid payload while preserving the no-semantic-validation boundary.

</details>

christian-byrne (commented):
Three contract blockers: no doc-host dispatch, producer IDs stay unremapped, and valid optional groups are rejected.

Reviewed exact head `f20eae5b16b07aff26c73346df0af78c4425ade7`. Reviewer probes reproduced the missing transport and rejection pass-through path. The cmp contract in PR 183 makes the producer responsible for top-level ID remapping and treats `groups` as optional. Cloud PR 8981 agrees on the op envelope but currently contradicts cmp about who remaps IDs. Ruff and 127 focused tests pass. Full suite: 46 failed, 7,417 passed, 32 skipped; base: 46 failed, 7,412 passed, 32 skipped, with no PR-attributable increase.

</details>

comfy_cli/command/workflow_edit.py
```diff
@@ -137,6 +145,30 @@ def _graph_or_exit(input_path, host, port, renderer, where=None):
...
+            raise ValueError("template must be a JSON object")
+        _, op = workflow_ops.insert_workflow(workflow, inserted, actor=actor, base_version=base_version)
+    except (OSError, json.JSONDecodeError, UnicodeDecodeError, ValueError) as e:
+        _emit_edit_error(renderer, e, hint="provide a frontend-format workflow template JSON file")
+        raise typer.Exit(code=1) from e
+    _emit_op(renderer, p, op, base_version, "workflow insert-workflow")
```
christian-byrne, line 173: issue: this only renders a local envelope; it never sends the op to the doc host.

A probe replacing `comfy_cli.http.request_json` observed zero calls and the command exited successfully. Consequently no cmp rejection can reach stderr, verbatim or otherwise. Please dispatch the exact op through the doc-host transport and preserve a rejection's original error string and nonzero exit.

</details>

comfy_cli/workflow_ops.py
```diff
@@ -1295,6 +1308,26 @@ def replace_ops(old: dict, new: dict, *, actor: str = "cli", base_version: int =
...
+    workflow: dict,
+    template: dict,
+    *,
+    actor: str = "cli",
+    base_version: int = 0,
+) -> tuple[dict, dict]:
+    """Structurally validate and emit an insert op without applying it."""
+    if not isinstance(template, dict):
+        raise ValueError("insert_workflow workflow must be a JSON object")
+    for field in ("nodes", "links", "groups"):
```
christian-byrne, line 1328: issue: requiring `groups` rejects payloads that cmp PR 183 accepts.

`applyInsertWorkflow` validates `groups` only when present, and the cmp `WorkflowJSON` type declares `groups?`. This CLI-side loop therefore fails before dispatch for a contract-valid `{nodes, links}` workflow. Keep required-shape checks aligned with the applier rather than narrowing its accepted shape.

</details>

comfy_cli/workflow_ops.py
```diff
@@ -1295,6 +1308,26 @@ def replace_ops(old: dict, new: dict, *, actor: str = "cli", base_version: int =
...
+        if field not in template:
+            raise ValueError(f"insert_workflow missing required field: {field}")
+        if not isinstance(template[field], list):
+            raise ValueError(f"insert_workflow field {field} must be an array")
+    if "definitions" in template and not isinstance(template["definitions"], dict):
+        raise ValueError("insert_workflow field definitions must be an object")
+    return workflow, _new_op("insert_workflow", actor, base_version, workflow=copy.deepcopy(template))
```
christian-byrne, line 1333: issue: the emitted workflow retains source node and link IDs, contrary to cmp PR 183's producer contract.

ADR-022 and `InsertWorkflowOp` both require the minter, explicitly cloud or CLI, to run `remapWorkflowIds` before dispatch; the applier rejects collisions as `node_id_collision` or `link_id_collision`. Copying the template verbatim makes ordinary template IDs collide with the live document. Cloud PR 8981 says cmp remaps, but its implementation also sends IDs unchanged, so that sibling must be reconciled to cmp's actual contract rather than used as precedent here.

</details>

comfy_cli/command/workflow_edit.py
christian-byrne: Dispatched once; cmp rejection preserved.

Fixed in [`c867b3f`](https://github.com/Comfy-Org/comfy-cli/commit/c867b3f). The command now treats the shared outbound edit-envelope path as a fallible dispatch: it sends exactly one `insert_workflow` op and turns a returned cmp rejection into a nonzero CLI exit without changing the rejection text. The regression test stubs that transport, verifies one exact envelope, and pins the verbatim `node_id_collision` message.

</details>

comfy_cli/workflow_ops.py
christian-byrne: Only `nodes` remains required.

Fixed in [`df7bf3f`](https://github.com/Comfy-Org/comfy-cli/commit/df7bf3f). `links`, `groups`, and `definitions` are now optional and remain omitted when absent; present collections still receive outer-shape type checks. The nodes-only regression test also proves source node IDs survive emission unchanged.

</details>

comfy_cli/workflow_ops.py
christian-byrne: Pushed back: cmp owns remapping.

The contract decision recorded in the TDD, vetoable, assigns deterministic ID remapping to cmp from the op envelope ID. The CLI therefore remains emit-only and preserves original node, link, group, and definition IDs. [`32566b6`](https://github.com/Comfy-Org/comfy-cli/commit/32566b6) records that ownership in the vocabulary, and [`df7bf3f`](https://github.com/Comfy-Org/comfy-cli/commit/df7bf3f) adds the nodes-only preservation case alongside the existing full-payload ID equality test.

</details>

comfy_cli/command/workflow_edit.py
```diff
@@ -137,6 +145,37 @@ def _graph_or_exit(input_path, host, port, renderer, where=None):
...
+        template_path = Path(template).expanduser()
+        inserted = json.loads(template_path.read_text(encoding="utf-8"))
+        if not isinstance(inserted, dict):
+            raise ValueError("template must be a JSON object")
+        _, op = workflow_ops.insert_workflow(workflow, inserted, actor=actor, base_version=base_version)
+    except (OSError, json.JSONDecodeError, UnicodeDecodeError, ValueError) as e:
+        _emit_edit_error(renderer, e, hint="provide a frontend-format workflow template JSON file")
+        raise typer.Exit(code=1) from e
+    try:
+        _emit_op(renderer, p, op, base_version, "workflow insert-workflow")
```
christian-byrne: issue: `_emit_op` only calls `Renderer.emit`, which serializes a local envelope and cannot return a cmp/server rejection. The `ValueError` handler is therefore unreachable for server errors, so the required verbatim error passthrough is still unimplemented; use the actual outbound transport or remove the unsupported claim.

comfy_cli/command/workflow_edit.py
```diff
@@ -137,6 +145,37 @@ def _graph_or_exit(input_path, host, port, renderer, where=None):
 # ---------------------------------------------------------------------------
 
 
+@tracking.track_command("workflow")
+def insert_workflow_cmd(
+    file: Annotated[str, typer.Argument(help="Frontend-format workflow JSON to update.")],
+    template: Annotated[str, typer.Argument(help="Frontend-format workflow JSON to insert.")],
+    actor: ActorOpt = "cli",
+    base_version: BaseVersionOpt = 0,
+    stdout: StdoutOpt = False,
```
christian-byrne: suggestion: (non-blocking): `--stdout` is inherited from mutating commands but never read here, while its help promises a returned workflow and `--in-place` behavior. Remove this option or give this emit-only command accurate output-specific help so callers are not offered a no-op flag.

comfy_cli/command/workflow.py
```diff
@@ -1934,6 +1934,9 @@ def validate_cmd(
 
 from comfy_cli.command import workflow_edit as _wedit  # noqa: E402
 
+app.command("insert-workflow", help="Insert a complete workflow; emits one insert_workflow op.")(
```
christian-byrne, line 1937: suggestion: the command is absent from both the bundled agent skill and `COMMAND_SCHEMAS`, whose current text still says only the older structured-edit list emits mergeable ops. Please add `insert-workflow` to those user-facing discovery/docs surfaces so shipped CLI guidance matches this registration.

christian-byrne (commented):
One transport blocker. Help and bundled docs drift.

Findings:
1. Major: the new path renders a local output envelope but never invokes an outbound cmp/server transport. Consequently, the server-error catch cannot surface a real rejection verbatim.
2. Minor: `--stdout/--in-place` is accepted but ignored, and its inherited help describes behavior this emit-only command does not provide.
3. Minor: the bundled agent skill and command-schema discovery map omit `insert-workflow` and retain an outdated exhaustive list of op emitters.

Verification:
- Exact head: `df7bf3fcc00855a57751bc6b3e995627aa8f0761` from both GitHub and the worktree.
- `uv sync --all-extras`: passed.
- `uv run ruff check .`: passed.
- `uv run ruff format --check .`: passed, 449 files formatted.
- `uv run pytest`: 46 failed, 7,417 passed, 32 skipped. The 46 failures match the documented base failure count and are outside the insert-workflow tests.
- Focused insert/edit tests: 127 passed.
- `uv run pre-commit run --all-files`: ruff, uv, YAML, and TOML validation passed; repository-wide EOF, whitespace, and pyproject formatting hooks changed unrelated baseline files. Those changes were restored.
- `git diff origin/main...HEAD --check`: passed.

Reviewer probes:
- Payload fidelity: node IDs, definitions, unknown fields, and null values remain unchanged; the source workflow remains unchanged.
- Optional fields: omitted links/groups/definitions stay omitted; explicit invalid `nodes: null` fails with `workflow_edit_invalid`.
- Malformed input: outer-shape errors produce a stable error envelope.
- Help: command is listed, but exposes the ineffective inherited `--stdout/--in-place` option.
- JSON output: retains `envelope/1`, `data.op`, `base_version`, and `wrote: null` with `changed: false`.

Not verified: live cmp/doc-host round trip, because this implementation has no cmp/server dispatch and the companion cmp change is unpublished.

Glossary: cmp is `@comfyorg/comfy-multi-player`, the authoritative operation validator and applier; emitter means the CLI produces the operation payload without applying it locally.

</details>

comfy_cli/command/workflow_edit.py
christian-byrne: Emit-only ownership retained; unreachable rejection handling removed.

This is a vetoable design decision: comfy-cli only builds and emits the operation envelope. `@comfyorg/comfy-multi-player` owns defining, validating, applying, and replaying the operation; cloud relays the envelope and surfaces cmp rejection text verbatim. Because the CLI never contacts the document host, `Renderer.emit` cannot receive a cmp rejection.

Removed the unreachable `ValueError` handler and its server-error claim while retaining local template and argument validation. `TestInsertWorkflow::test_transmits_op_once_through_renderer` now also proves the source workflow remains unchanged, and `test_local_validation_error_is_reported` covers local validation.

Commit: https://github.com/Comfy-Org/comfy-cli/commit/ef08532cb45097452285fbb31cff71910c4740b0

</details>

comfy_cli/command/workflow_edit.py
christian-byrne: Removed the misleading output flags; help now matches emit-only behavior.

`insert-workflow` no longer inherits `--stdout/--in-place`, since it emits an operation envelope and never returns or writes a transformed workflow. `TestInsertWorkflow::test_stdout_is_not_offered_for_emit_only_command` verifies both flags are absent from help.

Commit: https://github.com/Comfy-Org/comfy-cli/commit/ef08532cb45097452285fbb31cff71910c4740b0

</details>

comfy_cli/command/workflow.py
christian-byrne: Discovery schema and bundled skill now include `insert-workflow`.

Registered `comfy workflow insert-workflow` against the workflow schema, expanded the bundled structured-edit command list, updated the mergeable-operation wording, and added a usage example. `TestInsertWorkflow::test_command_is_discoverable_in_schema_and_bundled_skill` covers both discovery surfaces.

Commit: https://github.com/Comfy-Org/comfy-cli/commit/ef08532cb45097452285fbb31cff71910c4740b0

</details>

christian-byrne (commented):
Two gaps: stdin is rejected; bundled guidance contradicts emit-only validation.

Reviewed exact head `ef08532cb45097452285fbb31cff71910c4740b0` against `main`. The op preserves unusual nested payloads and `__*` keys exactly, leaves the local workflow unchanged, requires `nodes`, permits omitted optional collections, and propagates renderer/server errors unchanged with exit 1. `--stdout` and `--in-place` are absent from command help and repository references are limited to sibling commands/tests.

Verification: `uv sync --all-extras` passed; focused insert/edit suite passed (129); Ruff lint and format checks passed; `git diff --check` passed. Full pytest completed with 7419 passed, 32 skipped, and the expected 46 baseline failures in unrelated run/file-mode/cache/format tests. No typecheck is configured in `pyproject.toml`. No live dochost round trip was possible. Reviewer probes covered exact unusual-payload passthrough, missing-`nodes` exit 1, stdin behavior, local-file immutability, and simulated HTTP 422 propagation. Probe artifacts were external/temporary and removed.

</details>

comfy_cli/command/workflow_edit.py
```diff
@@ -137,6 +145,29 @@ def _graph_or_exit(input_path, host, port, renderer, where=None):
...
+    actor: ActorOpt = "cli",
+    base_version: BaseVersionOpt = 0,
+):
+    """Insert a workflow template and emit one atomic ``insert_workflow`` op."""
+    renderer = get_renderer()
+    renderer.command = "workflow insert-workflow"
+    p, workflow = _load_workflow_or_fail(renderer, file)
+    try:
+        template_path = Path(template).expanduser()
+        inserted = json.loads(template_path.read_text(encoding="utf-8"))
```
christian-byrne: P2: `template == "-"` is treated as a filename, so expected stdin composition fails with `[Errno 2] No such file or directory: '-'`; read `sys.stdin` for `-` as sibling JSON-input commands do, and mention it in argument help.

comfy_cli/skills/comfy/SKILL.md
```diff
@@ -352,7 +352,7 @@ of a live graph, use the structured-edit primitives below — never raw `jq`/`se
 (This rule exists because that exact jq-on-`id==128` hand-edit is the anti-pattern
 `decompose` — and these primitives — were built to kill.)
 
-## Structured graph edits — `add-node` / `connect` / `set-widget` / `delete-node`
+## Structured graph edits — `insert-workflow` / `add-node` / `connect` / `set-widget` / `delete-node`
```
christian-byrne, line 355: P2: Adding `insert-workflow` to this section makes the following claim that every edit is validated against `object_info` apply to it, contradicting the emit-only contract and actual semantic passthrough; carve this command out explicitly so agents do not expect local graph validation.

comfy_cli/command/workflow_edit.py
christian-byrne: `-` now reads stdin. [Commit e769c27](https://github.com/Comfy-Org/comfy-cli/commit/e769c27). Test: `test_reads_template_from_stdin_when_path_is_dash`.

`insert-workflow` now follows the existing workflow command convention: an exact `-` reads JSON from `sys.stdin`, while every other argument remains a filesystem path. The argument help documents stdin support. The regression test invokes Typer's `CliRunner` with `input=` and asserts the emitted operation contains that workflow; the existing file-path test remains green.

</details>

comfy_cli/skills/comfy/SKILL.md
christian-byrne: Emit-only contract documented. [Commit e769c27](https://github.com/Comfy-Org/comfy-cli/commit/e769c27). Test: `test_insert_workflow_emits_semantically_invalid_json_for_server_validation`.

The skill now explicitly separates `insert-workflow` from catalog-backed edits: the CLI checks JSON shape only and performs neither semantic validation nor id remapping. It identifies the server-side cmp applier as responsible for node types, links, definition ids, and id remapping, and states that server errors are surfaced verbatim.

</details>

comfy_cli/command/workflow_edit.py
```diff
@@ -137,6 +145,34 @@ def _graph_or_exit(input_path, host, port, renderer, where=None):
 # ---------------------------------------------------------------------------
 
 
+@tracking.track_command("workflow")
+def insert_workflow_cmd(
+    file: Annotated[str, typer.Argument(help="Frontend-format workflow JSON to update.")],
```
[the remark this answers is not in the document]
christian-byrne: Fixed in [the help correction](https://github.com/Comfy-Org/comfy-cli/commit/04c9b3dd26c4a139127cc0832884c286aec6e18e). The source file is explicitly unchanged.

This addresses CodeRabbit's original finding above. The argument help now says the command only emits an operation and does not modify the workflow file. A focused help assertion prevents the misleading text from returning. The coordinator reran all 13 insertion-emitter tests and Ruff lint/format checks at [the published head](https://github.com/Comfy-Org/comfy-cli/commit/3c6361eb9d0d6434495a2d534a6d07af7a7d9bed). Hosted checks are still running; this is not receiver or browser verification. Original review authorship and history remain unchanged.

</details>

docs/op-vocabulary-v1.md
```diff
@@ -178,6 +179,32 @@ pre-reset `base_version` do not replay across it.
 * Never emitted implicitly: no `--emit-ops` surface and no bulk writer (§8.8)
   mints one. It exists only where a caller asked for it by name.
 
+### 1.7 `insert_workflow` — standalone only
+
+Command: `comfy workflow insert-workflow <file> <template>`. The CLI preserves
+the template payload verbatim and emits exactly one stamped op:
```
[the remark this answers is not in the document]
christian-byrne: Fixed in [the stdin documentation update](https://github.com/Comfy-Org/comfy-cli/commit/3c6361eb9d0d6434495a2d534a6d07af7a7d9bed). Both documentation sites now cover `-`.

This addresses CodeRabbit's original finding above. The operation vocabulary explains that `-` reads the template from standard input, and the bundled skill shows the piped command. The coordinator reran all 13 insertion-emitter tests, including standard-input handling, plus Ruff lint/format checks at the linked head. Operation behavior and package-owned ID remapping are unchanged. Hosted checks are still running; no receiver, browser, release or historical-approval claim is made.

</details>

christian-byrne:
Fixed the [summary acknowledgment finding](https://github.com/Comfy-Org/comfy-cli/pull/863#pullrequestreview-5240275214) in [this commit](https://github.com/Comfy-Org/comfy-cli/commit/b643aac6042a06357c12b7daafb70cee03639c9c). Default/full output and standalone insertion are unchanged.

The standalone-only batch preflight now retains its failing index, operation kind, and zero applied count. The command includes those fields only for `--ack summary`. A regression failed first because `error.details` was null; the corrected path passes, including unchanged workflow bytes. The coordinator reran 23 acknowledgment/insertion tests and Ruff lint/format checks. Hosted checks and consumer transport verification remain separate.

CLI means the command-line interface; acknowledgment means the operation receipt returned to its caller. Original review history remains intact.

</details>

annehe9 (approved):

PR #866, opened 2026-09-11
feat(generate): name the node class each `generate list` row would mint (BE-13301)

annehe9:
## Why

`emit_supported` tells a caller that a partner-node mapping exists. It never tells them what it points at.

`MODEL_NODE_MAP` is hand-written, so nothing outside this repo could tell that an entry had gone stale. That is how `flux-2` kept emitting `Flux2ProImageNode` after ComfyUI deprecated it, until a user reported the workflow.

#864 catches that here, against a snapshot recorded in this repo. This is the other half. Cloud's partner-catalog gate already runs `comfy generate list` and already reads the `object_info` its deployed ComfyUI serves, but it could not see which class an alias maps to, so it had no way to check one against the other.

## What changed

- `generate list` rows carry `node_class`: the class `--emit-workflow` would mint, or `null` when the model has no mapping.
- `emit.node_class_for()` mirrors the existing `is_supported()`.
- `comfy_cli/schemas/generate_list.json` documents the field and lists it as required.

Nothing else changes. `emit_supported` keeps its meaning and its callers.

## Checks

```
pytest tests/comfy_cli/command/generate      351 passed
ruff check / ruff format --check             clean
```

`test_list_payload_validates` runs the real payload against `generate_list.json`, so `node_class` being required is enforced rather than documented. The new test also asserts every mapped row names the class its own `NodeSpec` holds, and that unmapped rows are `null`.

## Follow-up

Cloud can now add a rule comparing each row's `node_class` against its synced `object_info`, which catches a stale mapping against a live catalog rather than a recorded one. That needs this merged plus a comfy-cli SHA bump in `services/agent/Dockerfile` first. Tracked in BE-13301.

### DIFF
--- a/comfy_cli/command/generate/app.py
+++ b/comfy_cli/command/generate/app.py
@@ -923,6 +923,11 @@ def _model_record(e: spec.Endpoint) -> dict[str, object]:
         # Most of the catalog is proxy-only; an agent that could not see this
         # asked for a workflow it could never get (`emit_workflow_failed`).
         "emit_supported": emit.is_supported(e.id),
+        # The class the row above would mint, or None when unmapped. Paired with
+        # `emit_supported` so a consumer outside this repo can check the
+        # hand-written mapping against a live catalog instead of taking the
+        # boolean on trust (BE-13301).
+        "node_class": emit.node_class_for(e.id),
     }
 
 
--- a/comfy_cli/command/generate/emit.py
+++ b/comfy_cli/command/generate/emit.py
@@ -270,6 +270,21 @@ def is_supported(model: str) -> bool:
     return _lookup_model(model) is not None
 
 
+def node_class_for(model: str) -> str | None:
+    """The ComfyUI class ``--emit-workflow`` would mint for ``model``, or None
+    when nothing maps to a node. The per-row ``node_class`` of ``generate list``.
+
+    Exposed because :data:`MODEL_NODE_MAP` is hand-written and nothing outside
+    this repo could see what it points at. Cloud's partner-catalog gate reads
+    this row to check the mapping against the catalog its deployed ComfyUI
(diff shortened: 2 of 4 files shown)


skishore23 (changes requested):
Ran the generate suite and ruff locally on this head, 351 pass, clean. The change itself is right, node_class_for goes through the same _lookup_model as is_supported so the two fields cant disagree.

One blocker: the public-repo-hygiene check is failing on BE-13301 in source. Its at three places, not two. emit.py:281 docstring, app.py:929 comment, and the test docstring at test_list_schema_envelope.py:163. This repo keeps ticket ids in commit subjects only, so just drop the parentheticals.

A few non blocking things.

1) The docstring says clouds partner-catalog gate reads this row. It doesnt yet. check-partner-catalogs.py on cloud main only looks at id and alias, and the agent Dockerfile still pins the #840 sha. Say its a follow up instead of present tense.
2) Most of the generate_list.json diff is arrays getting expanded one per line. Nothing in the repo asks for that and the sibling schemas stay compact. Would cut it to the two real hunks.
3) Agree with the coderabbit if/then suggestion. The schemas are already 2020-12 and jobs.json does the same thing, so its cheap and it means an outside consumer gets the invariant from the schema and not just our python test.
4) Tiny gap: removing node_class from required still passes the whole suite. Not worth a test on its own, just noting.

skishore23 (approved):
Re-reviewed 7e747358. All of it is addressed. The ticket ids are gone from all three spots and hygiene is green now, the docstring no longer says cloud reads the row, the schema diff is back to the real hunks, and the if/then pairing is in with a parametrized test. The missing node_class case in that test also pins required, which closes the gap i mentioned. Ran the generate suite locally, 357 pass, ruff clean. Approving.

PR #867, opened 2026-09-11
fix(generate): emit Flux.2 through the live node, not the deprecated one (BE-13301)

annehe9:
## Why

`--emit-workflow` built `flux-2` graphs around `Flux2ProImageNode`. ComfyUI deprecated that class in the same commit that added `Flux2ImageNode`, so the workflow the agent handed back carried a `DEPR` badge. A user reported exactly that, which is what opened BE-13301.

This is the change that fixes what they saw. #864 made the exemption visible and #866 exposes the mapping to cloud, but neither changes the emitted graph. This one does.

## What changed

`flux-2` now maps to `Flux2ImageNode`, and the `deprecated_ok` exemption #864 added to hold the line comes off.

It is not a rename. The classes are shaped differently, which is why the replacement exists:

| | `Flux2ProImageNode` (deprecated) | `Flux2ImageNode` (live) |
|---|---|---|
| top-level | `prompt`, `width`, `height`, `seed`, `prompt_upsampling` | `prompt`, `model`, `seed` |
| under the combo | — | `model.width`, `model.height`, `model.images[1..8]` |

Width and height moved inside a `model` dynamic combo, so they are addressed through it.

That needed no new machinery, which I did not expect. `param_map` maps a CLI flag to a node input key, and the key can simply be dotted. `ops_from_api_workflow` already orders plain widget keys ahead of dotted ones so the combo selection lands before the sub-widgets it exposes, and `apply_specs` resolves them from there.

Two deliberate choices worth review:

- **`"Flux.2 [pro]"` is the fixed combo option.** It is the one that posts to this entry's `endpoint`, so the alias keeps meaning the model it has always meant. The node's other option, `"Flux.2 [max]"`, is a different model at a different price and wants its own alias rather than being reachable by accident from this one.
- **`prompt_upsampling` is dropped from the mapping.** The proxy takes it, this node does not expose it, so an emitted workflow cannot carry it either way. `comfy generate` without `--emit-workflow` still sends it straight to the proxy.

## What guards it

`Flux2ImageNode` is recorded in the partner fixture from the catalog cloud's deployed ComfyUI serves, which lets the two existing contracts do the work:

- the widget-coverage test confirms every widget input the node declares is emitted,
- #864's bidirectional `deprecated_ok` assert now requires the flag to be **absent**, so a stale exemption cannot outlive this migration.

The round-trip test is extended to assert the combo selection survives alongside its sub-widgets. Without the selector the sub-widgets have nothing to hang off, so asserting it is what makes the width and height assertions mean anything.

## Checks

```
pytest tests/comfy_cli/command/generate      350 passed
ruff check / ruff format --check             clean
```

Round trip verified directly, API format in and lowered back out:

```
emitted:  {"model": "Flux.2 [pro]", "model.width": 512, "model.height": 768, "seed": 0, "prompt": "a fox"}
ops:      set_widget model, seed, prompt  then  model.width, model.height
lowered:  same keys, same values, allow_deprecated False on every add
```

## Not in scope

`ImageBatch`, the other deprecated class this emitter mints, is tracked separately in BE-13516. It is blocked on BE-10726 because `apply_specs` cannot wire an autogrow slot at all, which is a different gap from the dotted combo keys this PR uses.

## Unmapped flags now fail on every alias

The unmapped-flag check is not flux-2 only. `--emit-workflow` and `--emit-ops` now fail on any flag that the alias's `MODEL_NODE_MAP` entry does not wire into the node. Before this PR, every one of these flags was dropped without a word.

| alias | flags that now fail |
|---|---|
| `flux-2` | `input_image`, `input_image_2` … `input_image_9`, `output_format`, `prompt_upsampling`, `safety_tolerance` |
| `flux-ultra` | `guidance_scale`, `negative_prompt`, `num_images`, `num_inference_steps` |
| `kling-i2v` | `callback_url`, `camera_control`, `dynamic_masks`, `element_list`, `external_task_id`, `image_tail`, `multi_prompt`, `multi_shot`, `shot_type`, `sound`, `static_mask`, `watermark_info` |
| `nano-banana` | none |
| `seedance` | `fps`, `return_last_frame` |

`test_emit_refused_flags_per_alias` pins this table. It walks `MODEL_NODE_MAP`, sends each flag from the alias's schema one at a time, and records which ones fail. Wiring a flag later means removing it from the table.

The cloud agent will see this after its next comfy-cli pin bump. Its `generate_workflow` tool passes the flags the model chose after reading `generate_schema`. The error names the flags and the fix:

```
--emit-workflow for 'flux-2' does not map --input_image onto Flux2ImageNode. Drop that flag, or run `comfy generate flux-2` without --emit-workflow, which sends every flag straight to the proxy.
```

The message blames the mapping, not the node. Some flags have no input on the node, such as flux-2's `prompt_upsampling`. Others have an input the mapping does not wire yet, such as flux-2's `input_image` and `model.images`. Wiring `model.images` is a follow-up.

### DIFF
--- a/comfy_cli/command/generate/emit.py
+++ b/comfy_cli/command/generate/emit.py
@@ -173,22 +173,28 @@ MODEL_NODE_MAP: dict[str, NodeSpec] = {
     # Flux2ProImageNode would emit a workflow for a *different* model, so
     # `flux-pro` falls through to the EmitError instead.
     "flux-2": NodeSpec(
-        node_class="Flux2ProImageNode",
+        node_class="Flux2ImageNode",
         endpoint="bfl/flux-2-pro/generate",
         param_map={
             "prompt": "prompt",
-            "width": "width",
-            "height": "height",
+            "width": "model.width",
+            "height": "model.height",
             "seed": "seed",
-            "prompt_upsampling": "prompt_upsampling",
         },
-        fixed={"width": 1024, "height": 768, "seed": 0, "prompt_upsampling": True},
+        # `model` is the dynamic-combo selector and its sub-widgets are addressed
+        # through it. It must be set before them, which `ops_from_api_workflow`
+        # guarantees by ordering plain keys ahead of dotted ones.
+        #
+        # "Flux.2 [pro]" is the option that posts to this entry's `endpoint`.
+        # The node's other option, "Flux.2 [max]", is a different model at a
+        # different price, so it needs its own alias and endpoint rather than
+        # being reachable by accident from this one.
+        #
+        # No `prompt_upsampling`: the proxy takes it, this node does not expose
+        # it, so the emitted workflow cannot carry it. `comfy generate` without
+        # --emit-workflow still sends it straight to the proxy.
+        fixed={"model": "Flux.2 [pro]", "model.width": 1024, "model.height": 768, "seed": 0},
         output="IMAGE",
-        # ComfyUI deprecated Flux2ProImageNode in favor of Flux2ImageNode,
-        # whose width/height live inside a dynamic combo that
-        # the flat `param_map` cannot express. Emitting the deprecated class is
-        # the status quo until that migration lands; this flag comes off with it.
-        deprecated_ok=True,
     ),
     # BFL Flux 1.1 [pro] Ultra (text-to-image). Node: FluxProUltraImageNode.
     # The node takes an `aspect_ratio` string where the proxy schema takes
--- a/comfy_cli/command/generate/emit.py
+++ b/comfy_cli/command/generate/emit.py
@@ -351,6 +351,26 @@ def build_workflow(model: str, values: dict[str, Any], *, output_prefix: str = "
         if width_given and height_given:
             node_inputs[ns.aspect_from_wh] = f"{values['width']}:{values['height']}"
 
+    # Same rule, generalized: every flag the user actually typed has to land
+    # somewhere. `parse_args` fills no defaults, so `values` holds only what argv
+    # carried — anything in it that no mapping consumes would be discarded in
+    # silence. A proxy flag the node does not expose cannot be honored by an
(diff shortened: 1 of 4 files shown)


skishore23 (approved):
Checked this one properly. 354 pass, ruff clean. The fixture Flux2ImageNode matches the catalog cloud actually serves, including the 256..2048 step 32 bounds, and Flux2ProImageNode is deprecated:true there. The plain-before-dotted ordering in ops_from_api_workflow is already on main from #840. Ran --emit-workflow and --emit-ops for flux-2 by hand and the graph comes out right. Reverting node_class back to the pro node trips the #864 assert like it should. Approving the migration.

Two things on the generic unsupported flag check, non blocking but id like them before or right after merge.

1) Its not flux-2 only. It now hard errors on every mapped model, so flux-ultra --negative_prompt, seedance --fps, kling-i2v --camera_control and about 14 more flags that used to be dropped silently. I think loud is better than silent, but its only tested on flux-2 and the description doesnt mention it. The cloud agents generate_workflow tool forwards whatever flags the model picked from generate_schema, so after the next pin bump the agent will see these errors. The message tells it what to do so thats ok, but please add one table driven test that walks MODEL_NODE_MAP and pins which flags raise for which alias, and say so in the description.
2) The message says the node has no matching input. Thats wrong for some of them. flux-2 --input_image raises but Flux2ImageNode does have model.images, we just dont wire it. Same for kling --image_tail. Blame the mapping not the node. Wiring model.images is a follow up.

Small stuff: the header comment at emit.py:168 still says Node: Flux2ProImageNode. "Drop the flag" is singular for a list. Stray f prefix at emit.py:371. deprecated_ok=True now has no live entry in the map so the only positive coverage is ImageBatch, fine for now. The value is not None guard and the dotted ordering both survive being deleted, the ordering only because fixed{} happens to insert model first.

PR #869, opened 2026-09-12
feat(agent): comfy agent permissions|allow and the comfy-agent-permissions reference skill

skishore23:
## Description

Linear: BE-13755

The local comfy agent (`comfy-agent` on a user's machine) runs every command in a sandbox and every file tool behind a fence. When something blocks it, the agent asks. A security review found that its `allow_path` / `allow_host` tools were model-callable and wrote the approval stores directly, so injected content in a workflow could persist an attacker-chosen host or folder. The agent side (cloud PR #9067) now has `request_path` / `request_host` tools that only **record a pending request** in `<data dir>/permissions.json`; nothing the model can call moves a target into an approved list. **This CLI is the human approval channel**: a person reads the exact target the request names and approves or refuses it from a terminal (a panel button can do the same later).

- `comfy agent permissions` — the requests waiting for approval (id, kind, target, reason, age), then the approved folders and hosts, the data dir, and for a running agent its sandbox status and ComfyUI workspace (read from `/health` on the port `agent.json` publishes). The grant hints name `comfy agent allow --approve <id>` first; the JSON payload carries `pending`.
- `comfy agent allow --approve <id>` — approve one pending request: its target is vetted exactly as a direct grant is (below), appended to `permissions.json` `paths` or `egress-allow.json` `hosts` with the request's reason and `approved_at` now, and the request is dropped. A target that can never be granted is refused with nothing written and the request left pending; an unknown id is `agent_unknown_request`; `--approve` with `--path`/`--host` is `agent_bad_args`. Approving one request never touches another.
- `comfy agent deny <id>` — drop a request; nothing is approved.
- `comfy agent allow --path <folder> | --host <host> [--reason]` — grant outright, with the agent's own refusals: relative or missing folders, the whole disk, credential stores and any folder that contains one, the agent's data dir, `.env` names; wildcard, bare, loopback and known-refused hosts.
- Both files are rewritten atomically; a running agent re-reads them within about 20 seconds, otherwise the change applies at its next start. Data dir resolution mirrors the agent and its launcher: `--data-dir`, else `AGENT_DATA_DIR`, else `~/.comfy-agent`.
- `comfy-agent-permissions` — a reference skill (`skills show`; never installed into a coding agent, imported on demand by the agent): what the agent may reach by default; that `request_path` / `request_host` record a request and change nothing, that a yes typed in chat is **not** an approval, and that the user approves with `comfy agent allow --approve <id>` or a panel button; the start-time settings (`AGENT_COMFY_PATH`, `AGENT_SANDBOX_ALLOW_PATHS`, `AGENT_EGRESS_ALLOW_HOSTS`, `AGENT_SANDBOX_UNENFORCED_OK`); what can never be granted; and the Windows AppContainer's limits. The `comfy` skill cites it, as the reference-skill test requires.
- **Windows: `--json` envelopes with a character outside cp1252 vanished** (`UnicodeEncodeError` is a `ValueError`, which the renderer's write path treats as "no stream"), so `comfy --json skills list` printed nothing and the agent imported zero packs there. The envelope is now written ASCII-escaped when the first write cannot be encoded. Measured on the Windows box: 0 bytes before, 4,149 bytes after, and the agent then offers every CLI pack.
- **`comfy-custom-nodes`** — a bundled skill for writing a custom node pack against the V3 schema (nodes 2.0: `io.ComfyNode`, `io.Schema`, `ComfyExtension`, `comfy_entrypoint`): template, schema fields, every input and output kind, previews and saving, progress, the frontend hook, testing without a restart loop, `pyproject` + `comfy node publish`, a V1 reading table, and a link to the deeper nine-skill MIT set. Grounded on docs.comfy.org/custom-nodes/v3_migration. The agent's always-on node-authoring bullet points at it (cloud #9067).

The on-disk contract both sides follow: `permissions.json` is `{"paths": [{path, reason, approved_at}], "pending": [{id, kind, target, reason, requested_at}]}` (`id` 16 lower-case hex, `kind` `path` | `host`, `target` the absolute folder or bare host, RFC 3339 UTC timestamps); `egress-allow.json` is `{"hosts": [{host, reason, approved_at}]}`, unchanged. Envelope shapes are in `comfy_cli/schemas/agent.json`; `comfy agent deny` is registered in `COMMAND_SCHEMAS`.

Pairs with cloud #9067 (agent side: `request_path` / `request_host`, store refresh, importing `reference_skills` from `skills list`).

## How has this been tested?

- `pytest tests/comfy_cli/agent tests/comfy_cli/skills tests/comfy_cli/output/test_discovery.py tests/comfy_cli/output/test_error_code_registry.py tests/comfy_cli/output/test_envelope_schemas.py`: 188 pass. New, written first: `read_state` returns `pending` and skips malformed entries; approve a path / a host request; unknown id raises and leaves both of two requests pending; a refused target (home folder, loopback host, missing folder) raises with the file byte-identical and the request still pending; approve dedupes against an existing approval; deny removes only its entry (junk and other requests kept); the CLI round trips for `permissions` (pending listed above the approved lists, `approve` hint first, pretty output shows id and age), `allow --approve` (envelope validates, second approve is `agent_unknown_request`, no-agent warning), `--approve` + `--path`/`--host` is `agent_bad_args`, a refused approve stays pending, `deny` round trip, a broken `permissions.json` is `agent_state_unreadable` and never overwritten; the schema rejects incomplete approve/deny payloads.
- `comfy skills validate comfy_cli/skills/comfy-agent-permissions` valid; `comfy --json skills list` lists it under `reference_skills`.
- `ruff check` / `ruff format --check` clean on the touched files; `tests/comfy_cli/output/test_json_line_encoding.py` red on the old renderer, green after.
- Exercised by hand against a temp data dir with a seeded `pending`: a loopback host request is refused and stays pending, `deny` clears it, a path request lands in `paths` with the request's reason and resolved target, `--approve` + `--path` is `agent_bad_args`.
- Windows box (deep-comfy): the patched renderer dropped into the agent's provisioned venv makes `comfy --json skills list` print again and the agent import all packs.
- Not yet exercised against a running agent from this branch (the agent provisions a pinned comfy-cli; the pin moves separately).

https://claude.ai/code/session_01AZNnrBCcNqHFgNFjk4wdSk


comfy_cli/agent/__init__.py
```diff
@@ -0,0 +1,183 @@
...
+def allow_path(root: Path, folder: str, reason: str) -> tuple[Path, bool]:
+    """Record an approved folder. Returns (folder, added); added is False when it was already there."""
+    p = vet_path(folder)
+    file = root / PERMISSIONS_FILE
+    data = _read_json(file)
+    entries = [e for e in (data.get("paths") or []) if isinstance(e, dict)]
+    if any(_same_path(str(e.get("path", "")), str(p)) for e in entries):
+        return p, False
+    entries.append({"path": str(p), "reason": reason, "approved_at": _now()})
+    _write(file, {"paths": entries})
```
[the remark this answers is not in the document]
skishore23: Leaving this as is. The other writer of these files is the agent itself (its allow_path/allow_host tools), which does not share a lock with the CLI, so a CLI-only lock would serialize only two terminal invocations landing in the same instant and still not cover the case that matters. atomic_write_text already rules out a torn file; the worst outcome of the remaining window is one lost approval that the user re-runs, and the command's added:false dedup makes that re-run safe. A shared lock belongs with a contract the agent side owns.

comfy_cli/agent/__init__.py
```diff
@@ -0,0 +1,205 @@
...
+    try:
+        p = p.resolve(strict=True)
+    except OSError as exc:
+        raise ValueError(f"{p} cannot be resolved: {exc}") from exc
+    if p.parent == p:
+        raise ValueError(f"{p} is the whole filesystem and stays closed")
+    home = Path.home().resolve()
+    for rel in _DENIED_HOME_SUBDIRS:
+        denied = (home / rel).resolve()
+        if _same_or_under(p, denied):
```
annehe9: This check is case-sensitive, but the default macOS disk is not. `vet_path("~/.SSH")` and `vet_path("~/library/keychains")` both pass. They open the same folders as `~/.ssh` and `~/Library/Keychains`.

`resolve()` does not fix the case on POSIX, and `relative_to` compares case exactly. I reproduced this on APFS. Comparing casefolded paths on darwin would close it. On a case-sensitive volume that only refuses a little more, which seems fine.

comfy_cli/agent/__init__.py
```diff
@@ -0,0 +1,205 @@
...
+
+def _same_or_under(child: Path, parent: Path) -> bool:
+    try:
+        child.relative_to(parent)
+    except ValueError:
+        return False
+    return True
+
+
+def vet_host(raw: str) -> str:
```
annehe9, line 297: `vet_host` lets through a lot that should be refused. These all came back unchanged: `*`, `com`, `a.com,b.com`, `127.0.0.1`, `169.254.169.254`. And `user:pw@evil.com` becomes `user`.

The permissions skill says telemetry endpoints and shared object storage can never be granted. Nothing here checks for either.

I haven't read the agent's host matching code, so I don't know how bad `*` or `com` is in practice. At minimum I'd reject wildcards, commas, `@`, single-label names, and loopback or link-local IPs.

comfy_cli/agent/__init__.py
```diff
@@ -0,0 +1,205 @@
...
+    except OSError as exc:
+        raise ValueError(f"{p} cannot be resolved: {exc}") from exc
+    if p.parent == p:
+        raise ValueError(f"{p} is the whole filesystem and stays closed")
+    home = Path.home().resolve()
+    for rel in _DENIED_HOME_SUBDIRS:
+        denied = (home / rel).resolve()
+        if _same_or_under(p, denied):
+            raise ValueError(f"{p} stays closed: it is a credential or agent-state folder")
+        if _same_or_under(denied, p) and denied.exists():
```
annehe9: Because of `denied.exists()`, a parent folder is refused only if the credential store is already there. I set up a HOME with no `.ssh`, `.aws`, `.docker` or `.kube`, and the data dir outside it. `vet_path("~")` passed.

A later `ssh-keygen` or `aws configure` then puts keys inside a folder the agent can already reach. The skill says the home folder can never be granted. Dropping `exists()` for this parent case would match that.

comfy_cli/agent/__init__.py
```diff
@@ -0,0 +1,205 @@
...
+from comfy_cli.file_utils import atomic_write_text
+
+PERMISSIONS_FILE = "permissions.json"
+EGRESS_ALLOW_FILE = "egress-allow.json"
+DISCOVERY_FILE = "agent.json"
+
+# Folders a terminal grant may never name, in either direction: the same
+# credential-store rule the agent's store enforces, so a grant that the agent
+# would refuse from chat cannot be smuggled in from the CLI.
+_DENIED_HOME_SUBDIRS = (
```
annehe9, line 70: The comment says this is the same rule the agent's store enforces. It is narrower than the skill's "What can never be granted" section. That section lists browser profiles and any `.env`, and neither is covered here.

`vet_path` approved `~/.config/gh`, which holds a GitHub token. From reading the list, it would also approve `~/Library/Application Support/Google/Chrome`, `~/.mozilla`, `~/.local/share/keyrings`, and any project folder with a `.env` in it.

I can't see the agent's own list. If it matches this one, the skill overpromises. If it is longer, the CLI is a way around it.

comfy_cli/agent/command.py
```diff
@@ -0,0 +1,141 @@
...
+    added: dict = {}
+    try:
+        # Vet everything before the first write so a refused --host does not
+        # leave a --path approved behind a failure exit.
+        if path:
+            vet_path(path, root=root)
+        if host:
+            vet_host(host)
+        if path:
+            folder, is_new = allow_path(root, path, reason)
```
annehe9, line 272: The pre-vet only checks the host string. It never reads `egress-allow.json`. If that file is bad JSON, `allow_host` raises after `allow_path` has already written.

With `{not json` in that file, `allow --path X --host ok.example.com` exits 1 and X stays in `permissions.json`. The error also comes out as `refused` with the narrower-folder hint. It should be `agent_state_unreadable`.

annehe9 (commented):
My main concern is the grant checks in `comfy agent allow`. The deny check is case-sensitive on macOS. `vet_host` accepts `*` and bare TLDs. The home folder can be granted when no credential folder exists yet. The rest are error handling gaps and a few skill doc fixes.

Details are inline. I ran the repro for most of them against 1b58dba. Where I only read the code or am guessing, the comment says so.

comfy_cli/agent/command.py
```diff
@@ -0,0 +1,141 @@
...
+PICKUP_NOTE = "a running agent picks this up within about 20 seconds; otherwise it applies at the agent's next start"
+
+
+def _sandbox_status(port: int | None) -> dict | None:
+    if not port:
+        return None
+    try:
+        with urllib.request.urlopen(f"http://127.0.0.1:{port}/health", timeout=2) as resp:  # noqa: S310 - loopback
+            body = json.loads(resp.read().decode("utf-8"))
+    except (OSError, ValueError, urllib.error.URLError):
```
annehe9: `http.client.HTTPException` is not an `OSError`, so it gets past this. I pointed the port at a socket that replies `SSH-2.0-OpenSSH`. `_sandbox_status` raised `BadStatusLine`. So `comfy agent permissions` prints a traceback with no JSON envelope when a stale port is taken by something else. `IncompleteRead` from `resp.read()` would do the same.

comfy_cli/agent/command.py
```diff
@@ -0,0 +1,141 @@
...
+            vet_path(path, root=root)
+        if host:
+            vet_host(host)
+        if path:
+            folder, is_new = allow_path(root, path, reason)
+            added["path"] = {"path": str(folder), "added": is_new}
+        if host:
+            h, is_new = allow_host(root, host, reason)
+            added["host"] = {"host": h, "added": is_new}
+    except ValueError as exc:
```
annehe9, line 280: Only `ValueError` is caught here. An `OSError` from `mkdir` or the atomic write comes out as a raw traceback. With `--data-dir` under a read-only folder this gave a `PermissionError`, exit 1, and nothing on stdout, so JSON callers get no envelope.

I think `is_dir()` in `vet_path` can also raise on EACCES on Python 3.10 to 3.12. I only read the stdlib for that one and didn't run it.

comfy_cli/agent/command.py
```diff
@@ -0,0 +1,141 @@
...
+]
+
+PICKUP_NOTE = "a running agent picks this up within about 20 seconds; otherwise it applies at the agent's next start"
+
+
+def _sandbox_status(port: int | None) -> dict | None:
+    if not port:
+        return None
+    try:
+        with urllib.request.urlopen(f"http://127.0.0.1:{port}/health", timeout=2) as resp:  # noqa: S310 - loopback
```
annehe9: This uses the global `urlopen`, so `HTTP_PROXY` applies to 127.0.0.1 unless `NO_PROXY` covers it. With `HTTP_PROXY=http://127.0.0.1:9` and a live /health server, this returned None, and the agent showed as not running.

My guess is a shell behind the agent's egress proxy is where this would show up. `build_http_only_opener(urllib.request.ProxyHandler({}))` from `comfy_cli/http.py` skips the proxy.

comfy_cli/agent/command.py
```diff
@@ -0,0 +1,141 @@
...
+        rprint(
+            f"[bold]agent[/bold]     {'running on port ' + str(state.port) if payload['agent']['running'] else 'not running'}"
+        )
+        if payload["sandbox"]:
+            rprint(f"[bold]sandbox[/bold]   {json.dumps(payload['sandbox'])}")
+        if payload["comfy_path"]:
+            rprint(f"[bold]ComfyUI[/bold]   {payload['comfy_path']}")
+        rprint("[bold]folders the user approved[/bold]" + ("" if state.paths else "  (none)"))
+        for e in state.paths:
+            rprint(f"  {e.get('path')}  — {e.get('reason', '')}")
```
annehe9: `path` and `reason` come from files on disk and go into Rich markup as-is. A reason of `notes [/] here` makes `comfy agent permissions` crash with `MarkupError`. A folder named `/x/[bold]draft` prints as `/x/draft`.

Lines 85, 87, 93 and 139 have the same issue with the sandbox JSON, `comfy_path` and host. `sanitize_markup` from `comfy_cli/output/sanitize.py` handles this.

comfy_cli/agent/__init__.py
```diff
@@ -0,0 +1,205 @@
...
+    hosts: list[dict]
+
+
+def read_state(root: Path) -> AgentState:
+    """What the data dir says: whether an agent published itself, and the approvals."""
+    disc = _read_json(root / DISCOVERY_FILE)
+    raw_port = disc.get("port")
+    # bool is an int subclass; a JSON true must not become port 1.
+    port = raw_port if type(raw_port) is int and 1 <= raw_port <= 65535 else None
+    paths = _read_json(root / PERMISSIONS_FILE).get("paths") or []
```
annehe9: `or []` doesn't check the type. `{"paths": 1}` in `permissions.json` makes `read_state` raise `TypeError: 'int' object is not iterable`. `comfy agent permissions` then shows a traceback instead of `agent_state_unreadable`. The same goes for `hosts` on the next line.

comfy_cli/agent/__init__.py
```diff
@@ -0,0 +1,205 @@
...
+def allow_path(root: Path, folder: str, reason: str) -> tuple[Path, bool]:
+    """Record an approved folder. Returns (folder, added); added is False when it was already there."""
+    p = vet_path(folder, root=root)
+    file = root / PERMISSIONS_FILE
+    data = _read_json(file)
+    entries = [e for e in (data.get("paths") or []) if isinstance(e, dict)]
+    if any(_same_path(str(e.get("path", "")), str(p)) for e in entries):
+        return p, False
+    entries.append({"path": str(p), "reason": reason, "approved_at": _now()})
+    _write(file, {"paths": entries})
```
annehe9: This writes back only `paths`, so any other top-level key is dropped. `{"version": 2, "paths": [], "denied": ["x"]}` came back as just `{"paths": [...]}`. If the file holds a list, `_read_json` returns `{}` and the file is replaced with no error. `{**data, "paths": entries}` would keep the rest. Line 189 has the same problem for hosts.

There is also no lock. If the agent writes a grant at the same moment, one of the two is lost. I haven't checked how the agent itself writes this file.

comfy_cli/agent/__init__.py
```diff
@@ -0,0 +1,205 @@
...
+
+
+def _same_path(a: str, b: str) -> bool:
+    if os.name == "nt":
+        return os.path.normcase(os.path.normpath(a)) == os.path.normcase(os.path.normpath(b))
+    return os.path.normpath(a) == os.path.normpath(b)
+
+
+def _write(file: Path, payload: dict) -> None:
+    file.parent.mkdir(parents=True, exist_ok=True)
```
annehe9, line 454: This creates whatever data dir it is given, and then the command says a running agent will pick the grant up. I only read the code for this one.

If the launcher sets `AGENT_DATA_DIR` for the agent but not in the user's shell, `comfy agent allow` writes to a new `~/.comfy-agent`. The agent never sees it. A typo in `--data-dir` does the same. A warning when the dir or `agent.json` did not exist before the write would catch both.

comfy_cli/skills/comfy-custom-nodes/SKILL.md
```diff
@@ -0,0 +1,174 @@
...
+license = { text = "MIT" }
+requires-python = ">=3.10"
+
+[tool.comfy]
+PublisherId = "<from registry.comfy.org>"
+DisplayName = "My Pack"
+```
+
+`comfy node init` writes the skeleton; `comfy node publish` uploads the
+version in `pyproject.toml` (every git-tracked file); set `COMFY_API_KEY` in
```
annehe9: Two things here don't match the CLI. `comfy node publish` never reads `COMFY_API_KEY`. It takes `--token` or calls `typer.prompt`, so in CI it would stop at a prompt with no TTY. And `comfy node init` only writes `pyproject.toml`. It doesn't write a skeleton.

comfy_cli/skills/comfy-custom-nodes/SKILL.md
```diff
@@ -0,0 +1,174 @@
...
+|---|---|
+| `INPUT_TYPES(s)` dict | `define_schema` → `io.Schema(inputs=…)` |
+| `RETURN_TYPES` / `RETURN_NAMES` | `outputs=[io.X.Output(display_name=…)]` |
+| `FUNCTION = "run"` | always `execute` |
+| `CATEGORY`, `OUTPUT_NODE` | `category`, `is_output_node` |
+| `IS_CHANGED`, `VALIDATE_INPUTS` | `fingerprint_inputs`, `validate_inputs` |
+| `NODE_CLASS_MAPPINGS` / `NODE_DISPLAY_NAME_MAPPINGS` | `ComfyExtension.get_node_list()`, `display_name` |
+
+Old workflows keep working across a migration when the extension's `on_load`
+registers `io.NodeReplace(new_node_id=…, old_node_id=…, input_mapping=…)`.
```
annehe9: Minor. The [V3 migration docs](https://docs.comfy.org/custom-nodes/v3_migration) call `await api.node_replacement.register(io.NodeReplace(...))` inside `on_load`. Showing that call here would stop an agent from building the object and never registering it.

comfy_cli/error_codes.py
```diff
@@ -548,6 +548,24 @@ class ErrorCode:
...
     ),
+    # --- agent ---------------------------------------------------------------
+    ErrorCode(
+        "agent_state_unreadable",
+        "A file in the local agent's data dir (agent.json, permissions.json, "
+        "egress-allow.json) could not be read or parsed.",
+        "fix or remove the file named in the message and try again",
+    ),
+    ErrorCode(
+        "bad_args",
```
annehe9: `bad_args` and `refused` are very broad names for the global registry. `usage_error` and `missing_argument` already cover the first one. Other areas scope their codes, like `generate_bad_args` and `build_missing_input`. Something like `agent_bad_args` and `agent_refused` would keep `comfy discover` clearer.

comfy_cli/agent/command.py
```diff
@@ -0,0 +1,130 @@
...
+        raise typer.Exit(code=2)
+    root = data_dir(data_dir_opt)
+    added: dict = {}
+    try:
+        if path:
+            folder, is_new = allow_path(root, path, reason)
+            added["path"] = {"path": str(folder), "added": is_new}
+        if host:
+            h, is_new = allow_host(root, host, reason)
+            added["host"] = {"host": h, "added": is_new}
```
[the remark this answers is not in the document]
skishore23: Fixed in c6b70ba9. allow_cmd now calls read_state (all three files) and vets both values before the first write; a refused --host or a broken egress-allow.json leaves nothing behind. Regression tests: test_cli_allow_refuses_a_bad_host_without_persisting_the_path and test_cli_allow_reads_every_state_file_before_the_first_write.

comfy_cli/agent/__init__.py
```diff
@@ -0,0 +1,195 @@
...
+        raise ValueError(f"{p} is not an existing folder")
+    try:
+        p = p.resolve(strict=True)
+    except OSError as exc:
+        raise ValueError(f"{p} cannot be resolved: {exc}") from exc
+    if p.parent == p:
+        raise ValueError(f"{p} is the whole filesystem and stays closed")
+    home = Path.home().resolve()
+    for rel in _DENIED_HOME_SUBDIRS:
+        denied = (home / rel).resolve()
```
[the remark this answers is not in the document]
skishore23: Addressed in 1b58dba (root is passed to vet_path from both allow_cmd and allow_path). c6b70ba9 adds the descendant case to the test.

tests/comfy_cli/agent/test_command.py
```diff
@@ -0,0 +1,250 @@
...
+    root.mkdir(parents=True)
+    sibling = tmp_path / "state" / "photos"
+    sibling.mkdir()
+    with pytest.raises(ValueError, match="agent"):
+        vet_path(str(root), root=root)
+    with pytest.raises(ValueError, match="agent"):
+        vet_path(str(tmp_path / "state"), root=root)
+    with pytest.raises(ValueError, match="agent"):
+        vet_path(str(tmp_path), root=root)
+    assert vet_path(str(sibling), root=root) == sibling.resolve()
```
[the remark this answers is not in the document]
skishore23: Added in c6b70ba9: test_vet_path_refuses_the_agent_data_dir_its_parents_and_its_children creates root/project and asserts vet_path raises.

comfy_cli/agent/__init__.py
skishore23: Fixed in c6b70ba9. Paths are compared through _fold: os.path.normcase plus lower() on darwin, so ~/.SSH and ~/library/keychains are refused there. test_vet_path_folds_case_where_the_filesystem_does forces sys.platform to darwin so it runs on Linux CI too. The agent's own UserPathStore only folds case on Windows (pathsFoldCase), so its allow_path tool has the same gap on macOS; I will take that to the cloud repo separately.

comfy_cli/agent/__init__.py
skishore23: Fixed in c6b70ba9. vet_host now refuses wildcards, commas, credentials (user:pw@ in either form), bare names and TLDs (localhost, com), loopback / link-local / unspecified / multicast / reserved addresses, and the hosts the agent's proxy never opens (t.comfy.org, api.mixpanel.com, storage.googleapis.com, r2.cloudflarestorage.com). Labels are validated and IDNA-encoded. On the agent side: user hosts are matched exactly (no wildcard or suffix), so * and com were useless entries rather than open ones, but 127.0.0.1 was real: allowsPort exempts loopback, so it would have opened every local port through the proxy on Windows. Parametrized test covers each input you listed.

comfy_cli/agent/__init__.py
skishore23: Fixed in c6b70ba9. The exists() check is gone; a parent of any listed store is refused whether the store exists yet or not, which is what the agent's UserPathStore does (lexical underPath, no stat). test_vet_path_refuses_the_home_folder_before_any_store_exists covers an empty HOME with the data dir outside it.

comfy_cli/agent/__init__.py
skishore23: Fixed in c6b70ba9. The list now mirrors the agent's CredentialDenyDirs (cloud repo, runner/sandbox_seatbelt_profile.go) on every platform: .config/gh, .git-credentials, .pypirc, .npmrc, .cache/huggingface/token, the macOS/Linux/Windows browser profiles, GNOME keyrings and the DPAPI vaults, plus .azure and the whole of .docker where the CLI is stricter. .env is handled the way the agent handles it: a name rule on every path component (.env, .env.*, .envrc, case-insensitive), so a folder named .env cannot be granted, and a .env file inside an approved folder stays unreadable at the agent regardless of the grant. The skill's never-granted section now says exactly that.

comfy_cli/agent/command.py
skishore23: Fixed in c6b70ba9. read_state runs before any vet or write, and an unreadable or malformed state file is a StateError mapped to agent_state_unreadable, distinct from a refusal. With {not json in egress-allow.json, allow --path X --host ok exits 1 with agent_state_unreadable naming the file and permissions.json is never created (test_cli_allow_reads_every_state_file_before_the_first_write).

comfy_cli/agent/command.py
skishore23: Fixed in c6b70ba9. http.client.HTTPException joins the except tuple. test_sandbox_status_survives_a_stale_port_that_is_not_http binds a real loopback socket that answers with an SSH banner and asserts the probe reads as no agent.

comfy_cli/agent/command.py
skishore23: Fixed in c6b70ba9. OSError from mkdir or the atomic write is reported as a new agent_state_unwritable code with an envelope; test_cli_allow_reports_an_unwritable_data_dir_as_an_envelope uses a 0o500 parent. You were right about is_dir(): on 3.10 to 3.12 it only swallows ENOENT/ENOTDIR/EBADF/ELOOP, so EACCES propagates; vet_path now wraps it into a refusal.

comfy_cli/agent/command.py
skishore23: Fixed in c6b70ba9. The probe goes through build_http_only_opener(ProxyHandler({})). test_sandbox_status_ignores_http_proxy_for_loopback runs a real /health server on a loopback port with HTTP_PROXY=http://127.0.0.1:9 and NO_PROXY unset.

comfy_cli/agent/command.py
skishore23: Fixed in c6b70ba9. Every value from disk or /health that lands in a markup string goes through sanitize_markup (path, reason, host, sandbox JSON, comfy_path, data dir, and the allow summary). test_pretty_output_escapes_rich_markup_from_disk uses a reason of notes [/] here, a folder named [bold]draft and a [link=...] host in pretty mode.

comfy_cli/agent/__init__.py
skishore23: Fixed in c6b70ba9. Entries are loaded through _load, which raises StateError when paths/hosts is not a list of objects; the command reports agent_state_unreadable. Parametrized in test_state_that_the_agent_would_not_load_is_reported_not_replaced.

comfy_cli/agent/__init__.py
skishore23: Fixed in c6b70ba9 for the key loss: writes are {**data, key: entries}, and a file whose top level is not an object is a StateError, never silently replaced (test_allow_keeps_other_top_level_keys and the not-replaced test). On the lock: as in the thread above with CodeRabbit, the other writer is the agent itself and it shares no lock contract with the CLI, so a CLI-only lock would not cover the case that matters; atomic_write_text rules out a torn file and the worst outcome is one lost approval the user re-runs safely. A shared lock belongs with a contract the agent owns.

comfy_cli/agent/__init__.py
skishore23: Fixed in c6b70ba9. When no agent.json exists in the data dir before the write, the command adds a warning to the envelope (schema: optional warning on the allow shape) and prints it in pretty mode, saying no agent has run from this dir and a running one elsewhere will not see the grant. Covered in the round-trip test and test_cli_allow_does_not_warn_when_an_agent_published_itself.

comfy_cli/skills/comfy-custom-nodes/SKILL.md
skishore23: Fixed in c6b70ba9. The publish paragraph now says comfy node init writes only pyproject.toml, and that comfy node publish takes --token or prompts, so CI passes --token; the COMFY_API_KEY line is gone.

comfy_cli/skills/comfy-custom-nodes/SKILL.md
skishore23: Fixed in c6b70ba9. The migration note now shows on_load awaiting api.node_replacement.register(io.NodeReplace(...)) with the ComfyAPI import, per the V3 migration page.

comfy_cli/error_codes.py
skishore23: Fixed in c6b70ba9: agent_bad_args and agent_refused, and the new write failure is agent_state_unwritable. The registry test pins the call sites.

annehe9 (approved):

PR #874, opened 2026-09-15
fix(knowledge): check an oss pick's weights before recommending it (BE-8974)

annehe9:
## Summary

`available_locally` on a knowledge pick only covers templates and nodes. On a local install, the skill could steer the agent to a rank 1 `oss` pick whose weights are not on disk.

This extends rule 2 of "Curated knowledge" in `comfy_cli/skills/comfy/SKILL.md`. When routing is local, the agent runs `comfy --json templates check <template>` on an `oss` pick's `template` before recommending it. On `missing-models` it still names the pick, lists `models.missing[]` with each `directory`, and names the first `oss` pick that comes back `runnable`. It asks before downloading. A pick with no `template` is skipped.

Skill text only. `attach()` is untouched on purpose. `templates check` fetches template JSON and calls the local server, and discovery enrichment must never fetch.

Linear: BE-8974. Paired cloud PR with the same rule for the local agent: Comfy-Org/cloud#9368.

## Testing

* `uv run --extra dev pytest tests/comfy_cli/test_knowledge.py tests/comfy_cli/test_command_mentions.py tests/comfy_cli/skills/test_installer.py tests/comfy_cli/command/test_nodes_cli.py tests/comfy_cli/command/test_transfer_download.py` gives 328 passed, 1 skipped. These are the test files that read the skill file.
* Full suite not run locally.
* Not tried against a live local ComfyUI.


skishore23 (approved):

annehe9:
Superseded by #875, which puts this check in code instead of skill text: `comfy knowledge pick --check-local` flags an oss pick whose model files are missing, and rule 2 of the skill already says what to do with an `available_locally: false` pick. Review kept finding cases this prose missed, and it edits the same lines #875 rewrites.

PR #883, opened 2026-09-17
feat(layout): crossing reduction and coordinate assignment for batched builds

christian-byrne:
Stacked on #882. Rebase onto `main` once that merges.

## Why the batch path

Langfuse, 600 sampled `apply_ops` calls on production traffic:

| | |
|---|---|
| batches that create nodes | 489 |
| …carrying their `connect` ops in the **same** batch | **481 = 98%** |
| node creations via `apply_ops` | 3,120 |
| node creations via serial `add_node` | 461 |

So ~87% of layout quality is decided by `assign_positions`, not the serial cascade. It is also the weaker of the two placers: it implemented one of Sugiyama's three steps.

| Step | Status before |
|---|---|
| 1. Layer assignment — which column | ✅ longest-path relaxation |
| 2. **Crossing reduction** — order within a column | ❌ ops-array order |
| 3. **Coordinate assignment** — y | ❌ stacked from `base_y` |

That is why a batched build could be correctly layered and still look wrong: three inputs feeding one sampler crossed their wires purely because of the order they were authored in, and a one-node column sat flush with the top of a five-node column instead of level with the node it feeds.

## Changes

- **Step 2** — order each column by the barycentre of its neighbours, forward and backward sweeps, insertion rank as a deterministic tiebreak. Determinism matters here beyond tidiness: placement is frozen at mint time so replay converges, and a non-deterministic ordering would break that.
- **Step 3** — centre each node on the mean centre of its predecessors, then push down only as far as column order requires.
- **Pinned siblings are now obstacles.** `collides()` compared movable nodes only against *existing* workflow nodes, so a new node could be placed on top of a new node pinned to an explicit `at`.

## Measured

A fixture whose ops are deliberately authored in a crossing order:

| | crossings | mean input-alignment deviation |
|---|---:|---:|
| before | 3 | 176px |
| after | **0** | **0px** |

A realistic SDXL graph: alignment deviation **46.6px → 30.4px**, crossings 0 both ways.

The scoring helper lands in the test file so future layout changes are measured rather than eyeballed — that metric is the prerequisite for the A/B/C prototype comparison.

## Tests

Three regression tests, verified red on the parent commit and green here. Full suite: 45 failures vs 46 on parent, zero introduced.

## One thing shipping unproven

The direct-jump collision change — replacing a `ROW_GAP`-at-a-time march bounded at `_GUARD * ROW_GAP` = 40,000px — **has no red-green test**, and there is a comment in the test file saying so. Every fixture I tried either did not reach the collision branch or was cleared by the old march as well, so the intended proof (old code exhausts its budget and leaves an overlap) was never demonstrated. Treat it as a robustness and efficiency change, not a verified bug fix.


christian-byrne:
@coderabbitai review

christian-byrne:
@coderabbitai review

huntcsg (approved):
## Summary

This PR extends batch layout from longest-path column assignment to a bounded barycentre pass that reorders nodes within columns, then vertically centres movable nodes on already-positioned movable predecessors. It also treats explicitly positioned new nodes as collision obstacles and replaces 40px-at-a-time collision shifts with direct jumps below the blocking rectangle. Three focused tests cover crossing reduction, simple predecessor alignment, and pinned-node collision avoidance.

## Risks

- The quoted comprehension is dense, but it is not an unchecked lookup: `preds` is populated only for endpoints already present in the movable-node map, every such `adds[p]` has a `size`, and the `"pos"` filter skips predecessors not yet placed. I do not see a crash path in that line under this function's construction.
- The filtering immediately above it excludes pinned new nodes from `preds`. Consequently, a movable node connected from a pinned predecessor at `[40, 1000]` is placed at y=60 rather than centred near y=1000. The same limitation means crossing reduction ignores pinned neighbours. This is not worse than the parent behavior, but it is a meaningful gap in the advertised coordinate assignment for mixed explicit/automatic batches.
- The direct-jump collision rewrite changes production behavior without a regression test, as the PR itself notes. Its geometry looks internally consistent, but this is the highest regression risk in the diff.
- The scoring tests cover a clean two-column fixture only; they do not exercise mixed pinned/movable graphs, existing anchors, unequal node heights, long edges, or cyclic/invalid batches.

PR #885, opened 2026-09-17
feat(layout): score layout quality as a number instead of a screenshot

christian-byrne:
Stacked on #884 (→ #883 → #882). Rebases onto `main` as the stack lands.

## Why

Every prior round of "is the agent's layout any good" has ended in screenshot opinions. Screenshots do not compare across runs, cannot fail a build, and cannot tell you whether the thing you just changed helped. PRs #882 and #883 both needed an answer to that question and each had to build a throwaway one.

`comfy_cli/layout_quality.score()` returns four numbers, every one lower-is-better with 0.0 as the ideal, so they compose and a threshold reads the same way for all of them:

| Metric | Measures | Objective it maps to |
| --- | --- | --- |
| `overlap_area` | total pairwise overlapping px² | the defect Jo reported |
| `crossings` | edge pairs crossing between shared columns | Sugiyama step 2 |
| `align_deviation` | mean px from a node's centre to the mean centre of its inputs | Sugiyama step 3 |
| `backward_edges` | links whose target is left of or level with their source | dataflow reads left to right |

It is a pure function of geometry and links — no rendering, no canvas, no I/O — which is what lets one implementation serve three callers: test assertions, a CI regression baseline, and a telemetry signal emitted at op-mint time. That last one is the point. The overlap took two months to reach us through FE-1653; a number computed at mint time surfaces the next one without waiting for a user.

## Two decisions worth reviewing

**Overlap includes the 30px title band.** A scorer comparing only `pos` + `size` reports zero for nodes that visibly overlap by up to 30px — the exact blind spot that made the placer itself wrong before #882, so repeating it in the grader would have hidden the bug it was built to catch. `test_title_band_counts_as_occupied` pins it.

**Crossings are counted only between edges sharing both columns**, not by geometric intersection. Counting every intersection would make the number depend on `COL_GAP`, so a spacing change would read as a layout regression. This way the metric tracks what crossing reduction actually optimises.

## Deduplication

`test_layout.py`'s local `_score` now delegates here. Two implementations of one metric is the shape of bug where the tests and the telemetry quietly disagree about whether a layout improved — and `test_title_constant_matches_the_placer` asserts the scorer and the placer still agree about what a node occupies, so they cannot drift into being self-consistently wrong.

## Honest accounting of the baselines

30 tests. Five of them score `assign_positions`'s real output. Verified against the pre-#883 parent: **exactly one goes red there** (3 crossings, 176px mean deviation). The other four pass on the parent too — they are forward guards locking in behaviour that is already correct, not evidence that #883 fixed them. That is written into the class docstring, because a green block of five reads as five fixes and only one of them is.

Specifically, `test_a_wide_batch_never_overlaps` does **not** exercise the new-vs-new collision fix: with no edges the placer assigns from a descending cursor, so the old code never reached that branch either. The case that does exercise it is `test_pinned_siblings_are_obstacles_for_movable_nodes` in `test_layout.py`.

## What this does not measure

Whether the layout is *meaningful*. A scorer built around overlap and crossings will rank a mechanically tidy arrangement above a semantically grouped one, so a falling score is evidence and not proof. Any gate built on this needs a human-judged fixture beside it — noted in the module docstring.


annehe9 (approved):
Nice approach, decisions make sense, no defects found.

christian-byrne:
retargeted this to `main` and force-pushed it rebuilt on `main` with only its own four commits.

it was stacked on #884, and the conflict that appeared came from the two branches having independently fixed the same LiteGraph parity bug, each touching `.github/workflows/litegraph-parity.yml` and `scripts/check_litegraph_parity.py`. #884 keeps that work, including its reviewed version of the checker. the stacking was incidental here: nothing in `layout_quality.py` depends on #884, the four commits cherry-pick onto `main` cleanly, and the 69 tests in `test_layout_quality.py` plus `test_layout.py` pass there. the diff is now just the scorer and its tests.

PR #887, opened 2026-09-17
fix(layout): charge multiline widgets their measured height, not a widget row

christian-byrne:
Stacked on #886. Closes the second half of the under-measure that [ComfyUI_frontend #17901](https://github.com/Comfy-Org/ComfyUI_frontend/pull/17901) reproduces.

## What was wrong

A multiline text box is a text **area**, not a widget **row**. `estimate_size` charged every widget the same 24px, so a `CLIPTextEncode` came out 142px shorter than it draws — the dominant term in the overlap the browser harness catches.

## The number is measured, not guessed

That distinction is the whole point of this change. I created twelve core node classes in a real browser, let the real renderer size them, and fitted their heights to:

```
H = 6 + 20 * max(link_inputs, outputs) + (Σ widget heights + 8)
```

**All twelve fit exactly** with an ordinary widget at 24 and a multiline one at **166**:

| class | rendered | predicted | |
| --- | ---: | ---: | --- |
| CLIPTextEncode | 200 | 200 | 1 multiline |
| KSampler | 262 | 262 | 7 ordinary, 4 links |
| EmptyLatentImage | 106 | 106 | |
| CheckpointLoaderSimple | 98 | 98 | |
| SaveImage | 58 | 58 | |
| LoadImage | 102 | 102 | |
| VAEDecode | 46 | 46 | 0 widgets |
| PreviewImage | 26 | 26 | 0 widgets |
| ConditioningCombine | 46 | 46 | 0 widgets |
| LatentUpscale | 130 | 130 | |
| CLIPSetLastLayer | 58 | 58 | |
| ImageScale | 130 | 130 | |

The fixture and the fit live in the test file, so the model is re-checkable without a browser.

## Two things the measurement settled beyond the headline

**It independently confirms #886.** The 24 and the 8 there were derived by reading `computeSize`; this arrives at the same two values from rendered pixels, by a completely different route.

**The fit needs *true* link inputs.** `KSampler` has seven widgets but only six widget-backed inputs, because `control_after_generate` is a widget with no input at all. Deriving links as `len(inputs) - len(widgets)` is therefore off by a row — and an earlier fit that did exactly that matched 11 of 12 and looked close enough to accept. Worth knowing before anyone re-derives this.

## Implementation

`object_info` already marks multiline inputs and the catalog parses it into `PortOptions.multiline`, so no new input is needed. Matching is by name, because `widget_order` is the render order rather than the declaration order.

The attribute guard covers `options` itself rather than the field on it. A port carrying no options is not hypothetical — every test double is one, and so is any catalog entry from an `object_info` that omitted the block; guarding only the inner field raises `AttributeError` before the default can apply. That was a real crash I introduced and caught here.

Additive: `estimate_size` keeps its signature and `n_multiline` defaults to 0, so every existing caller is unchanged.

## Tests

17 new, including one parametrized case per measured class asserting the widget term reproduces rendered geometry. All verified red on the parent. Full layout suite: 103 passing.


christian-byrne:
release-blocking rather than a follow-up, and I have numbers now rather than an argument.

I scored three realistic agent-built graphs by taking the positions the placer actually chooses and measuring them against the **rendered** geometry (the 12-class fit: base 6, slot 20, ordinary widget 24, multiline 166, block pad 8, title 30) instead of against the CLI's own model. Scoring against the model is what hides this — the model believes it left a gap.

| fixture | `main` today (#882+#883 merged) | with this PR |
| --- | ---: | ---: |
| txt2img, 7 nodes | 19,492 px² | **0** |
| 4 prompts + combines | 58,477 px² | **0** |
| hires-fix, 9 nodes | 19,492 px² | **0** |

On `main` the two `CLIPTextEncode` nodes in a plain txt2img overlap by **250x78 px**. That is the original report, still present after both merged PRs, because every one of those graphs has a multiline prompt node and multiline is the term that was wrong.

The reason #882 and #883 did not close it: #882 fixed the *width* model and the title band, #883 fixed *ordering and collision* between new nodes. Both are correct and both were necessary. Neither changes the node's modelled height, and the height error is 142px per prompt node — larger than the vertical gap the placer leaves.

`browser_tests/tests/agent/agentLayoutQuality.spec.ts` in [ComfyUI_frontend#17901](https://github.com/Comfy-Org/ComfyUI_frontend/pull/17901) is the independent check: it replays a recorded batched turn and reads `boundingBox()` off the rendered nodes, and its overlap case is marked `test.fail()` today for exactly this defect. When this lands and the fixtures are re-recorded, that marker flips and stops being expected-to-fail.

christian-byrne:
correcting my own numbers above, and the direction of the correction matters.

that table used measured node *heights* but the CLI's *estimated* widths. the widths are wrong too. measured in the same browser pass, all 8 classes:

| class | CLI estimate | rendered | delta |
| --- | ---: | ---: | ---: |
| CLIPTextEncode | 250 | 400 | -150 |
| KSampler | 289 | 270 | +19 |
| EmptyLatentImage | 210 | 270 | -60 |
| CheckpointLoaderSimple | 210 | 270 | -60 |
| SaveImage | 230 | 270 | -40 |
| VAEDecode | 146 | 140 | +6 |
| ConditioningCombine | 263 | 215 | +48 |
| LatentUpscale | 222 | 270 | -48 |

the pattern is clean: every widget-bearing node renders at exactly **270**, multiline at **400**, and nodes with no widgets are content-derived. the CLI derives width from label text for all of them, which is the wrong model rather than a mistuned constant.

re-scored with measured widths and heights:

| fixture | main | with #886+#887 |
| --- | ---: | ---: |
| txt2img, 7 nodes | 47,463 px² | **16,123 px²** |
| 4 prompts + combines | 104,395 px² | 0 |

so this PR is necessary and large, but not sufficient on txt2img. the residual is `pos/ks 70x230`: `CLIPTextEncode` renders 400 wide, the CLI models 250, so `KSampler` gets placed 330px right of it and the real node reaches 400.

nothing about this PR changes — the multiline *height* number is measured and the fix is right. flagging it because my earlier "all three go to zero" was wrong, and because the width error is the same root cause (multiline widgets modelled as ordinary rows) showing up in the other axis.

caveat on the width rule: the 270 floor is 5 independent classes and solid; the 400 is one class. worth more samples before encoding it. happy to open that as a third PR if wanted.

christian-byrne:
exact-head Fallow audit for `773acde3cf817bf7c8610397ba0f00277f218805`: this repository and the pinned base `298fdae7821b2aa232f54820011b1131af5eef75` contain no Fallow workflow/configuration, and the PR changes only Python source/tests, so Fallow and a Fallow-triggering rebase are not applicable.

`uv run pytest -q tests/comfy_cli/test_layout.py` passes 41/41; `uv run ruff check .`, `uv run ruff format --check .`, and `git diff --check` pass. The ambient `pytest` attempt was invalid because it lacked `typer`; rerunning through the locked `uv` environment resolved that setup failure.

A repository-wide `uv run pytest -q` was stopped after 47% once unrelated existing suites had accumulated failures in run/preflight fixtures and Rich wrapping assertions; the changed layout suite remained green. No product edit or push was made by this audit.

skishore23 (dismissed):
The multiline term only reaches the planner. The size that ends up saved on the node is still the old estimate.

`assign_positions` passes `n_multiline`, but `workflow_ops.add_node` calls `estimate_size` a second time without it (around `workflow_ops.py:551`) and writes that result into the node's `size`. `apply_specs` goes through the same `add_node`, so batched builds hit this too. The next call's collision check reads that saved size.

Repro at the #888 head: two `CLIPTextEncode` in one `apply_specs`, then one more in a second call.

```
planner in the batch:  a at y=60, b at y=366   (correct, uses 400x236)
saved node size:       [270, 94]               (no multiline term)
next call places c at: x=390                   (a really spans 40..440, so ~50px overlap)
```

Passing it in `add_node` fixes it. The saved size becomes `[400, 236]` and `c` lands at x=520:

```python
    size = layout.estimate_size(
        len([p for p in m.inputs if p.is_link]),
        len(m.outputs),
        len(_widget_names),
        n_multiline=sum(
            1 for p in m.inputs
            if getattr(getattr(p, "options", None), "multiline", False) and p.name in _widget_names
        ),
        ...
```

Better still, move the count into one helper in `layout.py` that both callers use, and add a test that checks the saved `size` after `apply_specs`. The current tests only exercise `assign_positions`, which is why this got through.

Nit: `test_multiline_count_cannot_exceed_widget_count` accepts `_widgets_height(1, 5)` charging five multiline areas to a one-widget node. Clamping `n_multiline` to `n_widgets` seems more honest than codifying that.

christian-byrne:
you were right, and the bug was worse than a missed argument.

`add_node` calls `estimate_size` a second time and writes that into the node's `size`, and that saved size is what every later collision check reads. So the planner was correct and the persisted state was not — a batch places cleanly, then the next call measures against a record of the first that is 142px too short. Fixed, and both callers now go through one `layout.count_multiline` helper as you suggested; two call sites each computing it their own way is what let them diverge.

Took the nit too. `_widgets_height` clamps `n_multiline` to `n_widgets`, and the old test that asserted a one-widget node could be charged five multiline areas is gone — codifying a 700px over-measure as intended behaviour was the wrong call.

New `tests/comfy_cli/test_persisted_node_size.py` asserts the saved artefact rather than the planner's return value, which is the gap you identified: every existing layout test exercised `assign_positions`. Two of them verified red with the `add_node` argument removed.

One thing I did not do. Your two-batches-then-one-more overlap repro does not go red on this branch — the height is fixed here but width is still text-derived (249.9 for `CLIPTextEncode`), so the third node lands at x=369.9 and clears. It needs the 400px width floor, which is why you saw it at the #888 head. I moved that test there rather than assert it here, where it would have passed for the wrong reason. Caught only because I checked it failed without the fix.

christian-byrne:
exact-head Fallow recheck for `c7872982c4a5b2586b26f9520cf99113f94212bb`: the pinned stacked base remains `298fdae7821b2aa232f54820011b1131af5eef75`, which contains no Fallow workflow/configuration. The changed surface is Python source and tests only, so Fallow and a Fallow-triggering rebase remain not applicable.

`uv run --extra dev pytest tests/comfy_cli/test_layout.py tests/comfy_cli/test_persisted_node_size.py` passes 46/46; repository-wide `uv run --extra dev ruff check .`, `uv run --extra dev ruff format --check .`, and exact-range `git diff --check` pass. A repository-wide 7,654-test Pytest run was bounded to two minutes and reached 15% with no observed failure. No product edit, rebase, or push was needed.

annehe9 (approved):

PR #888, opened 2026-09-17
fix(layout): floor node width at what widget-bearing nodes actually render

christian-byrne:
Stacked on #887. The third and last term of the same defect — and txt2img needs all three.

## Why

Under-estimating width is what produces the overlap users report: the placer puts the next column a node-width away, and the real node reaches past it.

LiteGraph's own formula is `NODE_WIDTH * (1.5 if widgets else 1.0)` = 210, which **no widget-bearing node actually renders at**. Measured across 27 classes in a real browser:

- every node carrying a widget renders at **270 or wider** — 14 of 14 non-multiline widget classes sit at exactly 270 when their content is narrower
- every node carrying a **multiline** widget renders at **400 or wider** — 7 of 7
- `KSamplerAdvanced` 312, `ControlNetApply` 317.9, `CheckpointLoader` 396.9 exceed it on content, which is what confirms this is a **floor** and not a fixed width

I originally read the first five samples as "widget nodes are exactly 270". That was wrong — it held by luck on a small sample, and the next nine classes disproved it. The floor is the rule that survives all 27.

## Effect

On the eight classes in the release fixtures, five were under-estimated: `CLIPTextEncode` by **150**, `EmptyLatentImage` and `CheckpointLoaderSimple` by 60, `LatentUpscale` by 48, `SaveImage` by 40. After this, zero are.

Scoring placer output against measured width **and** height:

| fixture | main | +#886+#887 | +this |
| --- | ---: | ---: | ---: |
| txt2img, 7 nodes | 47,463 px² | 16,123 px² | **0** |
| 4 prompts + combines | 104,395 px² | 0 | **0** |

#886 and #887 fixed the height term. This is the width term of the same root cause — multiline widgets modelled as ordinary rows — showing up in the other axis. txt2img does not reach zero without all three.

## What is deliberately not fixed

The three **over**-estimates (`KSampler` +19, `ConditioningCombine` +48, `VAEDecode` +6) are left alone. Over-spacing wastes canvas; nobody has ever filed that. Under-spacing is the bug.

`Note` and `MarkdownNote` carry a multiline widget but render at 140, because they are annotation nodes with no slots. They get floored to 400 and will be over-spaced. Left deliberately: the direction argument above applies, and special-casing them on "no inputs and no outputs" is a guess I have not measured.

## Tests

17 parametrized cases asserting the floor never exceeds what a class renders (a floor above the rendered width would over-space every node of that shape), plus four behavioural tests covering each branch. All verified red on the parent; 62 pass in the layout suite.


skishore23 (commented):
The floor logic looks right to me, and the tests check the right direction: the floor never goes above the measured width.

Same gap as on #887, just on the width side. `workflow_ops.add_node` doesn't pass `n_multiline`, so a multiline node gets saved at width 270 instead of 400, and the next call places against 270. The #887 fix covers both. I'll approve once that lands.

christian-byrne:
Exact-head Fallow audit at `a1bb6e7db4c3b24d1599b7d7d2afec5fdd27a14a`: Fallow is not applicable because the pinned base `773acde3cf817bf7c8610397ba0f00277f218805` contains no Fallow workflow/configuration and this PR changes only Python files. No Fallow-triggered rebase applies; the PR merge-base equals the pinned base.\n\n`uv run --extra dev pytest tests/comfy_cli/test_layout.py` passed 62/62, `ruff check .`, `ruff format --check .`, and exact-range `git diff --check` passed. The bounded full suite reached 28% before its two-minute cap; one transient failure in `test_build_digest_cache.py` reran green 7/7. The final live guard preserved the exact head/base, Christian authorship, no approval, and no auto-merge request. No product edit, push, or merge was needed.

annehe9 (approved):

christian-byrne:
retargeted to `main` ahead of merging #887, so that `--delete-branch` does not auto-close this PR the way it closed #883 when the bottom of this stack landed on 2026-09-17. a closed PR cannot be reopened or retargeted while its base branch is missing, so the order matters more than it looks.

will rebuild this on `main` once #887 lands: a squash merge makes a plain rebase replay the pre-squash commits against their own squashed result and conflict on every file.

PR #891, opened 2026-09-18
fix(workflow): treat unregistered socket types as links in UI->API lowering

skishore23:
## Summary

`_is_widget_input` in `workflow_to_api.py` gave a `widgets_values` slot to any socket type that wasn't all-uppercase (`if not input_type.isupper(): return True`). The frontend decides by registry instead: a type is a widget only if a widget constructor is registered for it (`widgetStore.inputIsWidget`). Letter case plays no part.

Mixed-case **link** types therefore stole a slot. `VHS_LoadVideo.meta_batch` (`VHS_BatchManager`) is declared before the `format` widget, so the converter lowered
```
widgets_values: ["clip.mp4", 0, 0, 0, 0, 0, 1, "None"]
→ "meta_batch": "None", "format": "AnimateDiff"
```
and the worker crashed on `meta_batch.inputs` with `'str' object has no attribute 'inputs'`. That's 117 failed cloud jobs in 30 days (Langfuse: 106 dev, 11 prod). Other affected link types in the cloud catalog: `SUPIR_cond_pos/neg`, `SEG_ELT_*`, `SmolLM2_MODEL`, `SegPipeline`, `Reel`.

The fix drops the case heuristic. The engine's `_is_link` already uses the registry rule, so the converter and the editor (`add_node` / `set_widget` widget order) now agree.

## Verification

- Red before the fix: `TestMixedCaseLinkType::test_mixed_case_link_socket_owns_no_widget_slot` failed with `{'format': 'AnimateDiff', 'meta_batch': 'None', ...}`
- `uv run --locked --extra dev pytest tests/comfy_cli/test_workflow_to_api.py tests/comfy_cli/test_workflow_to_api_promoted.py tests/comfy_cli/test_validate_lowers_ui.py tests/comfy_cli/test_subgraph_gallery_templates.py` (127 passed)
- Full suite (`uv run --locked --extra dev pytest tests/comfy_cli`): one failure, `test_http.py::test_an_unloadable_supplement_falls_through_to_the_platform_roots`, which also fails on untouched `main` on this machine
- `ruff check` / `ruff format --check` (0.15.15)
- Lowered all 641 official workflow templates (Comfy-Org/workflow_templates) against the cloud catalog before and after: no validate-result changes from this PR

## Follow-up

The cloud agent image must bump its pinned comfy-cli commit to pick this up. A companion PR adds the validate check that catches a literal in a link socket.


annehe9 (approved):
Approving. I checked the frontend rule (`widgetStore.inputIsWidget` is `widgets.value.has(type)`). The new test fails on main with the `meta_batch: 'None'` shift. A cached catalog snapshot has 11 input types that aren't all uppercase, and none has a registered widget. The GPU runner failure is in `test_e2e_uv_compile.py` and fails on every recent PR.

Small things, none blocking:
- `tests/comfy_cli/test_workflow_to_api.py:1137`: the `TestWildcardInputType` docstring still refers to the lowercase fallback in `_is_widget_input`, which this PR removes.
- The last sentence of the new comment in `_is_widget_input` is bug history. The commit message already has it.
- This also fixes `KSamplerVariationsStochastic+.variation_seed` (type `INT:seed`), which the old rule gave a slot. Worth adding to the list in the description.

I filed two older gaps in the same rule as #893. Newer frontend widget types like `COLOR` still get no slot, on main too. The engine and the converter also disagree on `NUMBER`.

PR #894, opened 2026-09-18
docs(skills): agent permissions — right data dir for approvals, call the CLI as comfy, read a proxy 403 correctly

skishore23:
## Why

These come from the tool-call log of a local agent on Windows: 52 threads, 131 shell calls, 68 of them failed. The same few mistakes kept coming back, and each one sent the user off to do by hand something the agent could do, or could have unblocked with a single approval.

| What the model did | What went wrong |
|---|---|
| Called the CLI as `…\venv\Scripts\comfy.exe` or `python -m comfy_cli` | The sandbox answered "Access is denied", and the model concluded it "cannot launch the CLI". Plain `comfy` had worked a few minutes earlier. |
| Read a refused `dir` / `if exist` (or an empty `2>nul`) as absence | It told the user the install had no `.venv` and no `models` folder. Both existed. |
| Handed the user `comfy agent allow --approve <id>` | The agent ran with a non-default `AGENT_DATA_DIR`, so the user's terminal looked in `~/.comfy-agent`. `--approve` answered `agent_unknown_request` and `permissions` showed the agent as not running, even though the request was waiting. |
| Pointed the user at a panel Approve button | That panel build had no such button. |
| Read `Tunnel connection failed: 403` from `comfy build refs resolve` as "my sandbox only allows loopback" | It was one host missing from the egress list, so `request_host` was the move. |

## Change

Mostly `comfy-agent-permissions/SKILL.md`, which already owns the approval flow and the Windows notes, plus one CLI change from review (below). Nothing is hardcoded:

- The model reads `data_dir` from `comfy --json agent permissions`, and passes `--data-dir` when it isn't the default.
- It mentions a panel button only when it knows the panel has one.
- Windows: call the CLI only as `comfy`, and treat a refused look as not proof of absence.
- In the host row, `Tunnel connection failed: 403` from a CLI command is named as the same refusal as `egress denied`.

**From review (ff030ebd):**

- `comfy agent permissions` now *builds* its `grant` hints rather than hardcoding `comfy …`: they name the CLI as this process was started (`comfy` / `comfy-cli` / `comfycli`, or `<python> -m comfy_cli`) and carry `--data-dir` when the agent's dir is not the one a bare terminal resolves. argv[0] is trusted only when its basename is one of our entry points, so a hint never hands the user a sandbox-only path — the `…\venv\Scripts\comfy.exe` hand-off in the log is exactly that failure. The skill copies `grant.approve` / `grant.deny` instead of composing a command.
- New table row: the hosts the proxy never opens (`_REFUSED_HOSTS`) must not go to `request_host` — `approve` vets with `vet_host` and refuses them, so such a request waits until denied. A test pins that row against `_REFUSED_HOSTS`.

## Testing

- `uv run --locked --extra dev pytest tests/comfy_cli/agent tests/comfy_cli/skills`: 151 passed (includes the new grant-hint and refused-host tests).
- Full `tests/comfy_cli`: 7479 passed; the one failure, `test_http.py::test_an_unloadable_supplement_falls_through_to_the_platform_roots`, fails identically on `main` on this machine.
- `ruff check` / `ruff format --check` (0.15.15).
- The wording matches real outputs from `comfy agent permissions`, `allow --approve` (`agent_unknown_request`), and the egress proxy's 403.


comfy_cli/skills/comfy-agent-permissions/SKILL.md
```diff
@@ -35,7 +35,7 @@ ask; it can never make the answer.
 | Blocked | Do | Then |
 |---|---|---|
 | A folder (write refused as outside the project; "Access is denied" from the shell on Windows) | call `request_path` with the folder and what you need it for | tell the user a request is waiting and how to approve it; retry only after it is approved |
-| A host (`egress denied: <host>`) | call `request_host` with the host and what the download or install is for | same; rerun the command after the approval |
+| A host (`egress denied: <host>`; from a CLI command, `Tunnel connection failed: 403`, which is the same refusal) | call `request_host` with the host and what the download or install is for. It is one host missing from the list, not a shell cut off from the network | same; rerun the command after the approval |
```
huntcsg: > It is one host missing from the list, not a shell cut off from the network

I'd recommend language like:
> The host is missing from the list and does not indicate the shell is cut off from the network

comfy_cli/skills/comfy-agent-permissions/SKILL.md
```diff
@@ -90,3 +99,15 @@ ComfyUI install, and console utilities (`choice`, `waitfor`, `timeout`,
 `for %f in (<folder>\*) do @echo %f`, use absolute paths, and read exit
 `0xC0000142` as "did not run". Python from the install's venv runs pure-Python
 code (`py_compile`) but cannot import torch there.
+
+- **Run the CLI as `comfy`, by that name only.** The `comfy` on your PATH is
+  the one that works from this shell. The CLI's own `comfy.exe`, or
+  `python -m comfy_cli`, from its venv folder answers "Access is denied": that
```
huntcsg, line 107: Can we solve this somehow by looking at sys.argv[0] so we are guaranteed to give them exactly the correct command back, whatever it may be?

huntcsg (approved):
PTAL at my comments. Nothing blocking.

comfy_cli/skills/comfy-agent-permissions/SKILL.md
skishore23: Taken in ff030ebd, using your wording: "The host is missing from the list and does not indicate the shell is cut off from the network".

comfy_cli/skills/comfy-agent-permissions/SKILL.md
skishore23: Good idea — done in ff030ebd, and it also covers the data dir, which was the actual incident.

`comfy agent permissions` now builds its `grant` hints instead of hardcoding `comfy …`:
- **The name**: the entry point this process was started as (`comfy`, `comfy-cli`, `comfycli`, with `.exe` folded), or `<python> -m comfy_cli` when started as a module.
- **The dir**: `--data-dir "<root>"` rides along whenever the agent's dir is not the one a bare terminal resolves.

So `grant.approve` is the whole command, and the skill now says to copy it rather than compose one.

One caveat on `sys.argv[0]` itself: I deliberately do **not** echo it back verbatim. These hints are for the *user's* terminal, a different process, and a full path out of the agent's sandbox is exactly what produced the `…\\venv\\Scripts\\comfy.exe` → "Access is denied" hand-off in the log. So argv[0] is trusted only when its basename is one of our entry points; anything else (a test runner, an embedding host) falls back to `comfy`. Tests cover all six shapes, including `pytest`.

comfy_cli/skills/comfy-agent-permissions/SKILL.md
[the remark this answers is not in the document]
skishore23: Valid, and fixed in ff030ebd. Confirmed in code: `request_host` only appends to `pending` (`comfy_cli/agent/__init__.py`), and vetting happens in `approve` → `vet_host`, which rejects `_REFUSED_HOSTS` outright. So a request for one of them can never be approved and waits until denied.

The blocked-by table now has a separate row for those hosts (telemetry, shared object storage) saying not to call `request_host` and what to do instead. `test_agent_permissions_skill_names_every_permanently_refused_host` pins that row against `_REFUSED_HOSTS` so a host added there can't drift out of the skill; I mutation-checked it (dropping a host from the row fails the test).

comfy_cli/agent/command.py
```diff
@@ -59,12 +61,46 @@
...
+
+def grant_hints(root: Path) -> dict[str, str]:
+    """The exact commands that approve or grant against ``root``.
+
+    ``--data-dir`` rides along whenever the agent's dir is not the one a bare
+    terminal resolves (``AGENT_DATA_DIR``, else ``~/.comfy-agent``): the user's
+    shell is a different process, and without it ``--approve`` answers
+    ``agent_unknown_request`` while the request is still waiting.
+    """
+    flag = "" if root == data_dir(None) else f' --data-dir "{root}"'
```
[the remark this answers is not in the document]
skishore23: Valid, fixed in a0aa1076. `data_dir()` only expands `~`, so a relative `--data-dir` (or a relative `AGENT_DATA_DIR`, which also made the "is it the default?" comparison succeed between two relative paths) went into the hint verbatim, and a shell with a different cwd would target another dir. `grant_hints` now leaves the flag out only when the root is absolute and equals the default, and otherwise emits `root.absolute()`. `test_permissions_grant_makes_a_relative_data_dir_absolute` covers both the `--data-dir rel` and `AGENT_DATA_DIR=rel` cases. I used `absolute()` rather than `resolve()` so a symlinked data dir is shown the way the user named it.

skishore23:
@coderabbitai, re the outside-diff finding on `comfy_cli/agent/command.py:210` (hardcoded approve/deny line in pretty `permissions`): valid, fixed in a0aa1076. The per-request line now renders `grant.approve` / `grant.deny` with the first pending id filled in, so it carries the detected CLI name and `--data-dir` the same way the JSON grant does. `test_cli_permissions_lists_pending_and_names_approve_first` now asserts the `--data-dir` form. `SKILL.md` also tells the agent to hand over `grant.path` / `grant.host` for direct grants. The static examples stay as illustrations.

skishore23:
@coderabbitai, re the new outside-diff item on `SKILL.md:46-48` / `66-67` (replace the static grant examples): declining. Those blocks show the user what the commands do, for the common default-dir case. The rule the agent has to follow comes right after them in bold: "Hand over the command the CLI prints, not one you compose", then copy `grant.approve` / `grant.deny`. a0aa1076 adds the same instruction for direct grants (`grant.path` / `grant.host`). Replacing the examples with field names would stop the skill from showing what the commands actually look like. This PR is already approved, so I'm keeping the doc change to that one line.

PR #898, opened 2026-09-19
fix(deploy): list every location the compute catalog sells in (BE-15857)

james00012:
**TL;DR:** comfy deploy refs compute and the comfy deploy up pickers now offer B200, H100 and H200, and show United States as a location of its own, where before they offered only the RTX PRO 6000 in one datacenter.

Closes [BE-15857](https://linear.app/comfyorg/issue/BE-15857).

## Why

The deploy service answers a catalog request with its datacenters alone unless the request asks for every level, and it sells B200, H100 and H200 only on the wider United States row. The command line tool never asked, so three of the four GPUs we sell could not be picked from it.

## What changed

```mermaid
flowchart LR
    subgraph cli["service: comfy-cli"]
        subgraph deploy["component: deploy"]
            REQ["the catalog request"]
            OUT["refs table and region picker"]
        end
    end
    subgraph svc["service: comfy-deploy"]
        CAT["endpoint: GET /v1/compute-catalog"]
    end
    REQ -- "levels=all" --> CAT
    CAT -- "every location, with level and parent" --> OUT
    classDef changed fill:#ffe08a,stroke:#b8860b,color:#000;
    classDef elsewhere fill:#f6f6f6,stroke:#999,color:#444;
    class REQ,OUT changed
    class CAT elsewhere
```

- **deploy · catalog request**: asks for every level, so Anywhere and United States come back above US-NE-1
- **deploy · refs table**: prints each location's level and parent as the server sends them
- **deploy · region picker**: names each choice with its level, such as United States (country)
- **skills · comfy-deploy**: says a wider location such as us is a valid region

## Landing it

- Revert is clean: one commit, no stored state.
- refs compute --json now carries the wider rows too, so a script reading its first row gets anywhere rather than a datacenter; the changelog says so.

## Validation

- Development environment: comfy deploy refs compute on this branch against staging comfy-deploy, with a minted staging token in place of the stored sign-in, listed anywhere and us with B200, H100, H200 and RTX PRO 6000, and US-NE-1 with RTX PRO 6000, each with its level and parent; the region picker offered Anywhere (global), United States (country) and US-NE-1 (datacenter) for the RTX PRO 6000, and the GPU picker offered each of the four classes once.
- Unit: three new tests fail on main and pass here (the wire request carries levels=all, the table shows level and parent, the picker names the level); 7,596 tests pass, and the two failures are identical on main.

**Not run:** A deployment created from a wide location picked here: the create path is unchanged and already accepts us and anywhere from the flags, so the picker only changes what is offered.; Production's catalog: its comfy-deploy tag already answers levels=all, read from the chart rather than called..


james00012:
Approving. Reads as: the CLI asks for every catalog level and shows each location's level. Head d73b52e, judged as comfy-cli's owner.

**Optional**
* **Correctness: an 80-column terminal cut the region id a user copies** `comfy_cli/command/deploy_refs.py:88`. Fixed in d73b52e by folding every column.
* **Contract: the bundled skill named which cards sell where and never said what a blank availability means** `comfy_cli/skills/comfy-deploy/SKILL.md:135`. Fixed in d73b52e.
* **Scope: the refs compute schema did not declare level or parent** `comfy_cli/schemas/deploy_refs_compute.json:18`. Fixed in d73b52e as optional strings with no enum.
* **Tests: no case covers a non-string level in the picker** `tests/comfy_cli/command/test_deploy_compute.py:40`. Noted; the table path was checked by hand.

Clean: Modularity, Verbosity, Resilience, Access, Reachability, Rollout.
Native review: 0 findings, folded into the list above.
Checked: prod comfy-deploy is pinned to a tag that already answers levels=all.

vqt123:
Approving at d73b52e00e.

Checks:
- The premise holds at the server: `services/comfy-deploy/openapi.yaml` on cloud `main` (6a920ff0cc) documents `levels` on `GET /v1/compute-catalog` — absent returns datacenters alone with no `level`/`parent`, `all` (any case) returns every level, any other value is a 400. `levels=all` is the one string that widens the answer.
- The one item under **Not run** holds: a wide location is creatable, so the picker cannot offer something `up` then refuses. `apiserver/compute/chooser.go` `ChooseProvider` — "The location may name any node of the tree, so the walk goes down" — and `apiserver/api/api.gen.go:239` describes the wide rows as locations a deployment can be created in.
- The re-indexed sort is right: `comfy_cli/command/deploy_refs.py:83` moved the third key from `row[3]` to `row[5]`, and `level`/`parent` went in at positions 3 and 4 of the tuple built at `deploy_refs.py:75-83`, so `row[5]` is still `gpu_class`.
- `comfy_cli/schemas/deploy_refs_compute.json:14` already had `additionalProperties: true`, so `level` and `parent` validated before this PR; declaring them documents the shape rather than fixing a rejection.
- Both readers of the catalog were checked, not only the two the diff touches: `deploy_refs.py:103` and `deploy_compute.py:10` are the only callers of `get_compute_catalog`. Neither indexes a fixed row nor switches on a level name, and both fall back cleanly when `level`/`parent` are absent (`deploy_refs.py:48-51`, `deploy_compute.py:46`), so the CLI still renders against a comfy-deploy old enough to ignore the parameter.
- Ran rather than read: `uv run --extra dev --frozen pytest tests/comfy_cli/command/test_deploy_compute.py tests/comfy_cli/command/test_deploy_refs.py tests/comfy_cli/test_deploy_api.py -q` at d73b52e in a worktree — 134 passed.
- Outside the diff and not in the body: `chooser.go` routes `us` and `anywhere` to Modal for a flagged workspace and "a datacenter id from a CLI or a direct API caller" to RunPod. Until this PR the CLI could only send datacenter ids, so its deployments always landed on RunPod; now a user picking `us` or `anywhere` can land on Modal. Same as the dialog, so parity rather than a defect — noting it because the change of vendor is invisible from the CLI.

Findings: none.

<sub>~276k effective tokens for this review (1.3M raw; cache reads weighted 0.1x, cache writes 1.25-2x)</sub>

vqt123 (approved):

PR #912, opened 2026-09-21
feat(build): push says how far along a model upload is, to a person and to an agent (BE-16052)

vqt123:
## Description

**TL;DR:** `comfy build push` says how far along a model upload is. Today it prints nothing between starting a multi-GB upload and finishing it.

**Why:** a private model is routinely many GB, and the command is silent for the whole transfer, to a person and to an agent alike. Part of [BE-15748](https://linear.app/comfyorg/issue/BE-15748).

**Before / After**

- **A plan line, then bytes, rate and time left (After).** Three surfaces, chosen by who is reading: one redrawn Rich line for a person at a terminal, a plain line per sample when the output is captured (no carriage returns in a log), and `upload_plan` / `upload_progress` / `upload_complete` events for an agent under `--json-stream`. Under plain `--json` the events go to stderr, because stdout there is contractually one envelope.
- **Follows the CLI's existing progress shape** for a transfer with a total (`file_utils.py:356`: description, bar, numbers, `transient=True`).
- **A 1 s ticker samples,** so a stalled socket still reports a rate falling to 0 rather than a frozen line.
- **`comfy_cli/output/progress.py` is new and shared:** `Surface`, `human_bytes`, `human_seconds`, `surface_for(renderer)`. The deploy watch in BE-16056 words its wait with the same functions, so the two cannot drift. Sizes count in 1024s because the portal's `formatBytes` does.
- Schema `comfy_cli/schemas/build_push_event.json`, registered in `STREAM_EVENT_SCHEMAS`. The build schema orphan sweep read only `COMMAND_SCHEMAS` and reported it as abandoned; it now reads both, as `test_discovery.py` already does.

Closes [BE-16052](https://linear.app/comfyorg/issue/BE-16052).

## How has this been tested?

`pytest`: 7738 passed, 2 failed. **Both failures are not from this change:** `test_node_deps.py::test_non_pep440_installed_version_is_unknown_not_a_crash` and `test_http.py::test_an_unloadable_supplement_falls_through_to_the_platform_roots` fail identically on the unmodified base commit `3aa2fc0` in a clean checkout with a fresh environment. That run was on Python 3.12 and CI targets 3.10, which is the likely cause. `ruff check` and `ruff format --diff` clean. 13 new tests in `tests/comfy_cli/output/test_progress.py`.

Manually: real uploads of 722,601,100 and 7,272,719,498 bytes, at a terminal and piped (313 bytes, 5 lines, 0 escape codes when captured). Two clean-room runs gave an agent only the command and asked it to narrate the upload; it reported progress and the finish correctly from the plain lines as well as from the events.

**Not run:** Windows.

**Follow-up filed:** `build push --dry-run` prints nothing to a person ([BE-16130](https://linear.app/comfyorg/issue/BE-16130)).

## Documentation
- [x] No user docs change; `--help` text covers it.

## Screenshots or screen capture
Before and after recordings are on the Linear ticket.


wei-hai (approved):
Reviewed streamed request sizing, upload callbacks, ticker lifecycle, progress output modes, and schema integration. No blocking findings. All 99 targeted upload/reporter/transport tests pass locally on Python 3.10.

The GPU CI job remains red: its log shows the two uv-compile conflict-message tests failing, the same failures across these four PRs in code outside these changes. Other reported checks pass. I did not rerun GPU E2E or live analytics/provider operations.

vqt123 (commented):
Self-review of the fix push, head e8f9310 (e8dd6fe, 61dc359, e8f9310), read cold against d3ae073 together with Rich's `Progress.add_task` / `refresh` and `Live.start`.

**Findings**

1. `tests/comfy_cli/command/test_build_upload_progress.py`, the ticker test as first written: the wait loop parsed the NDJSON stream from the main thread while the ticker thread was still writing it, so a read landing between two writes could hit a half-written line and fail on the parse, not on the behaviour. **Fixed in e8f9310**: the wait counts newlines and the parse happens after the thread is joined.
2. Same file, `_LiveRenderer.info` raised: the refused-display test passed only because the reporter was muted before the completion line, and that was the thing under test, so the assertion was implicit. **Fixed in e8f9310**: `info` records what it is told and the test asserts nothing was said.
3. `comfy_cli/command/build_upload_progress.py` `_open_live`: `live.stop()` also runs when `start()` itself raised. Checked: `Live.start(refresh=True)` stops itself when its refresh raises, and `stop` on a Live that never started returns early, so the second stop is a no-op. Left.
4. `comfy_cli/schemas/run_event.json` keeps an enum under the same "advisory, open-ended" wording, the contradiction this push removes from `build_push_event.json`. That is the `comfy run` contract, out of this PR's scope; noted for a separate change.

**Checked**: each new test fails with its fix reverted (the `OSError` escapes `_open_live`; the enum rejects the unknown type). Full suite on e8f9310: 7896 passed, 38 skipped, 2 failed, and the two are `test_node_deps::test_non_pep440_installed_version_is_unknown_not_a_crash` and `test_http::test_an_unloadable_supplement_falls_through_to_the_platform_roots`, which fail on `main` and touch no file in this branch.

<sub>~1.1M effective tokens for this review (19.9M raw; cache reads weighted 0.1x, cache writes 1.25-2x)</sub>

james00012 (approved):
Approving. Reads as: build push prints a plan, then bytes, rate and time left per file, as a redrawn line, plain lines or events on stderr under --json. Head e8f9310, judged as comfy-cli's owner.

**Optional**
* **Resilience: a push whose caller never drains stderr hangs after the upload finishes** `comfy_cli/command/build_upload_progress.py:179`. The ticker blocks writing under the lock once the pipe fills, and the join has no timeout.
* **Correctness: the time left beside a file is the whole push's** `comfy_cli/command/build_upload_progress.py:360`. File 1 of 3 showed 24m 56s left with about 8m to go on it.
* **Correctness: piped output wraps one sample over two lines** `comfy_cli/command/build_upload_progress.py:286`. The piped console is 80 columns, so a long filename pushes the time left onto its own line.
* **Correctness: piped output still gets redraws when a color variable is set** `comfy_cli/output/progress.py:68`. Rich reports a terminal under FORCE_COLOR, which breaks the ticket's no-redraw criterion.
* **Modularity: one line mixes GiB and MB/s for the same 1024-based bytes** `comfy_cli/command/build_upload_progress.py:302`. Render the byte counts with the shared formatter.
* **Verbosity: the no-reporter path runs only in tests** `comfy_cli/command/build_push.py:309`. The one production caller always passes a reporter, so make it required.
* **Scope: the diff is 1,490 lines, past the 400-line ceiling** `comfy_cli/output/progress.py:1`. Next time land the shared wording module on its own ahead of the reporter.

Clean: Access, Tests, Reachability, Contract, Rollout.
Native review: 10 findings, folded into the list above.
Checked: A scratch push whose parent read only stdout hung after its upload ended. Piped runs showed the redraws and the split line.
Bots: The ticker test fix holds at this head: it fails once the ticker is disabled.

PR #917, opened 2026-09-22
feat(deploy): support API keys for builds and deployments

huntcsg:
## Description

`comfy build` and `comfy deploy` can use `COMFY_CLOUD_API_KEY` or a key saved with `comfy cloud set-key` without an OAuth login. Both clients use the existing shared credential resolver and send the selected credential as a bearer token.

OAuth retains its existing precedence and refresh behavior. API keys and forwarded bearer tokens are never refreshed or replaced after a 401. Authentication errors and the README now describe the key-based workflow.

Depends on the server support in [cloud#10161](https://github.com/Comfy-Org/cloud/pull/10161) being deployed. Companion draft for [BE-11394](https://linear.app/comfyorg/issue/BE-11394/implement-api-key-support-for-managing-deployments).

## Validation

- `uv run ruff check .` and `uv run ruff format --check .`: passed.
- 289 focused tests passed, including the complete build/deploy command matrix with API keys, real bearer-header construction, OAuth precedence, and no refresh/fallback on rejected keys.
- Full suite with startup umask `022` and `TERM=xterm-256color`: **7,900 passed, 37 skipped, 1 failed**. The existing `test_edited_content_is_hashed_again[identical size]` digest-cache test also fails on untouched `origin/main` (`aec5220`); no unrelated code was changed.
- No live deployment performed.


vqt123 (approved):
Checks:
- The key goes out in the shape the servers read. `BuilderClient.from_session` and `DeployClient.from_session` (`comfy_cli/deploy_api.py:90`) now take `resolve_cloud_credential(purpose="cloud")` and send its value as `Authorization: Bearer`; the `WithIdentity` change in cloud#10161 (comfy-builder and comfy-deploy `httpkit/identity.go`) branches on a `Bearer comfyui-` token, so the two sides match.
- `deploy run` works with a key too: `comfy_cli/command/deploy_run.py:201` reuses the control-plane credential for the deployment endpoint, and the gateway on main already accepts `Bearer comfyui-` (`services/platform-gateway/server/auth.go:146`).
- Only a stored session refreshes: `_refreshes_on_401 = credential.source == "session"`, so a key or a forwarded `COMFY_CLOUD_AUTH_TOKEN` is never swapped after a 401. `COMFY_BUILDER_TOKEN` still wins in `_builder_client`.
- The red GPU-runner job is `tests/e2e/test_e2e_uv_compile.py` (Manager's uv-compile conflict message), not this diff; the three-platform test matrix passes.

Findings: none. Shipping this before cloud#10161 is deployed is harmless: a key-only caller gets a 401 whose hint already names the key.

<sub>~6.0M effective tokens for this review (41.8M raw; cache reads weighted 0.1x, cache writes 1.25-2x)</sub>

PR #919, opened 2026-09-22
fix: accept the set-widget addresses the agent derives

skishore23:
## Summary

Two set-widget errors from stg-v2 and nightly agent traces (Langfuse, 2026-09-20 to 21) where comfy-cli refuses input it produced itself:

1. **Binding key as address (22 calls).** `<instance>/primitive_string_multiline_2.value` is a key from `print_workflow`'s `bindings` map, and set-widget refused it as an unknown interior node. set-widget now resolves a binding key to the address it maps to, but only after the literal address fails, so a real node id always wins (`workflow_ops.py`).
2. **Legacy `PrimitiveNode.value` (8 calls).** A write to the node the primitive feeds already redirects onto the primitive, but a direct `51.value` was refused with "available widgets: (none)". A top-level PrimitiveNode now takes the write on its single widget, addressed as `value` or by its output marker's name (`cql/promoted.py`).

**Scope (review):** the Nano Banana and Flux emit work that was here moved unchanged to #921, so this carrier holds only the address edits. The pre-split head is kept on `backup/919-before-split`.

## Red-Green Verification

| Commit | CI Status | Purpose |
|--------|-----------|---------|
| `test: ...` | :red_circle: Red | Proves the tests catch the bugs |
| `fix: ...` | :green_circle: Green | Proves the fix resolves them |

## Test Plan

- [ ] CI red on the test-only commit
- [ ] CI green on the fix commit


christian-byrne:
Release split required: isolate address edits; preserve adapter work separately.

The release-critical minimum is only:

- `comfy_cli/workflow_ops.py`: binding-key fallback after literal-address failure.
- `comfy_cli/cql/promoted.py`: direct `PrimitiveNode.value`/marker resolution.
- `tests/comfy_cli/test_set_widget_agent_addresses.py`: discriminating coverage.

Please keep those changes in an existing-owner carrier and move/preserve the unrelated Nano Banana and Flux adapter work (`command/generate/emit.py`, partner-node fixture, emit/model-variant tests) in its separate existing-author route. Do not discard it, and do not make the final Agent pin depend on it.

After the split, return both exact heads and focused/full check receipts. The edit-only carrier also needs substantive review before inclusion.

</details>

skishore23:
@christian-byrne split done.

- **This PR (address edits only):** head `90b1fbe1b0f159061e5b983c1f0ea6609bf5c62d`. It carries `workflow_ops.py` (binding-key fallback after the literal address fails), `cql/promoted.py` (direct `PrimitiveNode` value/marker) and `tests/comfy_cli/test_set_widget_agent_addresses.py`.
  - **Red:** `46f15baa` failed on exactly its 3 tests (7857 passed).
  - **Green:** `90b1fbe1` passes (run 35705469421).
- **Adapter work, preserved unchanged:** #921, head `431cf0d34366f362a52e279a4dd3aeb236602dae` (`command/generate/emit.py`, the partner-node fixture, the emit and model-variant tests).
  - **Red:** it failed on exactly its 2 tests.
  - **Green:** passes (run 35705470502).
  - The Agent pin does not depend on it.
- **Pre-split head:** kept on `backup/919-before-split`.

comfy_cli/workflow_ops.py
```diff
@@ -636,9 +636,35 @@ def set_widget(
     try:
         return _set_widget_impl(workflow, graph, node_id, widget, value, actor=actor, base_version=base_version)
     except ValueError as e:
```
christian-byrne, line 638: P1: Restrict fallback to address-resolution failures; literal write errors must not redirect.

The wrapper catches every `ValueError`, including a literal node that exists but rejects the widget or value, then retries another node if `render_py().bindings` has the same key. That contradicts the documented “real node id always wins” rule and can silently mutate the alias target. Add a collision regression where the literal node exists, its write fails, and a same-spelled binding maps elsewhere; preserve the literal failure instead of redirecting.

</details>

huntcsg (approved):

